PR #2, opened 2017-02-10
Task/update vtkm

mclarsen:


cyrush:
should this guy got to master or develop?

mclarsen:
Once we get this stable, then we should make is master.

cyrush:
ok, in that case, I guess we would want to go branch -> develop -> master, but if thats too much overhead at this point, maybe we get ride for develop until we need it.

PR #12, opened 2017-04-14
Feature/noise source

mclarsen:


src/examples/synthetic/noise/noise.cpp
```diff
@@ -0,0 +1,450 @@
+#include <iostream>
```
cyrush: add our cpy right header here

mclarsen: So put the readme in the synthetic noise path?

PR #14, opened 2017-04-14
helper script for weak and strong scaling job sub scripts

cyrush:


cyrush:
These are now stale, I am declining.

PR #21, opened 2017-06-29
add circle ci support for testing docker builds

cyrush:


cyrush:
declining, merge target is incorrect (master instead of correct develop)

PR #29, opened 2017-07-09
add support for docker testing using circle ci

cyrush:
Currently testing w/o VTKm, until we can resolve opengl dep issues.


cyrush:
I had to explicitly add c++11 support to compile the noise examples. 

My hunch is that this worked before b/c when VTKm was in the picture, some cmake setting it imports turns on c++11 for all targets.

cyrush:
CI checks are looking better -- failures are due only to timeouts &  most cases succeed.

I still need to:
- [ ] test the patch for limiting the spack build jobs
- [ ] turn vtkm back on for docker

cyrush:
sans timeouts, things are healthy -- merging.

PR #31, opened 2017-07-14
use blt as core of alpines build system

cyrush:
Brings in blt: 

https://github.com/llnl/blt

To replace our current support for:

unit testing with gtest, fruit
lib support for cuda, mpi, and openmp


cyrush:
I have tested with and w/o cuda and alpine's tests are all passing. ( I checked for a cuda image manually as well)

The current failure is in BLT's smoke test for google mock. 

We see this on travis and on surface when using w/ gcc 4.9.3 + cuda 8.
I'll tackle early next week.

cyrush:
git archive doesn't handle submodules, so I'll need to update the `package.py` script.
This is whats undermining the docker example. 

- [ ] update `package.py` to package submodules

cyrush:
Going to test this script out:
https://github.com/Kentzo/git-archive-all/blob/master/git_archive_all.py

cyrush:
I think this is ready to merge, there is an issue with python in docker -- I'll hunt that down via another PR.

PR #38, opened 2017-09-12
the second great renaming

mclarsen:


mclarsen:
This seems like it is not being tested.

PR #40, opened 2017-09-21
update to use conduit v0.3.0 and a newer blt

cyrush:


cyrush:
@mclarsen , feel free to hold off it you like. 

There was one small change necessary to use conduit v0.3.0 (the mpi reduce func signature changed), so conduit v0.2.1 will not longer work after this is merged.

mclarsen:
Looking at the tests, this looks like everything is fine (even though the tests timed out), so I think it is safe to merge.

cyrush:
@mclarsen  thanks, still need to dive in to that python crash we see in docker. I was hoping the new conduit would fix that.

PR #44, opened 2017-10-02
WIP: testing caching of dependencies

mclarsen:


mclarsen:
Even a serial build takes a dump on marching cubes.

mclarsen:
@cyrush I am not sure why vtk-h builds "seemingly" reliably in its own repo, but I cannot build it here. Every time I am getting an internal compiler error and I have tried gcc-4.8, gcc-4.9, and gcc-5. 
I was attempting to play around with caching and I have figured out how to do it while still building vtk-m and vth-h every time (cache is only 450 MB).

cyrush:
for context, Spack defaults to build with RelWithDebugInfo, all the symbols for vtkm and vtkh overwhelmed the compilers in CI. As part of this PR, we are not building vtkm and vtkh with Release.

cyrush:
If you are trying to rebuild vtkm + vtkh, you can also us something like:

`uberenv_libs/spack/bin/spack uninstall --dependents vtkm`

PR #45, opened 2017-10-06
vtkm + vtkh spack: build with release instead of relwithdebuginfo

cyrush:


mclarsen:
These changes were integrated in #44

PR #46, opened 2017-10-18
add initial support for adios extracts

cyrush:


cyrush:
declining, this will go in as part of #47

PR #51, opened 2017-11-12
web interface updates

cyrush:
updates the web interface
adds web interface as option to flow and main (ascent) runtime
adds web tests


cyrush:
note: all hdf5 related urls moved last night, so this will undermine our tests.

PR #53, opened 2017-11-30
add flow python wrapper, enable python filters

cyrush:


cyrush:
changes need an updated conduit (which should be provided by uberenv on this branch), so it appears I need to force uninstall and reinstall conduit in travis

PR #55, opened 2017-12-08
moving scene management into vtkh

mclarsen:
This removes any scene management from Ascent. To accomplish this, several internal classes were created to manage intermediate results (so they will not be deleted) in the registry.


mclarsen:
@cyrush Does this look ok?

PR #56, opened 2017-12-18
Task/cinema

mclarsen:


src/ascent/runtimes/flow_filters/ascent_runtime_vtkh_filters.cpp
```diff
@@ -380,6 +389,238 @@ vtkh::Render parse_render(const conduit::Node &render_node,
...
+  {
+    m_times.push_back(m_time);
+    m_time += 1.f;
+
+    // add top level dir
+    string output_path = "cinema_databases";
+    
+    if(!conduit::utils::is_directory(output_path))
+    {
+        conduit::utils::create_directory(output_path);
```
cyrush: for MPI case, do we need to guard this to only occur on rank 0?

We could create our own ascent util (which uses the conduit one) that handles the MPI vs non MPI case transparently.

mclarsen: You are right. We don't need all ranks trying to create the same directory.

mclarsen: Ok. Fixed. I just grabbed the mpi comm and checked rank. I really don't see us have many other use cases for creating directories in filters.

cyrush: sounds good, make sure to push :-)

PR #61, opened 2018-01-09
Docs update

mclarsen:


mclarsen:
@cyrush Will you read my changes? I would like to have another set of eyes on them. Also, in the upper left corner of the sphinx template, there is a "v0.1.0" that I cannot figure where it is defined. Any idea?

cyrush:
The ver number is in the sphinx` config.py` file (I pushed a change to it already)

Did a quick read over & so far additions look great! I'll do a detailed review later today.

PR #62, opened 2018-01-10
add embedded py interpreter for flow, based on VisIt's implementation

cyrush:


cyrush:
@mclarsen  i think this is ready, do you want to give it a quick review?

src/ascent/runtimes/ascent_main_runtime.cpp
```diff
@@ -410,27 +410,69 @@ AscentRuntime::ConvertExtractToFlow(const conduit::Node &extract,
...
+       {
+         ASCENT_ERROR("broadcast of python script source failed");
+       }
+       // replace file param with source that includes actual script
+       params.remove("file");
+       params["source"] = n_py_src;
+     }
+#endif
+  
+    // todo, inspect args, if passed via file, read on root proc and broadcast
```
mclarsen: Isn't this what the above code does?

src/ascent/runtimes/ascent_main_runtime.cpp
```diff
@@ -410,27 +410,69 @@ AscentRuntime::ConvertExtractToFlow(const conduit::Node &extract,
...
     ASCENT_ERROR("Unrecognized extract type "<<extract["type"].as_string());
   }
  
   if(w.graph().has_filter(extract_name))
   {
-    ASCENT_INFO("Duplicate extract name "<<extract_name
-                <<" original is being overwritted");
+    ASCENT_INFO("Duplicate extract filter named " 
+                << "\"" << extract_name << "\"" 
+                << " original is being overwritten");
```
mclarsen: We need to get rid of these checks or make them ASCENT_ERRORS. Flow will not let us add duplicates anyway and will fail. Here we can give a more understandable error message than flow.

cyrush: thanks, i'll tackle these.

Not sure how to guard against the specific error about using the wrong python, b/c if thats the case we begin on a bad footing.

Issue #66, opened 2018-01-20
web svr: allow user set web root dir


cyrush:
and ability to copy necessary hosted assets to that web root dir

cyrush:
this will be necessary to support serving cinema databases as they are generated

cyrush:
resolved in #68

PR #68, opened 2018-01-23
add ability to set web doc root for web streaming case

cyrush:
adds utils for copying files and directories 
adds test for copying directories
adds Ascent option (web/document_root) that allows user to select where the web server doc root is, the Ascent web client resources (html files, js files, etc) are copied to this directory


cyrush:
resolves #66

cyrush:
@mclarsen, I'll add a test to check.

cyrush:
relative works fine (tried `.` and `here` and things were copied as expected)

PR #69, opened 2018-01-23
uber: add logic to find valid osx sdk root

cyrush:
@mclarsen, can you see if this resolves the issues on your mac with finding the osx sdk?


cyrush:
awesome, thanks for testing!

PR #70, opened 2018-01-25
install examples (proxies + noise) with supporting files

cyrush:
Installs lulesh, kripke, clover, and noise
adds readmes for each with simple examples, that ref the full ascent docs


cyrush:
thanks!

PR #76, opened 2018-01-26
refactor verbosity setting, provide option to catch exceptions

cyrush:
changes to Ascent::open() options

* Instead of accepting `ascent_info`, accept `messages`, valid entries are `quiet` or `verbose`

Controls if ASCENT_INFO messages are printed or muted.

`quiet` is the default case, ASCENT_INFO messages are muted

`verbose` enables printing of ASCENT_INFO messages 

* Add `exceptions` option, valid entries are `forward` or `catch`

`forward` is the default, and preserves the old behavior, where if an exception is thrown the calling app will need to handle it

`catch` catches conduit::Error exceptions at the interface. It's an important option b/c it provides an easy way to prevent program crashes when something goes wrong in ascent.

### DIFF
--- a/src/ascent/ascent.cpp
+++ b/src/ascent/ascent.cpp
@@ -77,7 +77,9 @@ quiet_handler(const std::string &,
 
 //-----------------------------------------------------------------------------
 Ascent::Ascent()
-: m_runtime(NULL)
+: m_runtime(NULL),
+  m_verbose_msgs(false),
+  m_forward_exceptions(true)
 {
 }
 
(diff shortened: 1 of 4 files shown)


cyrush:
@mclarsen check these out let me know if you think the options names and supported values are sane. 

The "trap all exceptions" at the interface is an important feature I have been wanted to add to insulate the host code from Ascent errors.

mclarsen:
Kitware's gitlab is currently down so the build is failing.

PR #85, opened 2018-02-03
add tutorial notes about fetching image from docker hub

cyrush:


cyrush:
@mclarsen we now have an image on docker hub, tested it against these instructions and everything worked. We can update the docker hub image if necessary. This is ready to merge pending CI checks.

PR #106, opened 2018-03-23
fixing save and removing print statements

mclarsen:


mclarsen:
I am having problems creating a unit test for an empty domain. It fails the blueprint::verify in the EnsureBlueprint filter

mclarsen:
vtkm filter API changed again.

mclarsen:
@cyrush 

There are multiple issues here. 

1. Thresholded data sets resulted in in a ExplicitCell set with mixed type, which we are not currently supporting fully in conduit. This should be fixed since Clipping will result non-uniform cell shapes. For now, I made a function on vtkh that checks to see if all cells are the same so we can use it.
2. I now use a `clean_mesh` function to replace `to_multi_domain` which doesn't fully check if the input data is invalid. This function checks to see what data may or may not be valid since some domains might not exist anymore or a domain could be empty ( no topology or fields).
3. This issue is unsolved. Currently `ASCENT_WARN` throws an exception which crashes the test. If the entire data set is culled by something like threshold, adding a warning crashes ascent. Should we change the default warning function or catch errors in flow filters?

cyrush:
thanks for fixing the exception behavior

PR #107, opened 2018-03-28
Fix Git URL for the repo

oruebel:
The Getting Started instructions list https://github.com/Ascent-DAV/ascent.git instead of https://github.com/Alpine-DAV/ascent.git as URL for the repo.


mclarsen:
Thanks Oliver! This was a side effect from find and replace during the great renaming.

mclarsen:
The checks are failing because vtkm changed. I am fixing that in vtkh and Ill merge this in.

PR #109, opened 2018-03-29
add test cases python extract + inception using the ascent runtime

cyrush:


cyrush:
merge fixes on trunk for vtk-m api changes

PR #118, opened 2018-06-13
Task/update build

mclarsen:


cyrush:
Now that we have some experience with packages.yaml -- I am going to look at using a packages.yaml for travis to avoid timeouts.

I'll aim for python, numpy, and mpich

Circle CI will still test building all tps via spack.

cyrush:
I think we are ready to merge?

mclarsen:
One thing  that is left out of this is a pinned spack version for ascent 0.x.0 and vtkh 0.y.o, but we can handle that outside of this merge request.

PR #119, opened 2018-06-15
expand field dtypes supported for rendering

cyrush:
adds support for float32, int8,int16,int32,int64, and uint8,uint16,uint32,uint64 f fields.

Because of the template paths we are using, only float32 and float64 are zero copied.


mclarsen:
Does this break rendering or does it work?

mclarsen:
Ok. I see what is going on. This is good first pass at this, since it should not break anything. We could in the future had a generic conduit ptr to vtkm::ArrayHandle for all types and this would allow us to zero copy everything. However, I think handing a renderer in int might break things in vtkm.

cyrush:
yes, all of these cases work - however as you can see they are not all zero copy.

We can expand the zero-copy support as long as it matches the selected / compiled vtkm code paths.
Theses choices are based on what template paths are currently available.

PR #121, opened 2018-06-21
changes to use new vtkh mpi interface

cyrush:


cyrush:
@mclarsen have external packages working, build time is ~45min. 

I can't external python or the python modules.
I think it make sense to swing over to this vs caching -- hoping it will be less brittle?

Still hitting the make -j issue with fortran modules some times, going to try to tackle that one can for all.

mclarsen:
Does it make sense to use caching and the external modules? It is helpful when builds only take 10-20 minutes?

cyrush:
yes you are right, 45 is too close to the timeout limit. 

I was getting confused about caching b/c I was deleting things, and frustrated that it was still using the cache.  I didn't see the caching statement near the top of the script. I'll get it all setup back up.

PR #128, opened 2018-07-12
uber spack add cfgs for stampede 2

cyrush:


cyrush:
adds gcc compilers and packages settings for building ascent on tacc's stampede 2

Issue #134, opened 2018-07-23
cuda failure when rendering 2d uniform grids


cyrush:
`t_ascent_mpi_render_2d` and the  `t_python_ascent_mpi_render` both fail when cuda is enabled. 

Poking a bit, in `ascent_vtkh_data_adapter.cpp:VTKHDataAdapter::UniformBlueprintToVTKmDataSet`

Something down stream isn't happy with the structured cell set for the 2d case.
Here is the logic which creates it.  @mclarsen traced this a bit, and I believe hit an assert failure about being able to resize an array. 

```
    if(is_2d)
    {
      vtkm::Id2 dims2(dims[0], dims[1]); 
      vtkm::cont::CellSetStructured<2> cell_set(topo_name.c_str());
      cell_set.SetPointDimensions(dims2);
      result->AddCellSet(cell_set);
    }
    else
    {
      vtkm::cont::CellSetStructured<3> cell_set(topo_name.c_str());
      cell_set.SetPointDimensions(dims);
      result->AddCellSet(cell_set);
    }
```

cyrush:
note: the test cases above fail in the pseudocolor plot, but I also tested the mesh plot and it failed as well.

cyrush:
Note: we cant test this w/o mpi in the mix -- I am going to change the mpi tests and move the failing case into t_ascent_render_2d

cyrush:
For the record, in #133 I did move this case to the non-mpi 2d case, so we can test it there

mclarsen:
#146

PR #143, opened 2018-08-30
refactor to support reging generic xform and extract filters with the…

cyrush:
… ascent runtime


cyrush:
@mclarsen looks like post vtk-m refactor, the 2d render test is failing with the error we use to only see with cuda (#134). Thats good, it gives us a better chance at tracking down the issue. 
I think its orthogonal to these changes.  There are also hwloc / mpi issues with cases where we don't use vtk-h.

mclarsen:
I did find someone else with the same problem:
https://github.com/travis-ci/travis-ci/issues/10019

PR #149, opened 2018-09-12
Added ADIOS2 Filter (Don't pull)

stevenwalton:
Added the ADIOS2 filter from the work done at ORNL. The filter allows
for the current version of ADIOS2 to export the data by means of
BPfiles, SST, DataMan, and HDF5.

### DIFF
--- a/scripts/uberenv/spack_configs/olcf/summit/packages.yaml
+++ b/scripts/uberenv/spack_configs/olcf/summit/packages.yaml
@@ -54,7 +54,7 @@ packages:
 # ORNL blueos CUDA
   cuda:
     paths:
-      cuda@9.2.148: /sw/summit/cuda/9.2.148/
+      cuda@10.1.105: /sw/summit/cuda/10.1.105/
     buildable: False
 # ORNL blueos mpi
 # note, we need to fake this as an existing mpi package


stevenwalton:
This setup does require a FindADIOS2.cmake to be placed in `adios2-install/etc`. I have provided the file (changed extension to be able to place here)
[FindADIOS2.txt](https://github.com/Alpine-DAV/ascent/files/2373256/FindADIOS2.txt)

stevenwalton:
Don't pull.

I know it probably won't be, but some ADIOS2 bindings changed. Should just be one line of code and I wouldn't expect another break like this one. 

Chuck also suggested that there is a cleaner way to connect with cmake. I'll resubmit a PR when those are fixed.

cyrush:
@stevenwalton got it, i think we ultimately want a generic path to ADIOS 2 via conduit.  Maybe its something we can discuss?

stevenwalton:
I think that would be good. I have a meeting with @mclarsen on Monday. Maybe we could chat about it directly after?

cyrush:
@stevenwalton this week is quite busy, I am open Thursday afternoon & most of Friday.
Maybe email can we can setup a time to chat?

PR #152, opened 2018-09-14
Added Lagrangian filter information

ssane113:
Modifications to include Lagrangian filter.


ssane113:
@mclarsen any further changes?

mclarsen:
Did you fix the static counter?

ssane113:
No. That's a VTK-m change and I'm not sure what the alternative is right now.

PR #171, opened 2018-11-06
Correct type of cached CMake variables

slabasan:
- Removing trailing whitespace in CMake files.
- Correcting type of cached variables set in CMake. In particular, those that
  should be booleans with a value of OFF/ON.


cyrush:
@slabasan  I merged, thanks! I also updated our spack logic that generates host config files to leverage BOOL for on / off.

PR #181, opened 2018-11-21
update to latest (and presumably greatest) blt

cyrush:


cyrush:
This does fix issues with building shared vs static libs (at some point we travelled back in time with blt)

We are having problems with mfem + hypre in travis.  Don't think its related to this build, but want to build locally to explore.

cyrush:
@mclarsen 
there was a real issue with passing the blob of mfem flags to blt, i think i resolved -- will merge pending travis pass

PR #182, opened 2018-11-22
Task/param check

mclarsen:


mclarsen:
plots no longer have member 'params'. 'plots/p1/params/field' becomes 'plots/p1/field'. Its simpler and better, but this will break things.

mclarsen:
@cyrush what do you think? I converted vtkh filters as a first step.

src/ascent/runtimes/flow_filters/ascent_runtime_vtkh_filters.cpp
```diff
@@ -1137,12 +1204,30 @@ DefaultRender::verify_params(const conduit::Node &params,
...
+
+    std::string surprises = surprise_check(valid_paths, ignore_paths, params);
+
+
+    // parse render surprises
+    if(params.has_path("renders"))
     {
-        info["errors"].append() = "Missing required string parameter 'image_prefix'";
-        res = false;
+      const conduit::Node renders_node = params["renders"];
```
cyrush: found case w/ missed node ref (to avoid copy)

cyrush (approved):
looks good found a missed case where you could use a ref

PR #213, opened 2019-01-04
only report info and errors on rank 0

mclarsen:


src/ascent/ascent.cpp
```diff
@@ -274,8 +275,10 @@ Ascent::publish(const conduit::Node &data)
         else
         {
             // NOTE: CONDUIT_INFO could be muted, so we use std::cout
```
cyrush: these 'NOTE' comments can go

cyrush (commented):
Looks good, we can pull the NOTE comments about std:out b/c the no longer make sense?

src/ascent/ascent.cpp
mclarsen: done

cyrush (approved):

PR #223, opened 2019-01-07
pass quick start, adding tutorials and fixing typos

cyrush:
Docs updates aimed towards the tutorial.

Still needs some work, but feedback would be helpful.

Also, if there are any suggested small docs fixes, lets try to get them in as part of this request.


mclarsen (approved):
The quick start looks good and thanks for combatting my spelling deficiencies.

PR #252, opened 2019-02-15
fix config mk logic for mfem include

### LINKED ISSUE #248: Using with make example tries to include mfem make config without mfem installed
https://github.com/Alpine-DAV/ascent/blob/782e426de6e61248dfbbed137e6a7661c47605c9/src/config/ascent_config.mk.in#L136

cyrush:
resolves #248


mclarsen (changes requested):
The using with make example is disabled. Re-enable it to test this.

cyrush:
@mclarsen , will do -- wasn't sure if that was the only blocker, but we are about to find out.

mclarsen:
You are right. We are also not including the openmp links inside the config, which means we are also missing the cuda libs too. It will at least fail during linking.

mclarsen:
Shared builds worked just fine since OpenMP was rpathed in, but the static linking of the make example failed.

PR #263, opened 2019-03-08
rework cmake exports to better support find_package

cyrush:


cyrush:
almost there but not done testing yet ...

PR #267, opened 2019-03-12
Add ParaView vis. using Python Extracts

danlipsa:


danlipsa:
@cyrush @mclarsen @utkarshayachit Please review.

cyrush:
@danlipsa, i think this looks good!

I have one suggested change which would be helpful for future understanding.  I see you use json to name many variables in the conduit to vtk binding code.  The stuff passed in isn't really json, it's a Conduit Node -- that may confuse folks into thinking it's consuming json strings or using some python json api. 

Would it be hard to change the names to refer to conduit instead of json?

danlipsa:
@cyrush Sounds good. Indeed, that variable name can be confusing. I'll make the change.

danlipsa:
@cyrush I made the change suggested.

cyrush:
Looks good thanks!

PR #279, opened 2019-04-29
Update

mclarsen:


cyrush:
@FYI: Docker builds via Azure are working, we may have to pivot there.

PR #286, opened 2019-05-29
add support for specifying actions with yaml

cyrush:


cyrush:
not perfect, but my first pass that includes a test for passing actions via a yaml file

mclarsen:
Looking good.

mclarsen:
What does ` "<<UNSET>>"` mean?

cyrush:
i set it to a goofy  `<<UNSET>>` default value to indicate that we should check for either `ascent_actions.json` OR `ascent_actions.yaml`. There are probably better ways to plum this

PR #289, opened 2019-05-30
allow subsets of mpi tasks without data

cyrush:


cyrush:
also trying to fix an issue related to crazy striding

src/ascent/runtimes/ascent_vtkh_data_adapter.cpp
```diff
@@ -245,8 +292,7 @@ vtkm::cont::Field GetField(const conduit::Node &node,
 
   int num_vals = node.dtype().number_of_elements();
 
-
-  const T *values_ptr = node.value();
+  const T *values_ptr = values_ptr = node.value();
```
mclarsen: This seems incorrect?

src/ascent/runtimes/flow_filters/ascent_runtime_blueprint_filters.cpp
```diff
@@ -153,24 +158,51 @@ BlueprintVerify::execute()
...
+    if(!n_input->dtype().is_empty())
     {
-        n_input->print();
-        v_info.print();
-        ASCENT_ERROR("blueprint verify failed for protocol"
-                      << protocol << std::endl
-                      << "details:" << std::endl
-                      << v_info["message"].as_string());
-                      //<< v_info.to_json());
+        if(!conduit::blueprint::verify(protocol,
```
mclarsen, line 172: I am not sure that this is the correct place to do this. If we have an extract that takes `source` then verify is never called. Two choices: 1) always call verify as a part of `source` or move this logic into `publish`. Are we supporting the case where a rank might not call publish?

mclarsen: There are some other checks in publish like consistent domain ids, so if it was not called on every rank then there would be a MPI hang.

cyrush: just to circle back: we aren't supporting cases where publish or execute aren't called on some subset of MPI tasks. All MPI tasks need to call to coordinate. It's ok if some publish empty data. All need to execute the same actions.

src/ascent/runtimes/ascent_vtkh_data_adapter.cpp
cyrush: i'll fix this

PR #297, opened 2019-05-31
python 3 fixes

cyrush:
fixes for our python interp to support Python 3
add python 3 testing in travis


cyrush:
still WIP

cyrush:
I think this is ready to merge, however our CI checks are degraded.

Travis is reporting false positives, most cases are failing b/c they can't fetch metis in support of mfem.

PR #306, opened 2019-06-12
fixing copy and past error that results in surprises

mclarsen:
The default options file was being loaded as actions.

@khuck, this should solve your problem.

### DIFF
--- a/src/ascent/ascent.cpp
+++ b/src/ascent/ascent.cpp
@@ -325,7 +325,7 @@ Ascent::execute(const conduit::Node &actions)
 
                 if(!conduit::utils::is_file(m_actions_file))
                 {
-                    m_actions_file = "ascent_actionsyaml";
+                    m_actions_file = "ascent_actions.yaml";
                 }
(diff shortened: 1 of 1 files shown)


cyrush:
@mclarsen  yikes, sorry for that one.

src/ascent/ascent.cpp
```diff
@@ -321,11 +321,11 @@ Ascent::execute(const conduit::Node &actions)
...
             if(m_actions_file == "<<UNSET>>")
             {
-                m_actions_file = "ascent_options.json";
+                m_actions_file = "ascent_actions.json";
 
                 if(!conduit::utils::is_file(m_actions_file))
                 {
-                    m_actions_file = "ascent_options.yaml";
+                    m_actions_file = "ascent_actionsyaml";
                 }
```
cyrush, line 329: i think we are missing a . here

cyrush: i pushed a fix

PR #308, opened 2019-06-13
cuda fixes

cyrush:


mclarsen:
Lesson: the github search bar is a liar and can't be trusted.

cyrush:
Just realized the uberenv install case actually always tests against develop. 
Thats confusing and not helpful, I am going to disable.

PR #324, opened 2019-07-10
Demo 5 Added

stevenwalton:
Added a more complicated demo for the tutorial. This is a python extract
that includes classes and definitions. This should be easier for someone
that is more familiar with python and less familiar with ascent.


stevenwalton:
Making note to fix deadlock issue from #332. Do not merge yet.

I understand there were some other things you wanted @mclarsen?

mclarsen:
closing as stale.

PR #325, opened 2019-07-12
[WIP] Task/new vtkm

mclarsen:


mclarsen:
@cyrush I tried `vtkh vtkm`. Are you saying that you think that the order matters?

cyrush:
Good work!

PR #334, opened 2019-08-02
change how conduit is imported in cmake

cyrush:
use proper package find to avoid not well understood cmake import issue on ubunutu 18


cyrush:
resolves #333

PR #339, opened 2019-08-06
update to new blt and hit less warnings with new cmake

cyrush:


mclarsen (approved):

cyrush:
I icked up one more change from blt develop, will merge post ci

Issue #338, opened 2019-08-06
rethink explicit execute + reset in actions


cyrush:
more and more, i think we should have a mode where we do an implicit execute + reset.
When things go wrong (omit one of these two for example) its quite hard to understand what is going on.

Explicit reset is there to allow us to reuse the pipelines, but that is a complex case not what we expect most folks to need out of the box

mclarsen:
One solution is to keep the actions node around inside the runtime, meaning it would represent the current assembled pipeline. When execute is called, we can check to see if the current actions and last actions match. If they do, them we just call execute on the graph, otherwise we can rebuilt the entire thing. It would be harder to pick and choose what we rebuild (ie, only changing some subset of the entire graph).

Another thing to think about is how the simulation is calling us. For example, some integrations keep the Ascent object around and others create one every time, but I guess that wouldn't effect to much.

mclarsen:
To be more clear: execute means "run these actions" and that would not require a explicit action of `execute` or `reset`.

PR #340, opened 2019-08-10
Task/keep graph

mclarsen:
Implicitly perform `execute` and `reset` and only rebuild graph if actions change.


mclarsen:
Resolves #338

cyrush (approved):

PR #344, opened 2019-08-13
Task/jupyter extract

SeifIbrahim:


cyrush:
looks great!

Here are a couple of small shifts we should do pre-merge:

mv ascent_bridge (the client) up to src/ascent/, name the dir ascent_jupyter_bridge

rename py_src/server/ (and the equiv mpi ver) 'py_src/bridge_kernel/' and wire in to reflect the changes

and change example.py to be called  server.py 


Add yourself to the contribs list:
https://github.com/Alpine-DAV/ascent/blob/develop/src/docs/sphinx/index.rst

Also can we add a link to the bridge kernel under:

https://github.com/Alpine-DAV/ascent/blob/develop/src/docs/sphinx/Licenses.rst

src/ascent/python/mpi/py_src/server/example.py
```diff
@@ -0,0 +1,151 @@
...
+        if render is None:
+          print("ERROR: no scenes found")
+        #keeps the defaults from info['actions']
+        render.update(transformation(info, *args, **kwargs))
+        #TODO this is temporary
+        render["image_name"] = "out_ascent_render_3d"
+        return info["actions"]
+
+    #TODO store MPIServer in global variable instead of passing it
+    #TODO why doesn't if rank==0 at the top work?
```
mclarsen: Is this still the case?

src/ascent/python/mpi/py_src/server/example.py
```diff
@@ -0,0 +1,151 @@
...
+
+            info = conduit.Node()
+            server_ascent.info(info)
+
+            data = message["code"]
+
+            if data["type"] == "transform":
+                call_obj = data["code"]
+
+                #TODO better way to do this so i don't have to import everything into this file
```
mclarsen: Is there a better way so you don't have to import everything into this file?

mclarsen (approved):
There are a number of TODOs. Are they on the wish list or are any of them on the critical path? Its ok if they are not on the critical path.

The camera helpers like `rotate` are defined in two different files. Can they be merged into a single file?

mclarsen:
One item I forgot: there are a bunch of notebook files called `untitled`. Should they get names?

src/ascent/python/mpi/py_src/server/example.py
SeifIbrahim: The problem I was having with this was that I am sending over a string from the client to tell the server what function it should run and with what arguments. I was sending a string like 'numpy.pi/2' that was needing to be evaled before being passed as an argument, but in order to eval it I had to import numpy into this file. What I might end up doing is instead of evaling the arguments in the 'dispatcher' that calls the function, I'll probably leave more of that functionality up to the function.

SeifIbrahim:
> There are a number of TODOs. Are they on the wish list or are any of them on the critical path? Its ok if they are not on the critical path.
> 
> The camera helpers like `rotate` are defined in two different files. Can they be merged into a single file?

The TODOs are mostly things that I want to be working eventually (they're mostly related to my trackball widget working nicely) but aren't super critical.

All the code for the bridge kernel server is duplicated at the moment since we have MPI and non-MPI  versions for the ascent python module. We might end up merging them into a module separate from the ascent module.

src/ascent/python/mpi/py_src/server/example.py
SeifIbrahim: I think I was experiencing some weird behaviour with openmpi when I wrote that comment, we eventually switched to mpich for testing.

SeifIbrahim: Actually it turns out this is still an issue, if I check rank==0 at the top the function doesn't seem to run at all.

Issue #305, opened 2019-06-11
ascent.mk not correct when vtk-h is turned off


cyrush:

cyrush:
```
/home/vsts/work/1/s
g++ -I /home/vsts/work/1/s/uberenv_libs/spack/opt/spack/linux-ubuntu16.04-x86_64/gcc-5.5.0/conduit-master-aeq2tlvdvsqwgdke2v4sdiywrtegl6la/include/conduit   -I/home/vsts/work/1/s/uberenv_libs/spack/opt/spack/linux-ubuntu16.04-x86_64/gcc-5.5.0/hdf5-1.8.21-xguzlrjck6livhpvqfzq7xjxu7dvc2n4/include -I /home/vsts/work/1/s/install/include/ascent  ascent_render_example.cpp -Wl,-rpath,/home/vsts/work/1/s/install/lib -Wl,-rpath,/home/vsts/work/1/s/uberenv_libs/spack/opt/spack/linux-ubuntu16.04-x86_64/gcc-5.5.0/conduit-master-aeq2tlvdvsqwgdke2v4sdiywrtegl6la/lib   -Wl,-rpath,/home/vsts/work/1/s/uberenv_libs/spack/opt/spack/linux-ubuntu16.04-x86_64/gcc-5.5.0/hdf5-1.8.21-xguzlrjck6livhpvqfzq7xjxu7dvc2n4/lib      -L /home/vsts/work/1/s/install/lib -lascent -lascent_flow -lrover -L /lib -lvtkh_rendering -lvtkh_filters -lvtkh_core -lvtkh_lodepng -lvtkh_utils -L /lib -lvtkm_cont- -lvtkm_rendering- -L /home/vsts/work/1/s/uberenv_libs/spack/opt/spack/linux-ubuntu16.04-x86_64/gcc-5.5.0/conduit-master-aeq2tlvdvsqwgdke2v4sdiywrtegl6la/lib -lconduit_blueprint -lconduit_relay -lconduit   -L /home/vsts/work/1/s/uberenv_libs/spack/opt/spack/linux-ubuntu16.04-x86_64/gcc-5.5.0/hdf5-1.8.21-xguzlrjck6livhpvqfzq7xjxu7dvc2n4/lib -lhdf5  -ldl -lrt -pthread    -fopenmp -o ascent_render_example
/usr/bin/ld: cannot find -lrover
/usr/bin/ld: cannot find -lvtkh_rendering
[...]

Issue #335, opened 2019-08-03
check openmp cmake usage for consistency   


cyrush:
issue so I don't forget to look at this in the future. 

in ascent lib cmake lists:
```
if(ASCENT_VTKM_ENABLED)
    set(ASCENT_VTKM_USE_CUDA ${CUDA_FOUND})
    set(ASCENT_VTKM_USE_OPENMP ${ENABLE_OPENMP})
endif()


set(ASCENT_USE_OPENMP  ${OPENMP_FOUND})
```

cyrush:
we should be using `OPENMP_FOUND`, I'll audit soon.
There are places where we use ZZZ_ENABLE_OPENMP in compiler defines -- that's fine. Just want to be consistent about our CMake If tests

cyrush:
resolved in #346

PR #346, opened 2019-08-14
resolves a few odds and ends

### LINKED ISSUE #317: no error when python extract is handed bad file name
It appears passing a bad file name results in executing a python extract with no code - which will always succeeded. Confusing to debug.

cyrush:
resolves #317 #335


cyrush:
also resolves #305

PR #349, opened 2019-08-15
adding new vtkm worklet lib to make file configuration

mclarsen:


cyrush:
only failure is a fetch during spack build, i think this is good to merge

PR #354, opened 2019-08-18
adding a default build type

mclarsen:


mclarsen:
I ran into a vtkm assert today and I remember that @cyrush said that there was a failure in a test for the same reason. First, this shouldn't happen, i.e. there should not be asserts in the build. What is happening is that the build is defaulting to debug, and the vtkm asserts remain. Thus, I am adding a default build type.

This is just masking the failure, as we should not be tripping asserts anyway. I will track this down, but we should consider adding an all debug all the time version of the build to test for Ascent + vtkh + vtkm failures to fail early rather than when I am not expecting it in a release build.

PR #357, opened 2019-08-20
Add ENABLE_SERIAL option, on by default.

danlipsa:
This allows users to build only ascent_mpi
(by setting ENABLE_SERIAL) to off) to avoid issues
when building ascent that links with vtkm with MPI enabled.

### DIFF
--- a/src/examples/proxies/lulesh2.0.3/CMakeLists.txt
+++ b/src/examples/proxies/lulesh2.0.3/CMakeLists.txt
@@ -67,7 +67,7 @@ else()
    set(lulesh_openmp_flags "")
 endif()
 
-if (ENABLE_SER)
+if (ENABLE_SERIAL)
     blt_add_executable(
         NAME        lulesh_ser
         SOURCES     ${LULESH_SOURCES}
--- a/src/examples/proxies/lulesh2.0.3/CMakeLists.txt
+++ b/src/examples/proxies/lulesh2.0.3/CMakeLists.txt
@@ -67,22 +67,24 @@ else()
    set(lulesh_openmp_flags "")
 endif()
 
-blt_add_executable(
-    NAME        lulesh_ser
-    SOURCES     ${LULESH_SOURCES}
-    DEPENDS_ON  ${lulesh_deps}
-    OUTPUT_DIR ${CMAKE_CURRENT_BINARY_DIR})
-
-
-blt_add_target_compile_flags(TO lulesh_ser FLAGS "-DUSE_MPI=0 ${lulesh_openmp_flags}")
-
-# install target for lulesh serial
-install(TARGETS lulesh_ser
-        EXPORT  ascent
-        LIBRARY DESTINATION examples/ascent/proxies/lulesh
-        ARCHIVE DESTINATION examples/ascent/proxies/lulesh
-        RUNTIME DESTINATION examples/ascent/proxies/lulesh
-)
+if (ENABLE_SER)
+    blt_add_executable(
+        NAME        lulesh_ser
+        SOURCES     ${LULESH_SOURCES}
+        DEPENDS_ON  ${lulesh_deps}
+        OUTPUT_DIR ${CMAKE_CURRENT_BINARY_DIR})
+    
+    
+    blt_add_target_compile_flags(TO lulesh_ser FLAGS "-DUSE_MPI=0 ${lulesh_openmp_flags}")
+    
+    # install target for lulesh serial
+    install(TARGETS lulesh_ser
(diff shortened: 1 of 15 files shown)


src/examples/proxies/lulesh2.0.3/CMakeLists.txt
```diff
@@ -67,22 +67,24 @@ else()
...
-blt_add_target_compile_flags(TO lulesh_ser FLAGS "-DUSE_MPI=0 ${lulesh_openmp_flags}")
-
-# install target for lulesh serial
-install(TARGETS lulesh_ser
-        EXPORT  ascent
-        LIBRARY DESTINATION examples/ascent/proxies/lulesh
-        ARCHIVE DESTINATION examples/ascent/proxies/lulesh
-        RUNTIME DESTINATION examples/ascent/proxies/lulesh
-)
+if (ENABLE_SER)
```
cyrush: should this be ENABLE_SERIAL?

src/tests/flow/CMakeLists.txt
```diff
@@ -50,31 +50,32 @@
...
-foreach(TEST ${FLOW_TESTS})
-    add_cpp_test(TEST ${TEST} DEPENDS_ON ascent_flow ascent)
-endforeach()
-
-if(PYTHON_FOUND AND ENABLE_PYTHON)
-    add_subdirectory("python")
-else()
-    message(STATUS "Python disabled: Skipping ascent python module tests")
+if (ENABLE_SERIAL)
+    set(FLOW_TESTS  t_flow_data
```
cyrush: In theory these don't need to depend on ENABLE_SERIAL, flow is always a serial lib & is built for use in both non-mpi and mpi-cases, it does not link to any VTK-h or VTK-m stuff. 

I do see we have ascent on as a dep, I think that is a mistake -- we should remove it as a dep.

src/ascent/python/CMakeLists.txt
```diff
@@ -42,62 +42,64 @@
...
-
-# TODO: ADD SUPPORT FOR SUPPORT PYTHON_MODULE_INSTALL_PREFIX
-# install the capi header so other python modules can use it
-# support alt install dir for python module via PYTHON_MODULE_INSTALL_PREFIX
-# if(PYTHON_MODULE_INSTALL_PREFIX)
-#     install(FILES ${ascent_py_headers} DESTINATION ${PYTHON_MODULE_INSTALL_PREFIX}/ascent/)
-# else()
-#     install(FILES ${ascent_py_headers} DESTINATION python-modules/ascent/)
-# endif()
+if (ENABLE_SERIAL)
```
cyrush: on develop we just changed how the python modules are created, there is one module that supports both the non-mpi and mpi case. 

You will have to change things a bit to get what you  are aiming for.

I think you can break out the non-mpi case to look like the mpi case - where you use a marco  to create only the compiled part of the python module.

https://github.com/Alpine-DAV/ascent/blob/4b5cf30a39690e2aa8d843edb38a7da376c984b1/src/ascent/python/ascent_module/CMakeLists.txt#L78

You will then need to modify current hybrid python module marco call to just be the macro call that uses distutils.

Additionally you can add error messages when the non-mpi compiled module fails to import in
https://github.com/Alpine-DAV/ascent/blob/develop/src/ascent/python/ascent_module/py_src/ascent.py

that mirror what is in:

https://github.com/Alpine-DAV/ascent/blob/develop/src/ascent/python/ascent_module/py_src/mpi/ascent_mpi.py

cyrush (commented):
a few things to resolve but overall looks good, let me know if you have questions

danlipsa:
Thanks @cyrush I was looking to resolve the conflict with develop. I'll address your comments as well and push a new version.

src/examples/proxies/lulesh2.0.3/CMakeLists.txt
danlipsa: Done. Indeed, this was a typo.

src/tests/flow/CMakeLists.txt
danlipsa: Yes, I was able to do that.

src/ascent/python/CMakeLists.txt
danlipsa: Not sure I did this, lets see what the tests do.

danlipsa:
@cyrush Can you take a look at this new version? Thanks!

cyrush (approved):
thanks @danlipsa!

there are still a few more python tweaks that need to be done -- but i'll tackle them in another PR.

Issue #367, opened 2019-08-30
enable serial isn't propagated correctly to our exported targets


cyrush:
This results in `ascent::ascent` from not being defined. 

We need to create an `ASCENT_ENABLE_SERIAL` cmake var, and then update the exported target logic in ascent_setup_targets.cmake to use this to guard creation of `ascent::ascent`

cyrush:
resolved in #368

PR #368, opened 2019-08-30
cori auto-build update and cmake exported target fixes

cyrush:
* update with latest compilers for nersc cori 
* add missing install of ascent.mod when fortran is on
* fix logic for creating ascent::ascent when ENABLE_SERIAL  == ON


cyrush:
resolves #367

PR #377, opened 2019-09-16
add sanity check for blt time travel

cyrush:


cyrush:
adds ci check to defend against accidental time travel with blt submodule
fix an instance of accidental time travel

PR #384, opened 2019-09-23
test path for relay extract with no cycle

mclarsen:


mclarsen:
@cyrush The test status is weird. Any idea why its not showing green? All the tests have passed.

PR #394, opened 2019-10-24
fix error with stats test

cyrush:


cyrush:
also fix output from cmake git message

Issue #395, opened 2019-10-25
need to be able to explicitly set the image name for a render


cyrush:
i tried "image_name", but that didn't seem to work.

We need a path were the user set the full file name. 

@mclarsen I feel like I am overlooking something, am I missing the right incantation?

mclarsen:
Do you care if its overwritten each time?

cyrush:
in this case thats exactly what I want (allow user to set fully control the filename)

cyrush:
resolved in #397

Issue #385, opened 2019-09-25
add info about if mfem is on to ascent::about


cyrush:
we can add to the ascent runtime info

cyrush:
resolve in #397

PR #397, opened 2019-10-30
tutorial reorg

cyrush:
will also address #385


mclarsen:
One more thing we need to do is update the vtkm spack package + update the hashes so we are using the vtkm 1.5 release. Also, test with cuda. I can do a build tonight.

mclarsen (approved):
I looked through the 175 files and 49 commits. Looks good to me

cyrush:
@mclarsen thanks! I still  have a few docker tweaks, then I'll merge. 
I suggest we use another branch to do the spack updates, this one is getting big.

mclarsen:
resolves #395

PR #401, opened 2019-11-06
more tutorial prep

cyrush:


cyrush:
@mclarsen i think this is the last major bunch of changes to support the tutorial examples and docs

PR #439, opened 2020-01-08
BabelFlow and feature extraction integration

spetruzza:
Additions:
- runtime support (BabelFlow, with default configuration using MPI)
- flow filter to perform topological segmentation using the parallel merge tree algorithm (depends and use BabelFlow as execution runtime)
- unit test for new filter

Uberenv/spack:
- new compiler config for OSX catalina
- new dependencies packages scripts (babelflow and pmt)
- not yet working as a spack variant (need some guidance)


mclarsen:
@spetruzza I haven't fully looked this over, but one thing I initially noticed is that DIY is included. This can conflict with down stream versions. Sensei uses DIY and they use ascent, so this can cause a conflict. Additionally, vtkh and vtkm both have DIY as well. In those libs, we have namespaced diy so they don't bump into each other, so I would suggest doing the same.

spetruzza:
Thanks Matt for noticing that, there was still a reference (in a CMake file) to the DIY libraries that we purposely removed as a dependency for this integration. I removed that reference.

cyrush:
`[exe: spack/bin/spack install uberenv-ascent%gcc+mpi+mfem ^python@3.6.3+babelflow ^mpich]`

In the azure yaml: I think you simply need to add the if babelflow case after the mfem case, and before the python case, that way you end up with:

`[exe: spack/bin/spack install uberenv-ascent%gcc+mpi+mfem+babelflow  ^python@3.6.3^mpich]
`

scripts/uberenv/packages/ascent/package.py
```diff
@@ -0,0 +1,504 @@
...
+    #######################
+    # MPI
+    #######################
+    depends_on("mpi", when="+mpi")
+    depends_on("py-mpi4py", when="+mpi+python+shared")
+
+    #######################
+    # BabelFlow
+    #######################
+    depends_on('babelflow@develop', when='+babelflow')
```
mclarsen:
```
when='+babelflow+mpi'
```
We don't want to depend on MPI or have spack conflicts ~mpi+mpi

scripts/uberenv/packages/ascent/package.py
```diff
@@ -0,0 +1,504 @@
...
+
+class Ascent(Package, CudaPackage):
+    """Ascent is an open source many-core capable lightweight in situ
+    visualization and analysis infrastructure for multi-physics HPC
+    simulations."""
+
+    homepage = "https://github.com/Alpine-DAV/ascent"
+
+    # TODO: revert this to the official repo when merged
+    git      = "https://github.com/xuanhuang1/ascent.git"
```
mclarsen: This should be reverted before merging, since this is going to be merged.

scripts/uberenv/packages/ascent/package.py
```diff
@@ -0,0 +1,504 @@
...
+
+    version('babelflow',
```
mclarsen: This should be like whatever is in the main ascent package.

mclarsen (changes requested):
There are some minor tweaks to the ascent package. Once this merges, we will move this package back into the spack fork. There are some other small cmake chages that we can fix later, but we can merge this soon

cyrush:
pending matt's suggested ascent package changes  -- this looks good to merge!

There are a few small improvements that can be applied to simplify how the linking and flags are specified -- but I can tackle those after this PR is merged.

mclarsen (approved):

PR #444, opened 2020-01-19
update cmake python logic to allow build against statically linked python

cyrush:


mclarsen:
Do downstream consumers need to link python in the dynamic case or does this only apply in the static case?

cyrush:
they may still need to, but I think we can fix with more work to the exported cmake targets and the config-mk -- lets not merge yet

cyrush:
yeah -- ci confirms still things to iron out.

mclarsen:
Don't forget about the make file.

cyrush:
closing since our CMake logic has diverged too much, see #977 for what we should salvage in the future.

PR #446, opened 2020-01-23
python 3.7 fix related to refs for flow python 

cyrush:


cyrush:
This solves the python 3.7 crash.

not 100% sure this is the end of story here - I'll like to do a detailed ref count trace to make sure we aren't leaving any hanging around.

PR #450, opened 2020-01-30
Consistent data object state

mclarsen:


src/ascent/runtimes/ascent_data_object.cpp
```diff
@@ -66,6 +66,15 @@
 namespace ascent
 {
 
+DataObject::DataObject()
+  : m_low_bp(nullptr),
+    m_high_bp(nullptr),
+    m_vtkh(nullptr),
```
cyrush, line 73: one more vtk-h guard?

cyrush: i pushed a change with the guard

PR #457, opened 2020-02-05
support exa prefix style filters in pipelines

cyrush:


mclarsen (approved):
We are ready for exascale.

PR #462, opened 2020-02-14
Dray

mclarsen:


mclarsen:
@cyrush I can't get this last test to pass, and I haven't been able to duplicate. The label says this is a static build, but the spack spec indicates ``+shared``

I was looking at the mk.in files and it seems like the ``ASCENT_LINK_RPATH``, and the devil ray one I modeled after it, don't actually use the rpath link. Is that intended?

PR #472, opened 2020-03-12
add install of web client supporting files

cyrush:
add install of web client supporting files for ascent and cinema clients
add dynamic discovery of web client root, between install and source locs


cyrush:
Fixes the fact that we were not installing the web client stuff, but actually were relying on the files from the source tree

mclarsen (approved):
download

PR #474, opened 2020-03-17
cuda ci test: try Release build to minimize disk space

cyrush:


mclarsen:
It finished. No idea how close to the edge we are.

PR #490, opened 2020-03-26
summit build script and docs about summit public install

cyrush:


cyrush:
we can skip ci on this one

PR #492, opened 2020-03-29
update to use new spack fork based on develop as of 2020-03

cyrush:
wip -- working on testing. 

Probably a good idea to merge after our next release to hedge.


cyrush:
closing this as old, will need to start the process again with newer spack

PR #501, opened 2020-04-03
remove things that are no longer used

mclarsen:
@cyrush This will reduce the confusion of people trying to use old things that are no longer maintained/used.


cyrush:
Sounds good, but can we keep config-build? (or just move into scripts?)
I do use this quite often.

mclarsen:
Doesn't config-build use the host-configs?

cyrush:
you can pass it a cmake cache file as the first arg - thats the mode that I always use it in.

cyrush:
./config-build.sh uberenv_libs/hostname.cmake

cyrush (approved):
:-)

PR #507, opened 2020-04-08
add num_files param to relay extract

### LINKED ISSUE #483: add baton based many task to less files aggregation for relay extracts

cyrush:
resolves #483


mclarsen:
I pushed the amr test to develop by accident. I removed the image check so the tests will pass now.

src/cmake/thirdparty/SetupConduit.cmake
```diff
@@ -93,7 +93,7 @@ endif()
 set(CONDUIT_FOUND TRUE)
 set(CONDUIT_INCLUDE_DIRS ${CONDUIT_DIR}/include/conduit)
 
-if(PYTHON_FOUND)
+if(ENABLE_PYTHON)
```
mclarsen, line 96: Just a note: i don't think we are consistent in `ENABLE_X`  versus `FOUND_X`

mclarsen (approved):
3s30go(1)

src/cmake/thirdparty/SetupConduit.cmake
cyrush: I agree, in this case, I need to change b/c it was being stubborn, I was mixing builds of TPLS. Since not being able to override caused an error, I decided that we need the extra control of ENABLE_PYTHON

PR #511, opened 2020-04-10
Update BabelFlow filter for task based feature extraction

spetruzza:
This PR contains several updates to the BabelFlow runtime and filter integration, in particular:
- "babelflow" extract is now a filter
- support for blueprint format as input and output of the filter
- simplified parametrization of the filter, now domains/patch bounding boxes are computed internally and have not to be provided by the user/application
- new parameters to handle patches/domain ghost cells of different size
- new parameter to produce segmentation field
- new parameter to run the analysis on a specific grid (if more uniform grids are present)
- updated auto test
- updated example adding also rendering pipeline of the segmentation field


spetruzza:
Current failure seems not to be related to this PR specifically, but rather to the "t_ascent_dray" test and the cloverleaf build

mclarsen:
@spetruzza I am fine with merging this if you want me to. The cuda builds are OOM, and I will track down the dray failure separately.

spetruzza:
Thank you @mclarsen, I guess we can merge then.

PR #512, opened 2020-04-15
Task/field filtering

### LINKED ISSUE #509: Field Filtering
Copying data is inevitable and we should provide a path to filter out fields that are not needed. Normally this is not a big deal, but some simulations have 300+ variables. When only plotting one, we don't need to bring the other 299 with us. 

Field filtering would be an optional mode that is off by default.

`task/field_filtering`  is an initial pass at detecting what fields we actually need.

Current gaps:
- Blueprint relay: by default, we save out the entire mesh and all variables, unless a list is provided.
- Scalar rendering: save all scalar fields
- Expressions: its possible to parse the expressions and see what fields are needed.
- Something I haven't thought of yet: i don't know what this is.

Relay and scalar rendering require thought. 
Options:
- Error if one of these two filters is in the actions when filtering is on and no field list is provided.
- Panic. Use all variables.

mclarsen:
resolves #509


mclarsen:
Don't filter the  original source: keep original data.

mclarsen:
I investigated the dray test that was failing on and off for different pipelines. After downloading the images and doing the diff myself, it totally passed and was not close to diff report. Not sure why this fails, but I added color tolerance to the image comparison in addition to the total pixel tol.

PR #520, opened 2020-05-01
change install dest for session converters [skip ci]

cyrush:
small tweak to install python to make sure it lives under the ascent dir like other things.


cyrush:
@mclarsen i assume we want under utilities instead of examples, but in either case we need the ascent dir in the path. If you want examples thats fine, just let me know.

mclarsen (approved):

PR #524, opened 2020-05-14
add perf test script

cyrush:


cyrush:
@mclarsen I think we can merge this one? I can check it vs new azure setup.

mclarsen:
go for it

PR #540, opened 2020-06-17
CMake Config: Install Destination

ax3l:
Correct the CMake Config package's Install prefix. Seen on Summit.

Refs.:
- https://cmake.org/cmake/help/v3.9/command/find_package.html


mclarsen:
I think the only other thing that has to change is our using-with-cmake example to update where its looking for the config.

mclarsen:
@cyrush Do you agree?

cyrush:
We may need to change the using-with-cmake project.  Let's see what happens in CI.

I did both "old (cyrus bad way) and new (expected way)" for conduit as a hedge:
https://github.com/LLNL/conduit/blob/420c69805942a77c10fa29f773f101eb61793f04/src/CMakeLists.txt#L169

ax3l:
I pushed a suggestion for the example, but rely on CI for testing.

https://github.com/LLNL/conduit/blob/420c69805942a77c10fa29f773f101eb61793f04/src/CMakeLists.txt#L171-L173

I would remove them tbh, break it once instead of having people confused :) The old dir location can also lead to problems with includes, etc. that trigger even more work-arounds downstream.

src/examples/using-with-cmake/ascent_render_example.cpp
```diff
@@ -48,11 +48,10 @@
...
 //-----------------------------------------------------------------------------
 
-#include <iostream>
-
-#include "ascent.hpp"
+#include <ascent.hpp>
+#include <conduit/conduit_blueprint.hpp>
 
-#include "conduit_blueprint.hpp"
+#include <iostream>
```
ax3l, line 54: recommended include order, e.g. std libs last (see e.g. [LLVM guide](https://llvm.org/docs/CodingStandards.html#include-style))

src/examples/using-with-cmake/ascent_render_example.cpp
```diff
@@ -48,11 +48,10 @@
 ///
 //-----------------------------------------------------------------------------
 
-#include <iostream>
-
-#include "ascent.hpp"
+#include <ascent/ascent.hpp>
+#include <conduit/conduit_blueprint.hpp>
```
ax3l, line 52: are those the official includes?

Makes this also a more robust `<..>` include in dual-install environments (e.g. system install + newer pkg manager install).

src/examples/using-with-cmake/CMakeLists.txt
```diff
@@ -73,27 +73,10 @@ cmake_minimum_required(VERSION 3.9)
...
-endif()
-
-#
-# Check for valid Ascent install
-#
-get_filename_component(ASCENT_DIR ${ASCENT_DIR} ABSOLUTE)
-if(NOT EXISTS ${ASCENT_DIR}/lib/cmake/AscentConfig.cmake)
-    MESSAGE(FATAL_ERROR "Could not find Ascent CMake include file (${ASCENT_DIR}/lib/cmake/AscentConfig.cmake)")
-endif()
-
```
ax3l, line 90: removed since sub-ideal in a _"copy-from-here"_ example

src/examples/using-with-cmake/CMakeLists.txt
```diff
@@ -51,19 +51,19 @@
...
+# In order to run directly in a sub directory below using-with-cmake in an ascent install, 
+# set Ascent_DIR to ../../..
 # 
 #  mkdir build
 #  cd build
-#  cmake ..
+#  cmake .. -DAscent_DIR=../../..
 #  make
 #  ./ascent_render_example
 #
```
ax3l, line 69: I feel this block would be more visible in a README.md or INSTALL.md file in the same directory, but that's taste.

src/examples/using-with-cmake/CMakeLists.txt
cyrush: one issue -- the default logic allows this to work out of the box for folks that install themselves, which is valuable for learning the tool on say their laptop. 

I don't think it precludes `copy-from-here`,  it just gives you an error in the case it's not set as needed?

cyrush: or does this undermine when you set something like the cmake module path?

src/examples/using-with-cmake/ascent_render_example.cpp
cyrush: internally, parts of the libs don't include using the prefix style (conduit/conduit_more_specific_header_zzz.hxx)

so for installs, we provide both styles of include paths either via cmake or our config.mk, both styles being a setup that enables both:

#include <conduit_blueprint.hpp>
#include <conduit/conduit_blueprint.hpp>

All of the headers are also prefixed with ascent_ (or conduit_), so they wont collide with other products (restated: nothing like utils.hxx exists to collide)

src/examples/using-with-cmake/CMakeLists.txt
cyrush: this gets pulled into the docs here:

https://ascent.readthedocs.io/en/latest/QuickStart.html#using-ascent-in-your-project

which is where I expect most folks to discover it.  Also folks cut-and-paste and hack away, so having it in the file they are editing is useful.

src/examples/using-with-cmake/ascent_render_example.cpp
ax3l: Sounds good. Since this is an example where people copy-paste from and since there is already a modern prefix dir (even though there is also a prefix in each name) I'll keep it at `<ascent/ascent.hpp>` and similar for conduit :)

src/examples/using-with-cmake/CMakeLists.txt
ax3l: Yes, the `if(NOT Ascent_DIR)` logic would overwrite an environment hint such as `CMAKE_PREFIX_PATH` or `Ascent_DIR` exports.

src/examples/using-with-cmake/CMakeLists.txt
ax3l: Then it's totally cool I think :)

src/examples/using-with-cmake/CMakeLists.txt
cyrush: Idea: I still want this to work in the install dir right after install --  how about (instead of being so strict) check

if 
`NOT ASCENT_DIR)` and `../../..//lib/cmake/ascent/AscentConfig.cmake` exists
set Ascent dir = ../../../
?

ax3l: ~~I think we could set the `HINTS` path instead of `PATHS`~~ Okay, `PATHS` hints might be the way to go, they have the sufficiently (low) priority if we do not set `NO_DEFAULT_PATH`: https://cmake.org/cmake/help/v3.18/command/find_package.html

cyrush (approved):

PR #547, opened 2020-07-01
update pubs, add missing uber activate for sphinx-rtd-theme

cyrush:


cyrush:
@mclarsen anything else missing from pubs?

PR #551, opened 2020-07-08
Task/ecf

### LINKED ISSUE #545: not right
https://github.com/Alpine-DAV/ascent/blob/28b6b51bfeb186c77d38599f082f823c0f7a8d22/src/ascent/runtimes/expressions/ascent_blueprint_architect.cpp#L227

SeifIbrahim:


src/ascent/runtimes/expressions/ascent_blueprint_architect.hpp
```diff
@@ -111,18 +111,32 @@ conduit::Node field_entropy(const conduit::Node &hist);
...
+
+conduit::Node binning(const conduit::Node &dataset,
+                  conduit::Node &bin_axes,
+                  const std::string &reduction_var,
+                  const std::string &reduction_op,
+                  const double empty_bin_val);
+
+void ASCENT_API paint_binning(const conduit::Node &binning, conduit::Node &dataset);
+
+conduit::Node ASCENT_API binning_mesh(const conduit::Node &binning);
```
cyrush: should the output be passed in as a ref? (If not the entire result will be copied)

src/tests/ascent/t_ascent_expressions.cpp
```diff
@@ -44,628 +44,939 @@
...
+  expr = "binning('field', 'max', [axis('z', [-5, 0, 5])])";
+  res = eval.evaluate(expr);
+  EXPECT_EQ(res["attrs/value/value"].to_json(), "[3.0, 0.0]");
+
+  expr = "binning('field', 'max', [axis('z', [-5, 0, 5], clamp=True)])";
+  res = eval.evaluate(expr);
+  EXPECT_EQ(res["attrs/value/value"].to_json(), "[3.0, 7.0]");
+
+  expr =
+      "binning('field', 'max', [axis('x', num_bins=4), axis('y', num_bins=4)], "
```
cyrush, line 696: I like how this turned out! 
Q: Is there a way to do something like:

`axis('x", num_bins=4, min=0, max = 1)`

src/tests/ascent/t_ascent_expressions.cpp
```diff
@@ -44,628 +44,939 @@
...
+  {
+    multi_dom.child(i)["state/cycle"] = 100;
+    multi_dom.child(i)["state/domain_id"] = 0;
+  }
+
+  runtime::expressions::register_builtin();
+  runtime::expressions::ExpressionEval eval(&multi_dom);
+
+  std::string expr;
+  expr = "binning('dist', 'std', [axis('x'), axis('y')], output='bins')";
```
cyrush: these cases use default # of bins?

cyrush (commented):
Looks good, one small suggestion and a few questions to help educate me :-)

src/tests/ascent/t_ascent_expressions.cpp
SeifIbrahim: Yes `axis('x', num_bins=4, min_val=0, max_val=1)`

src/tests/ascent/t_ascent_expressions.cpp
SeifIbrahim: If the number of bins is not specified it will try to calculate a reasonable number of bins. If the topo is uniform and the axis is x/y/z that comes from dim. For non-uniform meshes and x/y/z the number of bins is max_x - min_x. For fields it's just 256. Maybe it's better to just have the default be 256 everywhere in case it gets too big.

Min and max are also calculated if they aren't specified. For x/y/z it looks at all the fields and calculates the bounds from that.

src/ascent/runtimes/expressions/ascent_blueprint_architect.hpp
SeifIbrahim: Yes I will update that.

src/tests/ascent/t_ascent_expressions.cpp
cyrush: auto min and max is good, 

for the # of bins, a default # is easier to reason about. 

We don't want an unstructured astro code modeling the universe to allocate 1e+20 bins :-)

src/tests/ascent/t_ascent_expressions.cpp
SeifIbrahim: One thing I should say is that the bins won't be inclusive on max_val.

cyrush (approved):
Thanks for the changes!

PR #552, opened 2020-07-09
Bflow filter

sshudler:
PR for merging the BabelFlow compositing (bin-swap, radixk) extract


sshudler:
Need to solve the conflicts first

PR #556, opened 2020-07-10
fix mpi import

mclarsen:


cyrush (approved):

mclarsen:
We would have caught this with an using-with-cmake-mpi example.

cyrush:
#557

PR #559, opened 2020-07-10
Task/binning fixes

SeifIbrahim:


src/ascent/runtimes/expressions/ascent_blueprint_architect.cpp
```diff
@@ -1417,32 +1396,29 @@ binning(const conduit::Node &dataset,
...
-      res_bins[i] = bins[i];
-    }
-  }
-  else if(reduction_op == "pdf")
+  if(reduction_op == "pdf")
   {
-    int n = 0;
+    double total = 0;
 #ifdef ASCENT_USE_OPENMP
 #pragma omp parallel for reduction(+ : n)
```
mclarsen: There is no `n` since the previous line `int n = 0` seems to be gone.

mclarsen: probably should be `reduction(+ : total)`

SeifIbrahim: thanks!

src/ascent/runtimes/expressions/ascent_blueprint_architect.cpp
```diff
@@ -1417,32 +1396,29 @@ binning(const conduit::Node &dataset,
...
-    int n = 0;
+    double total = 0;
 #ifdef ASCENT_USE_OPENMP
-#pragma omp parallel for reduction(+ : n)
+#pragma omp parallel for reduction(+ : total)
 #endif
     for(int i = 0; i < num_bins; ++i)
     {
-      n += bins[i];
+      total += bins[2 * i];
```
mclarsen, line 1407: if `i` is num_bins, then won't `2 * i` be out of bounds?

cyrush: i assume there was over allocation and some fancy indexing? Maybe thats not the case. 
For the future, should we use separate arrays so the semantics are more clear?

SeifIbrahim: Yes that's right, the actual size of the array is bins_size. I am trying to put multiple values in the "same bin" for things like standard deviation which require me to keep track of 3 variables in the reduction (sum X, sum X^2, and n). I do it this way so that only one MPI call is necessary.

mclarsen: A suggested alternative.

```
int num_bins = 10;
int num_components = 3;
double * bin_array = new double[num_bins * num_components];
```

Being a bit more explicit can really help with the readability and intent.

SeifIbrahim: I do this here.
```  
 // number of variables held per bin (e.g. sum and cnt for average)
  int num_bin_vars = 2;
  if(reduction_op == "var" || reduction_op == "std")
  {
    num_bin_vars = 3;
  }
  const int bins_size = num_bins * num_bin_vars;
  double *bins = new double[bins_size]();

```
Would renaming the variables help?

PR #589, opened 2020-08-22
updates to ci containers

cyrush:


cyrush:
I think we are across the finish line, lets see how the tests go.

cyrush:
ready! take a look at the docs about the process and let me know if they make sense.

PR #598, opened 2020-09-05
adding ascent variables to turn-on/off 2d sampling filter

subhashis-lanl:
Adding ascent variable "use_gradient" to turn on/off 2d sampling feature


cyrush:
Closing since gradient related options aren't the VTK-h filters.
Let's revisit if that work picks up again.

PR #599, opened 2020-09-09
updating undocumented secret mfem/conduit conventions

mclarsen:


cyrush (approved):
for the nodes, thats actually a floating name. The topology should specify the exact name of which field corresponds to the grid function using `grid_function`.

https://llnl-conduit.readthedocs.io/en/latest/blueprint_mesh.html#optional-association-with-a-grid-function

the _nodes convention is the only one I have seen in used, but there could be more.

cyrush:
so that part wasn't secret or undocumented :-)

mclarsen:
There are "optimized_nodes" and "position". It seems to need both. If your missing one, mfem conduit data adapter just segfaults. If you are missing the other, it tells you.

mclarsen:
There are also "mesh_nodes" sometimes.

PR #602, opened 2020-09-19
Automatic camera

nicolemarsaglia:
Unit tests for 9 metrics.


nicolemarsaglia:
@cyrush It's finally green! When you get a chance can you review? No rush.

cyrush:
Nice work getting everything refactored!

User interface question:  Instead of a pipeline filter, could we support this in the scene params?

We already handle cinema as a special camera, could we do this same with this and auto add the auto cam filter when that case exists?

nicolemarsaglia:
> Nice work getting everything refactored!
> 
> User interface question: Instead of a pipeline filter, could we support this in the scene params?
> 
> We already handle cinema as a special camera, could we do this same with this and auto add the auto cam filter when that case exists?

Oooh yes, I like that more. I already was feeling wishy-washy about it being a filter and not an extract. But ya, I'll make that change.

nicolemarsaglia:
@cyrush I was able to get things moved over. Just did some minor edits, but it should go green. Let me know if I should make any other changes!

src/docs/sphinx/Actions/Scenes.rst
```diff
@@ -397,6 +397,74 @@ parameters:
...
+    scenes["s1/plots/p1/field"] = "radial";
+    scenes["s1/plots/p1/pipeline"] = "pl1";
+    
+    //camera knobs
+    scenes["s1/renders/r1/type"] = "auto_camera";
+    scenes["s1/renders/r1/auto_camera/metric"] = "dds_entropy";
+    scenes["s1/renders/r1/auto_camera/samples"] = 5;
+    scenes["s1/renders/r1/auto_camera/field"] = "radial";
+
+    scenes["s1/renders/r1/image_prefix"] = output_file;
```
cyrush, line 449: Looks great as part of the scene spec!

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rendering_filters.cpp
```diff
@@ -1218,10 +1230,77 @@ DefaultRender::execute()
...
 
-          vtkh::Render render = detail::parse_render(render_node,
-                                                     scene_bounds,
-                                                     image_name);
-          renders->push_back(render);
+	  if(is_auto_camera)
+	  { 
+            DataObject *source
+              = graph().workspace().registry().fetch<DataObject>("source_object");
+
```
cyrush, line 1237: Auto-cam selection is being run against the original data.

This will work well for many cases, longer term we should think about if we need to push auto cam params into the renderer, to get the actual data.

cyrush (approved):

cyrush:
Thanks for working all of this out, we are ready to merge!

PR #625, opened 2020-10-30
fix bug with graph build book keeping

cyrush:
resolves bug reported by @ax3l

where reuse of ascent leads to graph build errors due to old state hanging around.

### DIFF
--- a/src/ascent/runtimes/ascent_main_runtime.cpp
+++ b/src/ascent/runtimes/ascent_main_runtime.cpp
@@ -1392,7 +1392,12 @@ AscentRuntime::FindRenders(conduit::Node &image_params,
 void
 AscentRuntime::BuildGraph(const conduit::Node &actions)
 {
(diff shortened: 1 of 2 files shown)


src/ascent/runtimes/ascent_main_runtime.cpp
```diff
@@ -1392,7 +1392,12 @@ AscentRuntime::FindRenders(conduit::Node &image_params,
 void
 AscentRuntime::BuildGraph(const conduit::Node &actions)
 {
-  // exection will be enforced in the following order:
+  // make sure to clean helpers that we use
+  // to build the graph
+  m_connections.reset();
+  m_scene_connections.reset();
+
```
cyrush, line 1399: @mclarsen  -- are their other graph building related things we need to clear here?
These look like the only two to me, but maybe you can give it a quick look?

cyrush:
merging to build next container iteration from devleop

PR #631, opened 2020-11-05
improve bp index gen in relay extract 

cyrush:


cyrush:
@mclarsen can you give this a check?

PR #637, opened 2020-11-06
fix some things, add an about util

cyrush:


cyrush:
@mclarsen fingers crossed, this should have the right defense against the ENABLE_TEST issue in CI

cyrush:
ci is healty again

PR #650, opened 2020-11-11
prevent hang with bad actions file in parallel

### LINKED ISSUE #647: mpi hang when actions fail to parse

cyrush:
resolves #647


cyrush:
I reproduced the hang in a docker container and verified that this fix prevents it

cyrush:
@mclarsen can you give this a quick check?

Issue #642, opened 2020-11-07
add vim to ci docker containers


cyrush:
a nice to have / note to self so I can remember during a future update.

cyrush:
done in #657

PR #657, opened 2020-12-04
add vim to docker ci base images

cyrush:


cyrush:
resolves #642 + we will have vim after next tpl update

PR #666, opened 2021-01-06
Param expressions

mclarsen:
if this looks good then I can connect the expressions path to all filter parameters.


cyrush (approved):
Looks good, This is an awesome feature!

Before merge there are std:cout and the save of the flow html that should be commented out.

src/ascent/runtimes/flow_filters/ascent_runtime_param_check.cpp
```diff
@@ -87,11 +108,33 @@ check_numeric(const std::string path,
...
+    }
+
+    if(!params[path].dtype().is_number() && !is_expr)
+    {
+      if(supports_expressions)
+      {
+        std::string msg = "Numeric parameter '" + path +
+                          " : " + params[path].to_yaml()
+                             + "'  is not numeric and is not a valid expression";
+        info["errors"].append() = msg;
```
cyrush, line 132: is there a way to add the parse error, etc to this error message?

src/ascent/runtimes/flow_filters/ascent_runtime_param_check.cpp
```diff
@@ -232,6 +275,85 @@ path_helper(std::vector<std::string> &paths,
...
+  // check to see if this is an expression
+  if(node.dtype().is_string())
+  {
+    if(dataset == nullptr)
+    {
+      ASCENT_ERROR("Numeric parameter is an expression(string)"
+                   <<" but we can not evaluate the expression."
+                   <<" This is usaully for a parameter that is "
+                   <<"not meant to have an expression. expression '"
+                   <<node.to_string()<<"'");
```
cyrush, line 322: i guess we always need the dataset ptr, even if the expression doesn't invoke anything requiring it? 

Like: `1+1`

sure -- that seems like a goofy use case, but I could see folks wanting to use trig funcs, etc?

cyrush: if we want to do this, future PR would be best, just wanted to comment while thinking about it.

src/ascent/runtimes/flow_filters/ascent_runtime_vtkh_filters.cpp
```diff
@@ -440,7 +442,7 @@ VTKH3Slice::execute()
     const float eps = 1e-5; // ensure that the slice is always inside the data set
     if(params().has_path("x_offset"))
     {
-      float offset = params()["x_offset"].to_float32();
+      float offset = get_float32(params()["x_offset"], data_object);
```
cyrush, line 445: its nice how easy this is to wire up

cyrush (approved):

src/ascent/runtimes/flow_filters/ascent_runtime_param_check.cpp
mclarsen: That is why I suggested creating a way to invoke the expression eval without a data set. That said, if we don't give it a data set, the there are a class of functions that will not be available. I suppose we can always give it access to the source.

mclarsen: One suboptimal thing is that if we evaluated the expression inside a pipeline, converting from vtkm -> conduit is probably making a copy, even if the expression was `1+1`.

mclarsen: I take that back. Looks like we do honor the zero copy in this case.

PR #678, opened 2021-01-14
2020 12 update vtkm

mclarsen:


cyrush:
FYI: need to take another pass at the containers, doesn't seem they picked up the new vtk-m.

mclarsen:
https://github.com/Alpine-DAV/spack/commit/47a5dfc2f7df51debe0af4fc7391f3a9a629cd0a

Note for vtkh, it has c copy of the vtkm spack package, so if you aren't using the one I updated in the pull request, that could override it.

cyrush:
@mclarsen -- passing

PR #688, opened 2021-02-18
Compile without vtkm

mclarsen:


mclarsen:
@cyrush I can't think of anything I missed. Maybe something in the spack package. Should we add a build that is `+dray~vtkm`?

Issue #690, opened 2021-02-18
linking issue with pmt


cyrush:
updating ascent to use 1.0.0 tags of both babelflow and pmt, `t_ascent_mpi_smoke` fails b/c it can't find pmt. 

Logic for linking babelflow + pmt in ascent hasn't changed.

My hunch is, this has something to do with cmake rpath options that are passed to babelflow and/or pmt.

They are different from our other deps, in that they report `@rpath` rel paths or have fully stripped paths as their so names. 

When linking in ascent's build system -- the full paths are passed. 

Here is an example of `t_ascent_mpi_smoke`'s deps:

```
[harrison37@elder build-debug (task/2021_02_babel_flow_use_tags)]$ ldd tests/ascent/t_ascent_mpi_smoke 
tests/ascent/t_ascent_mpi_smoke:
	@rpath/libascent_mpi.dylib (compatibility version 0.0.0, current version 0.0.0)
	/Users/harrison37/Work/alpine/ascent/build-debug/lib/libgtest_maind.dylib (compatibility version 0.0.0, current version 0.0.0)
	/Users/harrison37/Work/alpine/ascent/build-debug/lib/libgtestd.dylib (compatibility version 0.0.0, current version 0.0.0)
	/Users/harrison37/Work/alpine/ascent/ul_t/spack/opt/spack/darwin-mojave-x86_64/clang-10.0.0-apple/mpich-3.3.1-kkqsshrtda4dgcnbbwxndombmgfsmp4f/lib/libmpi.12.dylib (compatibility version 14.0.0, current version 14.7.0)
	/Users/harrison37/Work/alpine/ascent/ul_t/spack/opt/spack/darwin-mojave-x86_64/clang-10.0.0-apple/mpich-3.3.1-kkqsshrtda4dgcnbbwxndombmgfsmp4f/lib/libpmpi.12.dylib (compatibility version 14.0.0, current version 14.7.0)
	/Users/harrison37/Work/alpine/ascent/ul_t/spack/opt/spack/darwin-mojave-x86_64/clang-10.0.0-apple/mpich-3.3.1-kkqsshrtda4dgcnbbwxndombmgfsmp4f/lib/libmpicxx.12.dylib (compatibility version 14.0.0, current version 14.7.0)
[...]

PR #691, opened 2021-02-18
use tagged babelflow and pmt

cyrush:
blocked by:

https://github.com/Alpine-DAV/ascent/issues/690


cyrush:
@mclarsen some whack-a-mole going on on here with with ascent_render_3d:

Screenshot 2021-03-11 172415
Screenshot 2021-03-11 172337


I just rebased lined these in docker, and got a prior pass on last pr. Any ideas what could be causing the differences?

PR #697, opened 2021-02-25
cinema db: use compiled resources instead of file system ones

### LINKED ISSUE #692: provide option to set path to file resources for cases when an install is moved
to help resolve:

https://github.com/spack/spack/issues/21739

cyrush:
cinema db: avoid resources on file system, generate from compiled-in base64 encoded data

resolves #692


cyrush:
I did timing test to see if new case was worse than old filesystem logic.

Sifting though lots of noise - overall its a little slower, but it should be negligible.

PR #701, opened 2021-03-10
use new cuda container setup for ci

cyrush:


cyrush:
@mclarsen  back in business. I'll start working on adding babelflow back in via #691

PR #710, opened 2021-03-26
more spack env updates

cyrush:
I hit this issue while testing these specs:

https://github.com/spack/spack/issues/22585


cyrush:
worked around the spack issue with help from spack team

PR #714, opened 2021-04-01
fix binning issue and add tutorial example

mclarsen:


src/examples/tutorial/ascent_intro/cpp/Makefile
```diff
@@ -50,128 +50,135 @@
...
 
-# See $(ASCENT_DIR)/share/ascent/ascent_config.mk for detailed linking info 
+# See $(ASCENT_DIR)/share/ascent/ascent_config.mk for detailed linking info
 include $(ASCENT_DIR)/share/ascent/ascent_config.mk
 
 # make 'all' the default target
 .PHONY: default
 default: all
 
+CXX_FLAGS = -std=c++11
```
cyrush, line 70: yep, that's where we needed -- good add!

src/examples/tutorial/ascent_intro/cpp/plot_binning_2d.py
```diff
@@ -0,0 +1,49 @@
+import yaml #pip install --user pyyaml
+import pandas as pd
```
cyrush: we can drop pandas here, and I will install pyyaml as part of the extra container setup.

cyrush (approved):
Looks good thanks!

PR #729, opened 2021-04-30
[WIP] fix ascent relay runtime for conduit-57702db

xjrc:
The changes in this pull request attempt to fix an issue with the Ascent/Relay runtime introduced by [conduit@57702db](https://github.com/LLNL/conduit/commit/57702db0ce35b5cbeaee31179659a88385d94bce), which changed the logic for `blueprint::mesh::verify` in a way that necessitates different handling for empty nodes. For a more complete overview of this issue/fix, see [conduit#752](https://github.com/LLNL/conduit/issues/752) for details on the Ascent issue and [conduit#731](https://github.com/LLNL/conduit/pull/731#issuecomment-814366371) for details on how Conduit has changed.

An important note about these changes: the new implementation isn't as lenient as the prior version, which may or may not be a problem. More specifically, the prior version performed domain-by-domain verification for multi-domain meshes and allowed through all valid domains. By contrast, the new version performs a bulk mesh verification and only passes through the constituent domains if the verification succeeds (i.e. all domains **must be valid**). If the Ascent Relay runtime needs to support cases with one or more invalid domains (i.e. `blueprint::mesh::verify(domain)` is `false`), then this fix needs to be reworked.

### DIFF
--- a/src/ascent/runtimes/flow_filters/ascent_runtime_relay_filters.cpp
+++ b/src/ascent/runtimes/flow_filters/ascent_runtime_relay_filters.cpp
@@ -254,40 +254,16 @@ void make_domain_ids(conduit::Node &domains)
 bool clean_mesh(const conduit::Node &data, conduit::Node &output)
 {
   output.reset();
-  const int potential_doms = data.number_of_children();
-  bool maybe_multi_dom = true;
 
-  if(!data.dtype().is_object() && !data.dtype().is_list())
+  conduit::Node info;
+  if(blueprint::mesh::verify(data, info))
   {
-    maybe_multi_dom = false;
-  }
-
-  if(maybe_multi_dom)
-  {
-    // check all the children for valid domains
-    for(int i = 0; i < potential_doms; ++i)
+    const auto domains = blueprint::mesh::domains(data);
+    for(auto it = domains.cbegin(); it != domains.cend(); ++it)
     {
-      conduit::Node info;
-      const conduit::Node &child = data.child(i);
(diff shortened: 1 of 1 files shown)


cyrush:
The new logic is much nicer - simplifies what we need to do in ascent. 

The spirit of clean mesh was to take us from a place where we possibly have empty data (which used to be invalid blueprint), back to a place were we have valid blueprint mesh and have removed all empty data. 

Since the domains functions skips empty data, using the domains function cleans all of that up. 

One complication, since we have folks building ascent develop against a few versions of conduit, we will need to find a path to support both old and new. 

If we want to test against containers with newer conduit, you can make the azure pipelines changes that use newer containers, which are also in this PR:

https://github.com/Alpine-DAV/ascent/pull/726/files

cyrush:
I verified that this fixes the hang.  I'll find a path that can support he case where old conduit and new both exist.

cyrush:
Here is another attempt: #731

xjrc:
I'm planning to work on this a bit tomorrow, so I will try out the new container then. Thanks for the heads up and for implementing a more version-agnostic solution.

cyrush:
I did test with the new containers, so I can verify it works.

xjrc:
Ah okay, you're way ahead of me! Thanks, Cyrus!

cyrush:
@xjrc closing to use #731 in lieu of this. When we have compile time + runtime checks for bp features we can adopt the new simpler API in addition to the old path, and then eventually deprecate the old  path.

PR #731, opened 2021-05-03
guard needed for new conduit mesh bp

cyrush:


cyrush:
@xjrc I think this simple guard works for both cases (old and new mesh bp). Does this make sense to you given the changes?

xjrc:
@cyrush: Yes, this also looks like a valid solution to me. The logic is becoming a bit hard to follow so I hope we can move to the `blueprint::mesh::domains` solution eventually, but this should meet our needs for the time being.

PR #737, opened 2021-05-11
update ci containers, add ubuntu 20.10 container

cyrush:


cyrush:
@mclarsen  I think we need the changes in:

https://github.com/Alpine-DAV/ascent/pull/732

for the latest release to work.

Should we try to merge these changes into that branch?

PR #746, opened 2021-05-21
Add the ability to add comments to png files.

brugger1:
I added the ability to add comments to png images. The comments are visible in XV.

This change involved changes to vtk-h and ascent. These are the ascent changes.

I modified the publish method to look for specific fields in the conduit tree and add them to the Metadata node so they would be available in the execute method.

I modified the main runtime to pass the comments to the vtk-h Render class.

I modified lulesh-init to add a couple of additional comments that ascent now looks for.

I modified lulesh to add the image_prefix so that images would get generated. I also fixed some whitespace/style issues in the same general area so that it was consistent with the rest of lulesh.

Here is an image from the resulting file showing the comments in XV from lulesh.

xv_comments


brugger1:
I believe the CI container needs to have vtk-h updated for the tests to pass.

cyrush:
@brugger1 I agree.
I think @mclarsen wants to do a vtk-h release? If that is coming soon, I can update the containers then.

brugger1:
@***@***.******@***.***> sounds good.

cyrush:
we have new vtk-h in our CI, so tests now pass.


Thanks @brugger1

PR #752, opened 2021-06-04
Initial implementation of htg save flow filter.

brugger1:
This is an initial implementation of the htg save flow filter.

I added a filter as well as a test that saves an asci vtk htg file. I will probably change this to a binary vtk htg file at some point, but we can save that for another commit.

This is only a draft pull request and I would like some feedback in addition to reviewing the code.

1) What is the appropriate level of error checking in the filter?
2) I named the filter htg under extract. Is that appropriate?
3) I created a "blueprint" object to store my data. I think it is actually pretty good except that I have field values in the topology. I should probably put those in fields. The values would be element, where each element would be a tree, and the number of elements per tree is arbitrary and varies among the trees.
4) My test creates an htg file. Is there a way to save that and then check that the one generated matches the expected result?

### DIFF
--- a/src/ascent/CMakeLists.txt
+++ b/src/ascent/CMakeLists.txt
@@ -105,6 +105,7 @@ set(ascent_sources
     runtimes/flow_filters/ascent_runtime_param_check.cpp
     runtimes/flow_filters/ascent_runtime_relay_filters.cpp
     runtimes/flow_filters/ascent_runtime_blueprint_filters.cpp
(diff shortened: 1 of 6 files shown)


cyrush:
@brugger1 does the work in #915 supercede this?

brugger1:
Yes, #915 supersedes this. I totally forgot about this. It was a different approach to what is committed. This pull request saves out an htg that has already been created, whereas #915 creates an htg from a regular mesh and saves it out. I'm closing this.

PR #777, opened 2021-07-27
add path for saving sessions by name and with selections

mclarsen:


cyrush (approved):
looks good, should we add a test that checks the existence of a named session file? (or maybe a ticket to remind is in the future)

PR #786, opened 2021-07-29
Various fixes to paraview_ascent_source.py

gpotter2:
I'm not sure how used this file is, but I had quite a bunch of issues with it while trying to pipe WarpX->Ascent->Paraview.

- support `point` shape
- extent wrongly computed in some cases (`size[i]` does not exist)
- the information was constantly overwritten when using different output ports: using `OutputPort(ascentSource, 0)`, `OutputPort(ascentSource, 1)`... would return objects with missing extents.

### DIFF
--- a/src/examples/paraview-vis/paraview_ascent_source.py
+++ b/src/examples/paraview-vis/paraview_ascent_source.py
@@ -169,10 +169,13 @@ def ascent_to_vtk(node, topology=None, extent=None):
             elif (shape == "quad"):
                 npoints = 4
                 cellType = vtkConstants.VTK_QUAD
+            elif (shape == "point"):
+                npoints = 1
+                cellType = vtkConstants.VTK_PIXEL
             else:
-                print("Error: Shape not implemented")
+                print("Error: Shape '%s' not implemented" % shape)
                 return None
-            c = c.reshape(c.shape[0] / npoints, npoints)
+            c = c.reshape(c.shape[0] // npoints, npoints)
             # insert the number of points before point references
             c = np.insert(c, 0, npoints, axis=1)
             ncells = c.shape[0]
@@ -239,7 +242,7 @@ def write_vtk(prefix, node, data):
         writer = vtkXMLUnstructuredGridWriter()
         extension = "vtu"
     else:
-        print("Error: Unknown datatype.")
+        print("Error: Unknown datatype: '%s'" % data.GetClassName())
         return
     fileNoExt = get_filenoext(prefix, node)
     writer.SetFileName(fileNoExt + "." + extension)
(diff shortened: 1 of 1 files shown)


cyrush:
thanks @gpotter2! 

@danlipsa these look like good suggestions to me, can you take a look at them (as paraview expert :-) )

danlipsa:
@gpotter2 Thanks for your contributions and I apologize for my delay in responding. I had a deadline yesterday. Indeed, this was only tested on the demo simulations. These look good. I am rebuilding spack right now to test your branch on those (that stopped working as building doxygen hangs). I'll get back to you once I tested your changes on the demo simulations.
I've seen you you got rid of some of 'self.' prefixes. Is there a PEP rule that specifies one way or the other? Fields already have _ but I followed VTK rule to prefix all fields anyway.

src/examples/paraview-vis/paraview_ascent_source.py
```diff
@@ -435,39 +435,41 @@ def RequestData(self, request, inVector, outVector):
     def RequestInformation(self, request, inVector, outVector):
         for outputPort in range(self.GetNumberOfOutputPorts()):
             oi = outVector.GetInformationObject(outputPort)
-            if (self._outputType != vtkConstants.VTK_UNSTRUCTURED_GRID):
+            _topology = self._topologies[outputPort]
+            _coords = self._coords[outputPort]
```
gpotter2: I got rid of the `self.` because I redefined those. Maybe I should call them something else for it to be clearer?

gpotter2:
By the way, does this:
https://github.com/Alpine-DAV/ascent/blob/def0079116861a4afa7c4a3e64eca4f99e2028a4/src/examples/paraview-vis/paraview_ascent_source.py#L338
mean that it does not support multi-domain data? Is there a way of making it work?

danlipsa:
> mean that it does not support multi-domain data? Is there a way of making it work?

Can you point me to an example of multi-domain data as blueprint json? I don't know exactly what it is. Yes, I only implemented what I could test = the example integrations available in ascent.

src/examples/paraview-vis/paraview_ascent_source.py
danlipsa: Do you refer to _topology? That does seem to be a field so it should be prefixed with self. for consistency isn't it?

cyrush:
@danlipsa 

for the multi domain case (which is how ascent presents all meshes) each child of the root node is a mesh domain. 

`self._node = ascent_extract.ascent_data().child(0) `

This grabs the first domain on each MPI task, which is fine if the sim code is not domain overloaded. 

But in cases where there are multiple domains per mpi task, you need to process all the children.

Does that help?

danlipsa:
So, the mesh domains are just different meshes (related to the same data)?
Seems the different mesh domains can have different time values.

Probably, the different domains can be stored on different VTK output
ports. Not sure what to do with the possibly different time values.

cyrush:
don't worry about different time values, when presented it is implied that they are spatially domain decomposed mesh.

src/examples/paraview-vis/paraview_ascent_source.py
gpotter2: It's just a utility variable to avoid repeating `self._coords[outputPort]` we don't want to store it in the object, so no `self.` ?

danlipsa:
In this case it should not have an underscore
https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

cyrush:
@danlipsa  @gpotter2  are these set of changes ready to go in?

danlipsa:
Looks good to me. +1

cyrush:
thanks @gpotter2  + @danlipsa !

PR #793, opened 2021-08-04
add time function

mclarsen:


cyrush (commented):
should we  add a test for time before we forget?

PR #794, opened 2021-08-05
paraview_ascent_source.py - multi-domain support for explicit topology

gpotter2:
**requires https://github.com/Alpine-DAV/ascent/pull/786**

This PR:
- add support for multi-domains with explicit topologies. This works fine using WarpX particles

Notes:
- I haven't implemented multi-domain support for uniform/rectilinear topologies because it looks much harder and I don't use it
- The `structured` is untested (but should probably work), since WarpX particles use `unstructured`

This might not be useful or a great implementation. I doubt this is very efficient, so I'm not sure if I'll be able to expand on this PR, this is more of a PoC regarding the discussion in https://github.com/Alpine-DAV/ascent/pull/786

### DIFF
--- a/src/examples/paraview-vis/paraview_ascent_source.py
+++ b/src/examples/paraview-vis/paraview_ascent_source.py
@@ -25,37 +25,78 @@ import conduit.relay.io
 _keep_around = []
 
 
-def read_fields(node, topology, data):
+def is_multi_domain(node):
+    '''
+    Returns if the node is multi-domain.
+    '''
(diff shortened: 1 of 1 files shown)


cyrush:
@danlipsa @gpotter2 I would like to get these changes in, what do you think the best path forward is from here?

danlipsa:
I'll take a look at it.

gpotter2:
I rebased the PR for it to be easier to review.

PR #795, opened 2021-08-06
fix initialization of first_index

mclarsen:


src/ascent/runtimes/expressions/ascent_expression_filters.cpp
```diff
@@ -1382,35 +1382,43 @@ ScalarGradient::execute()
...
+  {
     (*output)["value"] = -std::numeric_limits<double>::infinity();
     (*output)["type"] = "double";
     set_output<conduit::Node>(output);
     return;
   }
 
-  int first_index = -1, current_index = entries - 1;
-  if(execution_points) {
+  int first_index = 0, current_index = entries - 1;
```
mclarsen, line 1408: This is what I changed (other than some formatting).

mlawsonca: Good catch. first_index needed to be set to -1 for a subsequent expression, but not for this one.

mlawsonca (approved):

cyrush:
@mclarsen should I merge in develop (with new ci setup) and then get this merged in?

mclarsen:
We can if we get rid of the mesh quality stuff which is not merged in vtkh.

mclarsen:
The problem is that the mesh quality only works on completely unstructured grids, and I think in the branch, I have to turn it to unstructured. We can decide what to do about that. They fixed it in vtkm develop, but that could be a pain.

cyrush:
got it, lets discuss in the future

mclarsen:
closing as changes were merged in #810

PR #835, opened 2021-11-16
Fixing bug where ascent_filter_timings.csv is populated twice

yuya737:
`delete m_runtime;` already calls `m_runtime->Cleanup();` via its destructor.

In the current setup, `m_runtime->Cleanup()` is called twice, populating the `ascent_filter_timings.csv` twice if `timings == "true"` is set.


cyrush (approved):
Good catch, thanks for the fix!

Issue #827, opened 2021-10-16
check html diag files


cyrush:
they are always on (flow graph and expressions graph)

cyrush:
resolved in #845

PR #845, opened 2021-12-23
update to use vtk-m 1.7.0 and vtk-h 0.8.0

cyrush:


cyrush:
resolves #846

cyrush:
resolves #827

Issue #770, opened 2021-07-21
ascent.mk entries for fides and adios2


cyrush:
reminder we need to add these if non-cmake apps want to leverage these features

cyrush:
resolved in #847

Issue #557, opened 2020-07-10
add using-with-cmake-mpi example


cyrush:

cyrush:
resolved in #847

PR #847, opened 2021-12-28
add cmake-with-mpi example

cyrush:


cyrush:
resolves #557

cyrush:
resolves #770

Issue #834, opened 2021-11-15
update hashes.txt


cyrush:
they are currently out of date. 

(ascent develop requires vtk-h develop)

cyrush:
resolved in #853

Issue #416, opened 2019-11-15
update copyright in license


cyrush:
Copyright (c) 2015-2018,

(showing in runtime lic msg), we should also move to simpler spidx scheme for source files.

cyrush:
resolved in #853

PR #853, opened 2021-12-30
update source file headers

cyrush:
to remove dates which get stale


cyrush:
resolves #416

cyrush:
resolves #834

PR #854, opened 2021-12-30
Add phi_range and theta_range to the cinema rendering options.

### LINKED ISSUE #828: Add new cinema options for camera combos
1. User provides two lists of values (phi, theta) and the cross product of those lists are the camera positions output. For example:

    phi: [0,90]
    theta: [0,45]

    Would create camera views from [phi, theta] pairs of [0, 0] [0, 45] [90, 0] [90, 45]

    2. User provides a list of pairs, and these are the [phi, theta] pairs of camera positions output. For example:

    camera_positions: [[0,0], [0,10], [0,20], [0,30]]

    This would create four camera positions.

    3. User provides two sets of range, increment data, and the cross product of the camera positions are created. For example:

    phi: [0-90], 10
    theta: [0-90], 10

    would be shorthand for:

    phi: [0,10,20,30,40,50,60,70,80,90]
    theta: [0,10,20,30,40,50,60,70,80,90]

brugger1:
This resolves https://github.com/Alpine-DAV/ascent/issues/828

I added `phi_range` and `theta_range` to the cinema rendering options. Here are examples of using the new options.

```
    scenes["scene1/renders/r1/type"] = "cinema";
    scenes["scene1/renders/r1/phi_range"].set({-180., 180.});
    scenes["scene1/renders/r1/dphi"] = 90.;
    scenes["scene1/renders/r1/theta_range"].set({0., 180.});
    scenes["scene1/renders/r1/dtheta"] = 90.;
```

This results in the following angles:
phi: -180, -90, 0, 90, 180
theta: 0, 90, 180

```
    scenes["scene1/renders/r1/type"] = "cinema";
    scenes["scene1/renders/r1/phi_range"].set({-180., 180.});
    scenes["scene1/renders/r1/phi_num_angles"] = 5;
    scenes["scene1/renders/r1/theta_range"].set({0., 180.});
    scenes["scene1/renders/r1/theta_num_angles"] = 3;
```

This also results in the same angles as above:
phi: -180, -90, 0, 90, 180
theta: 0, 90, 180

I added a couple of tests with these new options in `src/tests/ascent/t_ascent_cinema_a.cpp`.

I didn't update the documentation, since I believe the existing documentation isn't completely correct in the description of `phi` and `theta`. I'm not sure if the documentation should be fixed or if the code should be changed. Once the correct action is determined, then I will update the documentation.

### DIFF
--- a/src/ascent/runtimes/flow_filters/ascent_runtime_rendering_filters.cpp
+++ b/src/ascent/runtimes/flow_filters/ascent_runtime_rendering_filters.cpp
@@ -185,10 +185,12 @@ check_renders_surprises(const conduit::Node &renders_node)
   r_valid_paths.push_back("phi_range");
   r_valid_paths.push_back("dphi");
   r_valid_paths.push_back("phi_num_angles");
+  r_valid_paths.push_back("phi_angles");
   r_valid_paths.push_back("theta");
   r_valid_paths.push_back("theta_range");
   r_valid_paths.push_back("dtheta");
   r_valid_paths.push_back("theta_num_angles");
+  r_valid_paths.push_back("theta_angles");
   r_valid_paths.push_back("db_name");
   r_valid_paths.push_back("render_bg");
   r_valid_paths.push_back("annotations");
@@ -499,6 +501,7 @@ class CinemaManager
 protected:
   std::vector<vtkm::rendering::Camera> m_cameras;
   std::vector<std::string>             m_image_names;
+  std::vector<std::tuple<float,float>> m_camera_angles;
   std::vector<float>                   m_phi_values;
   std::vector<float>                   m_theta_values;
   std::vector<float>                   m_times;
@@ -531,6 +534,15 @@ public:
       m_phi_min = -180.0;
       m_phi = render_node["phi"].to_int32();
       m_phi_inc = 360.0 / double(m_phi);
+      this->create_angles_from_range(m_phi_values, m_phi_min, m_phi_inc, m_phi);
+    }
+    else if(render_node.has_path("phi_angles"))
+    {
+      float64_accessor phi_angles = render_node["phi_angles"].as_float64_accessor();
+      for(int p = 0; p < phi_angles.number_of_elements(); ++p)
+      {
+        m_phi_values.push_back(phi_angles[p]);
+      }
     }
     else if(render_node.has_path("phi_range") && render_node.has_path("dphi"))
     {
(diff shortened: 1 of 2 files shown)


src/ascent/runtimes/flow_filters/ascent_runtime_rendering_filters.cpp
```diff
@@ -499,34 +505,101 @@ class CinemaManager
...
-      m_phi(phi),
-      m_theta(theta),
       m_image_name(image_name),
       m_time(0.f)
   {
+    // Handle phi.
+    if(render_node.has_path("phi"))
+    {
+      m_phi_min = -180.0;
+      m_phi = render_node["phi"].as_int32();
```
cyrush: small suggested change - in the places were you have `as_int32` you can use `to_int32` to avoid an error if folks passed an int64 (or other int type)

src/ascent/runtimes/flow_filters/ascent_runtime_rendering_filters.cpp
```diff
@@ -499,34 +505,101 @@ class CinemaManager
...
+      m_phi_inc = 360.0 / double(m_phi);
+    }
+    else if(render_node.has_path("phi_range") && render_node.has_path("dphi"))
+    {
+      const double *phi_range = render_node["phi_range"].as_float64_ptr();
```
cyrush: FYI: with new conduit release, you can use:
```
float64_accessor phi_range = render_node["phi_range"].as_float64_accessor();
```
then it wont matter what type the user provided.

cyrush:
Looks good! I added few suggestions, just to make the parsing more robust. 
Also: Expect the babel flow test to fail (I am working on disabling in another pr until we have a fix)

cyrush:
for the docs: Lets talk about it. If the version in the docs matches other expectations (like mathematica , etc),  we might simply have errors with the implementation that were not intended. I think we should try to make it match common expectations from other tools or sources.

src/ascent/runtimes/flow_filters/ascent_runtime_rendering_filters.cpp
brugger1: Changed.

src/ascent/runtimes/flow_filters/ascent_runtime_rendering_filters.cpp
brugger1: Changed.

cyrush (approved):
Looks good!

PR #874, opened 2022-02-08
include fix for ascent_mpi imported target

cyrush:


cyrush:
i think this is already there.

Issue #857, opened 2022-01-07
Update ascent spack package to add "+testing" to vtk-m spec for vtk-m 1.7.0.


brugger1:
The new vtk-m 1.7.0 spack install no longer includes the testing header files. In particular, `vtkm/cont/testing/MakeTestDataSet.h`, which is used by some of the ascent tests.

I have a pull request for spack https://github.com/spack/spack/pull/28324, that adds a `testing` variant, that will add the testing files back into an install. When that change is accepted, the ascent spack package should be updated to add "+testing" to the vtk-m spec.

cyrush:
We refactored so Ascent develop no longer depends on `vtk-m+testing`.

PR #903, opened 2022-04-12
Allow shared cuda

kwryankrattiger:
@cyrush This one needs an update for this too.

### DIFF
--- a/src/rover/CMakeLists.txt
+++ b/src/rover/CMakeLists.txt
@@ -4,6 +4,10 @@
 # other details. No copyright assignment is required to contribute to Ascent.
 ###############################################################################
 
+# Always build rover as a static PIC lib
+set(BUILD_SHARED_LIBS OFF)
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+
 set(rover_thirdparty_deps vtkh_lodepng vtkm vtkh conduit conduit_relay)
 
 set(rover_headers
--- a/src/rover/CMakeLists.txt
+++ b/src/rover/CMakeLists.txt
@@ -4,10 +4,6 @@
 # other details. No copyright assignment is required to contribute to Ascent.
 ###############################################################################
 
-# Always build rover as a static PIC lib
-set(BUILD_SHARED_LIBS OFF)
-set(CMAKE_POSITION_INDEPENDENT_CODE ON)
-
 set(rover_thirdparty_deps vtkh_lodepng vtkm vtkh conduit conduit_relay)
 
 set(rover_headers
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -387,6 +387,13 @@ stages:
           # static linking unit tests eats up alot of disk space,
           # so skip building tests in CI to keep us away from clif
           ENABLE_TESTS: OFF
+        cuda-11.4.0-shared:
+          containerImage: ${{ variables.ubuntu_18_cuda_11_4_0_tag }}
+          BUILD_SHARED_LIBS: ON
+          CMAKE_BUILD_TYPE: Release
+          ENABLE_BABELFLOW: OFF
+          # shared linking unit tests shouldn't eat disk space,
+          ENABLE_TESTS: ON
     container: $[ variables['containerImage'] ]
     steps:
       - checkout: self
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -50,10 +50,6 @@ else()
   cmake_minimum_required(VERSION 3.9)
 endif()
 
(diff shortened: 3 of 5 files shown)


kwryankrattiger:
When building this version of ascent with spack I am getting issues with not finding `vtkm_cont_testing` and `vtkmdiympi_nompi` targets. There are two solutions to this, either fix the spec to force these targets to exist, or generate warnings instead of errors in the `source_objects` function when a target in the list is not found.

Another option could be to append these libraries if certain conditions are met, but I am not sure what those conditions should be.

This is the error:
```
get_target_property() called with non-existent target "vtkmdiympi_nompi"
```

cyrush:
Q: In your spack builds, are you building vtk-m with mpi support?

Sounds like we need some guards or logic when mpi is on related to DIY deps.

kwryankrattiger:
@cyrush VTK-m is built with MPI support, but no testing tools (#857).

`vtk-m@1.7.1+ascent_types+cuda+cuda_native+doubleprecision~logging+mpi+openmp+rendering+shared~tbb~testlib~virtuals build_type=Release cuda_arch=86`

kwryankrattiger:
@cyrush Okay, it looks like I fixed the configure issues, but it looks like the shared cuda CI needs a different image, mfem, and probably others, need to be built shared. Not sure what your workflow is for that.

cyrush:
I see, looks like some static tpls don't have `fpic`, so they can't be linked into shared ascent. 

For this case I will need to create a new container that builds tpls with shared libs + cuda. 

Might be good for me to test using newer spack, vs back porting changes to support `cuda+shared` to our stable fork.

kwryankrattiger:
Awesome. I am working on a spack PR that will allow concertizing this case, currently there are a few issues to allowing `+shared+cuda` for dependent packages.

On that note, I am having other issues getting VTK-m@1.7.x to work with ascent 0.8.0. Not sure why, but it seems some of the `vtkm::cont` stuff is not defined for cuda. Still working on figuring out why.

src/rover/CMakeLists.txt
```diff
@@ -4,6 +4,10 @@
 # other details. No copyright assignment is required to contribute to Ascent.
 ###############################################################################
 
+# Always build rover as a static PIC lib
+set(BUILD_SHARED_LIBS OFF)
```
cyrush: thought: Rover uses VTK-m, so if we do need to change to static libs here, I would like to better understand why. This could be a signal something is missing. 

We will need to run tests with real cuda hw --  b/c even if we survive linking, we could get immediate runtime sadness.

kwryankrattiger: ah, I missed that dep. The only reason it is forced static is because it is hard-coded as librover.a in a couple [places](https://github.com/Alpine-DAV/ascent/blob/develop/src/ascent/CMakeLists.txt#L411).

Although, does it matter if rover is compiled with PIC? I would think since it is internal and has PUBLIC deps from cmake the libraries would be linked correctly in the end.

cyrush: gotcha -- I think we can adjust to look for it in the right place and use top level choice of static/shared.  I think we just need to ask cmake about the rover target.

kwryankrattiger: Okay, updated to use a generator expression to resolve the target lib location.

cyrush: Thanks!

kwryankrattiger:
@cyrush I finally got this working with building shared in spack. Thanks for your help on this!

kwryankrattiger:
Ran the ctest for ascent using this branch and spack dev-build -b install and it seems everything is passing with cuda and shared for me. I think at this point it just needs an new CI image and rebase some of these changes and it should be good to go!

```
96% tests passed, 3 tests failed out of 70

Total Test time (real) = 1500.07 sec

The following tests FAILED:
          4 - blt_mpi_smoke (Failed)
         18 - t_ascent_cinema_a (Timeout)
         20 - t_ascent_render_3d (Timeout)
```

cyrush:
@kwryankrattiger awesome & thanks so much for your effort here!

Can you give me your exact dev-build command?

I'll work on the new container for CI later this week.

kwryankrattiger:
@cyrush 
> Can you give me your exact dev-build command?

```
spack dev-build -j 40 --skip-patch --no-checksum --drop-in bash -b install ascent@0.8.0-dev+cuda cuda_arch=75
```

cyrush:
@kwryankrattiger  FYI: Merging this now, I plan to disable the shared CI test in a follow up to keep things green until I can get the spack stuff ready to build the new container with shared tpls.

PR #909, opened 2022-04-25
Add blueprint partition and flatten 

nicolemarsaglia:
Add the partition and flatten functionality provided by conduit blueprint to ascent as a pipeline and extract, respectively. Includes unit tests.

### DIFF
--- a/src/ascent/runtimes/flow_filters/ascent_runtime_blueprint_filters.cpp
+++ b/src/ascent/runtimes/flow_filters/ascent_runtime_blueprint_filters.cpp
@@ -21,6 +21,7 @@
 #include <conduit.hpp>
 #include <conduit_relay.hpp>
 #include <conduit_blueprint.hpp>
+#include <conduit_blueprint_mesh.hpp>
 
 //-----------------------------------------------------------------------------
 // ascent includes
--- a/src/ascent/runtimes/flow_filters/ascent_runtime_blueprint_filters.cpp
+++ b/src/ascent/runtimes/flow_filters/ascent_runtime_blueprint_filters.cpp
@@ -207,11 +207,11 @@ BlueprintPartition::verify_params(const conduit::Node &params,
     info.reset();
     bool res = true;
 
-//    if(params.has_child("selections") )
-//    {
-//        const int num_selections = params["selections"].number_of_children();
-//	  if(num_selections 
-//    }
+    if(! params.has_child("target") ||
+       ! params["target"].dtype().is_int() )
+    {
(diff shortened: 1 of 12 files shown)


src/ascent/runtimes/flow_filters/ascent_runtime_blueprint_filters.cpp
```diff
@@ -174,6 +177,81 @@ BlueprintVerify::execute()
...
+{
+    if(!input(0).check_type<DataObject>())
+    {
+        ASCENT_ERROR("blueprint_data_partition input must be a DataObject");
+    }
+
+    DataObject *d_input = input<DataObject>(0);
+    std::shared_ptr<conduit::Node> n_input = d_input->as_node();
+
+    conduit::Node n_output;
```
cyrush: The data object will own the Node and take care of managing when it needs to be deleted.  So, we can change this to conduit::Node *n_output = new conduit::Node();

src/ascent/runtimes/flow_filters/ascent_runtime_relay_filters.cpp
```diff
@@ -1240,6 +1243,161 @@ RelayIOLoad::execute()
...
+    int rank = 0;
+    int root = 0;
+ 
+#ifdef ASCENT_MPI_ENABLED
+    MPI_Comm_rank(mpi_comm, &rank);
+#endif
+
+    if(rank == root)
+    {
+        if(protocol.empty())
```
cyrush, line 1363: we can simplify this here, here is a suggestion:

if empty - use csv, otherwise just pass the protocol to `relay::io::save()`

for the non csv case, we want to support `hdf5`, `json`, `yaml` etc -- passing the protocol simplifies plumbing.

nicolemarsaglia: Sounds good. I'll fix my cases.

cyrush:
https://dev.azure.com/alpine-dav/Ascent/_build/results?buildId=1656&view=logs&j=20a2107f-01e8-5a4b-5092-cc79c8a52d69&t=e251c6c5-edbb-51d8-71a7-a95d41a17939&l=239

Looks like the missing conduit header issue, we should be able to remove the explicit include.

cyrush:
looking great! 

One last suggestion:

We should test the partition cases with the normal relay extract -- not flatten.

B/c with the normal relay extract we will get mesh blueprint data, instead of the csv-like tables. 

Then we can check number of domains (the current check "works" but its getting lucky :-) )

cyrush (approved):
Thanks!

PR #915, opened 2022-04-30
Add the htg extract filter.

brugger1:
Added the htg extract filter. The filter outputs a HyperTreeGrid (HTG) as an ASCII VTK htg file. The file can viewed with ParaView 5.9.0-RC3 or later.

I also added a test for it which creates a basic uniform grid and saves it using the htg extract filter.


nicolemarsaglia:
Overall, looks good. Is there a use case for htg + mpi? If so, does it need logic for parallel coordination?

brugger1:
There's the potential to do this in parallel if the meshes on all the ranks fit together properly. That would be a bunch of additional development, so I'll wait with that for another day.

nicolemarsaglia:
Sounds good. In that case, I'm going to add an #if ASCENT_MPI_ENABLED that will throw an ASCENT_ERROR in execute(). And then I think we are good to go.

nicolemarsaglia (approved):
Looks good.

brugger1:
Thanks for the improvement and merging!!

PR #922, opened 2022-05-09
Update search paths for ADIOS(2).

### LINKED ISSUE #916: SetupADIOS2 issues
Currently Ascent is using the variable `<PACKAGE>_DIR` to explicitly find the config files for third-party libraries. The problem with this is `find_package` sets the `<Package>_DIR` after a find package. This currently doesn't cause issues because many packages, like `Conduit`, have the ascent version of the var `CONDUIT_DIR` and the `find_package` version of the var is `Conduit_DIR`.

However, in the case of `ADIOS2` both ascent and find_package use the `ADIOS2_DIR`. What this means is packages like sensei depend on both ascent and adios2 there is a collision.

A solution for this is to use the `<Package>_ROOT` variable instead since the sole purpose of this variable is to be the prefix path to the package's installation. Another option is to check if `lib/cmake/adios2` is already appended to the `ADIOS2_DIR` and conditionally append it in the `find_package` call.

Note: The solution work around in sensei was to re-order configuring the packages. This is not stable because it will always fail to re-configure unless the `ADIOS2_DIR` is manually removed from the `CMakeCache.txt` or the variable is specified in each configure. I would imagine this would also cause issues in a re-configure of ascent with ADIOS2 support as well.

@cyrush

kwryankrattiger:
Fixes #916

@cyrush

### DIFF
--- a/src/cmake/thirdparty/SetupADIOS.cmake
+++ b/src/cmake/thirdparty/SetupADIOS.cmake
@@ -10,32 +10,32 @@
 #
 ###############################################################################
 
-# first Check for ADIOS_DIR
+# first Check for ADIOS_ROOT
 
-if(NOT ADIOS_DIR)
-    MESSAGE(FATAL_ERROR "ADIOS support needs explicit ADIOS_DIR")
+if(NOT ADIOS_ROOT)
+    MESSAGE(FATAL_ERROR "ADIOS support needs explicit ADIOS_ROOT")
 endif()
 
-MESSAGE(STATUS "Looking for ADIOS using ADIOS_DIR = ${ADIOS_DIR}")
+MESSAGE(STATUS "Looking for ADIOS using ADIOS_ROOT = ${ADIOS_ROOT}")
 
 # CMake's FindADIOS module uses the ADIOS_ROOT env var
-set(ADIOS_ROOT ${ADIOS_DIR})
+set(ADIOS_ROOT ${ADIOS_ROOT})
 set(ENV{ADIOS_ROOT} ${ADIOS_ROOT})
 
 # Use CMake's FindADIOS module, which uses hdf5's compiler wrappers to extract
 # all the info about the hdf5 install
-include(${ADIOS_DIR}/etc/FindADIOS.cmake)
+include(${ADIOS_ROOT}/etc/FindADIOS.cmake)
 
-# FindADIOS sets ADIOS_DIR to it's installed CMake info if it exists
-# we want to keep ADIOS_DIR as the root dir of the install to be
+# FindADIOS sets ADIOS_ROOT to it's installed CMake info if it exists
+# we want to keep ADIOS_ROOT as the root dir of the install to be
 # consistent with other packages
 
-set(ADIOS_DIR ${ADIOS_ROOT} CACHE PATH "" FORCE)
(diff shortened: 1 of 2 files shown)


kwryankrattiger:
I set it up so it _should_ still work with the existing CI images and existing workflows. It also resets the _DIR variable if it was already set for ADIOS2.

I could also change the rest of the packages to use _ROOT, but I have a pretty targeted issue with the ADIOSx packages because of the nature of the naming convention.

src/cmake/thirdparty/SetupADIOS.cmake
```diff
@@ -10,35 +10,35 @@
...
 #set(ADIOS_FOUND TRUE CACHE PATH "" FORCE)
 
 if(NOT ADIOS_FOUND)
-    message(FATAL_ERROR "ADIOS_DIR is not a path to a valid ADIOS install")
+    message(FATAL_ERROR "ADIOS_ROOT is not a path to a valid ADIOS install")
 endif()
 
 blt_register_library(NAME adios
-                     INCLUDES ${ADIOS_INCLUDE_DIRS}
+                     INCLUDES ${ADIOS_INCLUDE_ROOTS}
```
kwryankrattiger: Fix me

src/cmake/thirdparty/SetupADIOS2.cmake
```diff
@@ -10,21 +10,39 @@
...
-MESSAGE(STATUS "Looking for ADIOS2 using ADIOS2_DIR = ${ADIOS2_DIR}")
+MESSAGE(STATUS "Looking for ADIOS2 using ADIOS2_ROOT = ${ADIOS2_ROOT}")
 
-set(ADIOS2_DIR_ORIG ${ADIOS2_DIR})
+set(ADIOS2_DIR_ORIG ${ADIOS2_ROOT})
 
 find_package(ADIOS2 REQUIRED
              NO_DEFAULT_PATH
-             PATHS ${ADIOS2_DIR}/lib/cmake/adios2)
+             PATHS ${ADIOS2_ROOT}/lib/cmake/adios2)
```
kwryankrattiger: So this line is more problematic than I originally thought. I am getting an issue with the installation of ADIOS2 not always being under the `lib/cmake/adios2` directory. Currently I am testing on crusher, and the install is under `lib64/cmake/adios2`.

cyrush: Yes, we had to deal with this style for Kokkos as well:

https://github.com/Alpine-DAV/ascent/blob/develop/src/cmake/thirdparty/SetupKokkos.cmake

kwryankrattiger: I believe just specifying the `PATH` as the `Package_ROOT` should be sufficient. CMake should search under that path following the standard installations locations even if `NO_DEFAULT_PATH` is used. I tested it this way on crusher and was successful. So unless you have multiple installations of adios2 in the same location, which you couldn't anyway, it will find it regardless of the `lib` directory name.

https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure

cyrush: Sounds good, good to see things are evolving to be hopefully simpler over time :-)

cyrush (approved):

PR #936, opened 2022-05-27
Amalgamate VTK-h

cyrush:


cyrush:
using-with-make-example failure:

custom cmake device link step needs new logic, given that vtkh targets are now builtin
(working on it)

cyrush:
green, here we go ....

PR #943, opened 2022-06-02
add feature map tables

cyrush:
@nicolemarsaglia here is a branch to use for the feature map tables docs


cyrush:
@nicolemarsaglia this looks good, what do you think about getting this batch in and then following up with another PR for more changes?

nicolemarsaglia:
@cyrush sounds good to me. Just added to the changelog.

cyrush:
@nicolemarsaglia thanks!

PR #948, opened 2022-06-06
apply indexing bugfix from drangara

### LINKED ISSUE #947: Image filename appears to overwrite index 0
I ran into an issue which seems to overwrite the image name only for the first of multiple files, but works fine thereafter. The actions file has:
```
image_prefix: "./slice_vel_%04d"
```
and I see `slice_vel_0000.png` getting overwritten first time but subsequently the files `slice_vel_0001.png`, `slice_vel_0002.png` are created.

The `counter` that gets sent to `expand_family_name` in https://github.com/Alpine-DAV/ascent/blob/develop/src/ascent/utils/ascent_string_utils.cpp#L42 is always 0. 

If I change the logic inside `expand_family_name` as below, I get the expected behavior and it looks to me as this is the required fix for this issue:
```
--- a/src/ascent/utils/ascent_string_utils.cpp
+++ b/src/ascent/utils/ascent_string_utils.cpp
@@ -89,8 +89,8 @@ std::string expand_family_name(const std::string name, int counter)
     }
     else
     {
-      counter = s_file_family_map[name];
-      s_file_family_map[name] = counter + 1;
+      counter = s_file_family_map[name] + 1;
+      s_file_family_map[name] = counter;
     }
   }
```

Wondering if the code needs the above bug fix. Please let me know if you need additional info or have any questions about the case I ran. Thanks for the help!

cyrush:
resolves #947 

@drangara your fix

Checking vs our CI to see if we had any tests that depend on old (wrong) behavior.

### DIFF
--- a/src/ascent/utils/ascent_string_utils.cpp
+++ b/src/ascent/utils/ascent_string_utils.cpp
@@ -51,8 +51,8 @@ std::string expand_family_name(const std::string name, int counter)
     }
     else
     {
-      counter = s_file_family_map[name];
-      s_file_family_map[name] = counter + 1;
+      counter = s_file_family_map[name] + 1;
+      s_file_family_map[name] = counter;
     }
   }
 


cyrush:
CI looks good, merging.

PR #957, opened 2022-06-22
Fixed uberenv submodule

mvictoras:
When cloning the repo recursively or using Spack, I get this error:

Cloning into '/Users/vmateevitsi/src/alpine/ascent/scripts/uberenv'...
ERROR: Repository not found.
fatal: Could not read from remote repository.

Looks like the submodule url is incorrect, so this PR fixes that.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -6,4 +6,4 @@
 	url = https://github.com/Alpine-DAV/spack_configs.git
 [submodule "scripts/uberenv"]
 	path = scripts/uberenv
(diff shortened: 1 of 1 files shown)


cyrush:
thanks @mvictoras -- strange we haven't encountered this before. I'll test this in the process we used to build our CI docker containers and then merge.

mvictoras:
Very strange indeed. I also tried on theta, and it worked there. Issue is on my Mac.

cyrush:
Merged, thanks @mvictoras !

PR #961, opened 2022-06-29
Add log base 10 filter

nicolemarsaglia:
*Add a log base 10 filter: "log10"
*Update Docs: Add log10 example & differentiate natural log and log10 examples.


cyrush:
Changes and example look great!

Can you add a unit test that tests your example in `src/tests/ascent/t_ascent_log.cpp` as well?

cyrush:
Also, add  changelog entry?

nicolemarsaglia:
@cyrush I updated the changelog and added unit tests, but they don't match my example since that came from lulesh. Do you want them to match?

cyrush (approved):
@nicolemarsaglia  thanks for adding, everything looks good. 

It's fine if the example don't match the unit tests.

PR #919, opened 2022-05-03
vtkm 1.8 update

cyrush:

PR #962, opened 2022-07-14
Find deps config

kwryankrattiger:
The extra path info breaks cases were other projects define these variables and they get re-set to the full path to the config file.

By default CMake will search for the config files under the given path in the locations the deps config is hardcoding. It will also look in `<root>/lib64/cmake` path which may be required depending on the system and what `CMAKE_INSTALL_LIBDIR` is set to. This has been an issue for a couple of projects so I changed a few instances.

### DIFF
--- a/src/config/ascent_setup_deps.cmake
+++ b/src/config/ascent_setup_deps.cmake
@@ -51,7 +51,7 @@ endif()
 
 if(VTKH_DIR)
     if(NOT EXISTS ${VTKH_DIR}/lib/VTKhConfig.cmake)
-        message(FATAL_ERROR "Could not find VTKh CMake include file (${VTKH_DIR}/lib/VTKhConfig.cmake)")
+      message(FATAL_ERROR "Could not find VTKh CMake include file (${VTKH_DIR}/lib/VTKhConfig.cmake)")
     endif()
 
     ###############################################################################
@@ -59,7 +59,7 @@ if(VTKH_DIR)
     ###############################################################################
     find_dependency(VTKh REQUIRED
                     NO_DEFAULT_PATH
-                    PATHS ${VTKH_DIR}/lib/)
+                    PATHS ${VTKH_DIR}/lib)
 endif()
 
 ###############################################################################
@@ -71,15 +71,16 @@ endif()
 
 if(VTKM_DIR)
     # use VTKM_DIR to setup the options that cmake's find VTKm needs
-    file(GLOB VTKm_DIR "${VTKM_DIR}/lib/cmake/vtkm-*")
-    if(NOT VTKm_DIR)
-        message(FATAL_ERROR "Failed to find VTKm at VTKM_DIR=${VTKM_DIR}/lib/cmake/vtk-*")
+    if(NOT EXISTS ${VTKM_DIR})
+        message(FATAL_ERROR "Failed to find VTKm at VTKM_DIR=${VTKM_DIR}")
     endif()
 
     ###############################################################################
     # Import CMake targets
     ###############################################################################
-    find_dependency(VTKm REQUIRED)
+    find_dependency(VTKm REQUIRED
+      NO_DEFAULT_PATH
+      PATHS ${VTKM_DIR})
 endif()
 
 ###############################################################################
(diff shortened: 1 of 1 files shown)


kwryankrattiger:
@cyrush

cyrush:
I understand the lib vs lib64 woes, and the resetting problem as well. 

Kokkos is an example where we had to deal with lib vs lib64 depending on the platform:
https://github.com/Alpine-DAV/ascent/blob/648d566cb3afef7378ab2e55c2c1a002054ae7ec/src/cmake/thirdparty/SetupKokkos.cmake#L20

resetting: If you re-configure you can get into a bad state with `ZZZ_DIR`. 

We have been using `ZZZ_DIR_ORIG` to get around the resetting problem, but its not perfect. 

https://github.com/Alpine-DAV/ascent/blob/648d566cb3afef7378ab2e55c2c1a002054ae7ec/src/cmake/thirdparty/SetupADIOS2.cmake#L37

I would *love* to not have to worry about these two issues: Over time, it seems like we need to move to use `ZZZ_ROOT` (path to install root) instead of `ZZZ_DIR` (what I know understand to be path to installed cmake logic) to avoid these issues?

Current changes break our tests that use cmake to build against an installed Ascent. 
They can't find the TPLs, so I think we need some adjustments, but I don't have a specific suggestion how to resolve.

kwryankrattiger:
Okay, So it seems like setting `VTKm_DIR` that way didn't work, but if I specify it under `PATHS` and `NO_DEFAULT_PATH` it does the intended behavior and searches the correct default locations.

It looks like Conduit installs the config in a non-standard location which is why it failed, so I just reverted.

VTKh was also behaving strangely, but since it seems to always install under `lib` it is also reverted.

cyrush:
Thanks, I already "fixed" the Conduit cmake path layout once, it's currently at: 

`{root_dir}/lib/cmake/conduit`

This seems to match other libs (except for lib64 vs lib issue?)

Can you tell me what the current best practice is?

Part of this is rooted in how I have kept ZZZ_DIR for years when it seems ZZZ_ROOT is the current best naming practice? I am willing to fix this for the next Ascent release if I know what the blessed path is.

kwryankrattiger:
So conduits configure should be `ConduitConfig.cmake` or `conduit-config.cmake`, but right now it is just `conduit.cmake` so it doesn't get found auto-magically by cmake.

VTK-h installs it's config under `lib`, but for auto detection it needs to be under `lib/vtkh` or `lib/cmake/vtkh` or one of the other standard locations. Being just under `lib` requires providing the full path so it still needs it.

cyrush:
@kwryankrattiger `ConduitConfig.cmake` is there, in addition to `conduit.cmake`, here is an example:

```
ls -1 uberenv_libs/spack/opt/spack/darwin-catalina-skylake/clang-11.0.0-apple/conduit-0.8.2-hvr5v2wvpkerwbv62vtvlhvtqb23blsn/lib/cmake/conduit/
ConduitConfig.cmake
ConduitConfigVersion.cmake
conduit-relwithdebinfo.cmake
conduit.cmake
conduit_setup_deps.cmake
conduit_setup_targets.cmake
```

VTK-h will be part of Ascent from now on - so that part will go away.

cyrush:
Also see the related story about when it used to be called `conduit-config.cmake`:

https://github.com/LLNL/conduit/blob/2742ca073be7ccc280615019c2c355db65a30aae/src/config/ConduitConfig.cmake.in#L5

cyrush (approved):

cyrush:
Everything looks good. Do you have any more tweaks to make?

If not, let me know and I will be happy to merge.

kwryankrattiger:
I am going to run some additional integration testing first. I will follow up shortly.

kwryankrattiger:
Finished up my testing with this, it was with with warpx using sensei/ascent and ascent in the same build of warpx. Seems to succeed now with this change and a couple others.

cyrush:
@kwryankrattiger thanks, I am going to merge. 

Also, as a heads up I am also going to merge (#919) which will change develop to require vtk-m 1.8.0.

PR #968, opened 2022-07-15
overwrite rendering bounds of dataset

nicolemarsaglia:
Allow the user to overwrite the default bounds (that are based on the dimensions of the dataset) with their own dimensions for the rendered boundary box of the dataset. A valid value is an array of six doubles that define dimensions larger than the default. If dimensions provided are smaller than the default, the default is used. 

renders/r1/dataset_bounds: [xMin,xMax,yMin,yMax,zMin,zMax]


cyrush (commented):
Looks good! Can we add a unit test to the t_ascent_render_3d tests?

nicolemarsaglia:
> Looks good! Can we add a unit test to the t_ascent_render_3d tests?

I guess I only half read this, classic -_- I made the unit test, but did not put it in t_ascent_render_3d_tests.

But the new file makes sense since I did a 2d and 3d test.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ and this project aspires to adhere to [Semantic Versioning](https://semver.org/s
 - Added Log base 2 filter. Filter type is `log2`
 - Added Feature Map in the docs. Detailing Devil Ray and VTKh features 
 - Added `scripts/build_ascent/build_ascent.sh` a script that demonstrates how to manually build Ascent and its main dependencies.
+- Added ability to overwrite dimensions for the rendered bounding box around a dataset
```
cyrush: override?

src/tests/ascent/CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ set(BASIC_TESTS t_ascent_smoke
                 t_ascent_render_3d
                 t_ascent_render_2d_poly
                 t_ascent_render_3d_poly
+		t_ascent_render_bounds
```
cyrush: There may be a tabs vs spaces thing going on here.

cyrush (approved):
Extra test is just fine as well (actually better!)

Check two minor comments and then we are good to merge.

jfavre:
Well done. It works like a charm. Thanks.

Issue #897, opened 2022-03-18
fortran flags needed for OpenAPI (IntelLLVM) Compilers


cyrush:
Like this logic:

https://github.com/Alpine-DAV/ascent/blob/fcbb50f420839f14606618f44582a76f83be8427/src/examples/proxies/cloverleaf3d-ref/CMakeLists.txt#L88

for, `IntelLLVM`, we need:

`-nofor-main`

To avoid multiple `main` methods

(Note: `-nofor_main` is what we used for other intel compilers, this will be a different case in name and detail)

Also - BLT 0.5.0 has support for `IntelLLVM` case in its macros.

cyrush:
resolved in #978

PR #978, opened 2022-08-01
support for oneapi ifx compiler

rscohn2:
-nofor_main has been deprecated for ifort, but still works
Use -nofor-main for ifx

### DIFF
--- a/src/examples/proxies/cloverleaf3d-ref/CMakeLists.txt
+++ b/src/examples/proxies/cloverleaf3d-ref/CMakeLists.txt
@@ -88,6 +88,7 @@ if(MPI_FOUND AND FORTRAN_FOUND)
     blt_append_custom_compiler_flag(
             FLAGS_VAR clover_link_flags
             INTEL   "-nofor_main"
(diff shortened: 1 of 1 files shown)


rscohn2:
Attempt to implement #897. It does not appear to work, so setting this to draft.

cyrush:
@rscohn2 thanks -- this will work for ascent@develop, however Ascent 0.8.0 has an older version of the BLT macros that doesn't support `INTELLLVM`

rscohn2:
We will do your recommended patch in the spack package. I am closing this PR. ascent@develop was getting compile errors with latest oneapi compilers so this change alone isn't useful. It is possible that it didn't work because I misconfigured. If it helps, I could add a oneapi build to https://github.com/Alpine-DAV/ascent/blob/develop/.github/workflows/build_ascent.yml. Compiler installation adds 2 minutes to the time.

cyrush:
thanks for the update, it would be amazing to have a openAPI case for our gitlab ci.

PR #980, opened 2022-08-04
Add intel compilers to ci

rscohn2:
Draft PR while debugging CI


rscohn2:
@cyrush: azure/rtd ci are triggering, but github actions is not. I will check back later. I chose not to build the third party libs with intel compilers to reduce the dependence on other projects. Feel free to push changes to this branch if you prefer a different arrangement.

rscohn2:
I will debug in my fork. I see that runs need approval and it may take a few runs to get right.

PR #981, opened 2022-08-04
Add oneapi compilers to CI

rscohn2:
Adds oneapi compilers to CI and updates build scripts for oneapi compilers. Builds  ascent and tpl with icx/icpx/ifx.

I tried building tpl with gcc and ascent with oneapi but ifx cannot use modules built with gfortran. I also tried adding icc/icpc/ifort, but the build is very slow. It did not finish after 2 hours.

It passes in my fork.

### DIFF
--- a/.github/workflows/build_ascent.yml
+++ b/.github/workflows/build_ascent.yml
@@ -6,8 +6,21 @@ on:
 
 jobs:
   build_pip_basic:
-    name: Ubuntu Build Ascent
+    strategy:
+      matrix:
+        include:
+        - cc: gcc
+          cxx: g++
+          fc: gfortran
+        - cc: icx
+          cxx: icpx
+          fc: ifx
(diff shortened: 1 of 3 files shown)


cyrush:
Thanks!

Since the TPL builds take a bit, I will build from what you have here and create a Docker container that contains pre-built oneapi TPLs. The recipes for properly installing and setting up one api that you provided are key info I need to make this happen.

cyrush (approved):
Looks good! Thanks for adding this.

rscohn2:
> Since the TPL builds take a bit, I will build from what you have here and create a Docker container that contains pre-built oneapi TPLs.

Another option is to use github actions caching. It basically tars a directory and saves it in nearby storage. You force a rebuild by changing the key. Here is an example: https://github.com/Sandia-OpenSHMEM/SOS/blob/8a54aa651e24972bec6c07ecc0ba3885e49e9fb2/.github/workflows/ci.yml#L160

Issue #982, opened 2022-08-05
camp-2022.03.0 isn't exporting BLT HIP targets


cyrush:
From ascent's using-with-cmake example:

```
/cmake-3.23.2-linux-x86_64/bin/cmake -E cmake_link_script CMakeFiles/ascent_render_example.dir/link.txt --verbose=1
[...]

PR #987, opened 2022-08-13
adjustments for hip targets

cyrush:


cyrush:
using with cmake failure here is due to: https://github.com/Alpine-DAV/ascent/issues/982
We need to test with and update to newer camp, raja, umpire. 

We still need these hip changes, so again we will wait for another PR to re-enable using with cmake

PR #994, opened 2022-08-23
regen example actions docs

cyrush:


cyrush:
This PR picks up several more Actions examples and adds them to our docs.

PR #995, opened 2022-08-24
update to conduit 0.8.4

cyrush:
FYI: I'll re-build CI containers as part of this PR.


cyrush:
We have a build issue due to spack drift  -- We accidentally boarded the spack develop train after last container updates.

Issue #1001, opened 2022-08-31
macos build issue: "no valid value for variant 'cuda'"


BenWibking:
I tried installing Ascent via uberenv on a new Mac laptop, but it fails with this error:

```
benwibking@Bens-MacBook-Pro ascent % python3 scripts/uberenv/uberenv.py --install --prefix="build"

[uberenv python: /opt/homebrew/opt/python@3.10/bin/python3.10]
[uberenv project settings:
  package_name: ascent
  package_version: develop
  package_source_dir: ../../
  spack_build_mode: uberenv-pkg
  spack_configs_path: scripts/uberenv_configs/spack_configs/configs
  spack_packages_path: scripts/uberenv_configs/packages
  spack_url: https://github.com/spack/spack
  spack_branch: develop
  spack_activate: {'py-numpy': ['+python'], 'py-mpi4py': ['+python', '+mpi'], 'conduit': ['+python'], 'py-sphinx': ['+python', '+doc'], 'py-sphinx-rtd-theme': ['+python', '+doc']}
  force_commandline_prefix: False
]
[uberenv command line options:
  install: True
  prefix: build
  spec: None
  vcpkg_triplet: None
  mirror: None
  create_mirror: False
  upstream: None
  reuse: False
  spack_config_dir: None
  vcpkg_ports_path: None
  package_name: None
  spack_build_mode: None
  package_final_phase: None
  package_source_dir: None
  project_json: /Users/benwibking/ascent/.uberenv_config.json
  build_jobs: None
  ignore_ssl_errors: False
  repo_pull: False
  spack_clean: False
  run_tests: False
  macos_sdk_env_setup: False
  setup_only: False
]
[uberenv spack build mode: install]
[skipping MACOSX env var setup]
[spack spec: @develop%apple-clang]
[info: destination '/Users/benwibking/ascent/build' already exists]
[installing to: /Users/benwibking/ascent/build]
[info: destination '/Users/benwibking/ascent/build/spack' already exists]
[spack python: /opt/homebrew/opt/python@3.10/bin/python3.10]
[disabling config scope (except defaults) in: /Users/benwibking/ascent/build/spack/lib/spack/spack/config.py]
[copying uberenv compiler and packages settings from /Users/benwibking/ascent/scripts/uberenv_configs/spack_configs/configs/darwin]
[...]

PR #1012, opened 2022-09-08
Fixes for macos uberenv spack builds

cyrush:


cyrush:
This should resolve the openmp issue  https://github.com/Alpine-DAV/ascent/issues/1001

PR #1015, opened 2022-09-14
Rewrite VTKh log to use VTK-m log, update to VTK-m 1.9

nicolemarsaglia:
This eliminates the worklets beings used by Log and updates VTK-h to use the new Log filter in VTK-m. 

This is based on VTK-m's next release (1.9?)

Also updates some deprecations and header files required by changes in VTK-m.


cyrush:
:-)

PR #1019, opened 2022-09-16
update camp in build_ascent scripts

cyrush:


cyrush:
issue: build_ascent ci -- automatic cmake find mpi for using-with-cmake-mpi example isn't working -- not sure why yet.

cyrush:
This is working in the spack + azure cases b/c HDF5 has mpi support, so MPI is begin found by Ascent implicitly there.  
I will add explicit find for MPI in the example.

PR #1024, opened 2022-09-26
Depth composite bug

nicolemarsaglia:
Vtkm is handing us negative depth values which we aren't equipped to handle. This is a change of strategy that allows negative values through.


nicolemarsaglia:
Issue: I'm losing legend annotations

output: 

tout_render_bounds_2d100

baseline: 
https://github.com/Alpine-DAV/ascent/blob/develop/src/tests/baseline_images/tout_render_bounds_2d100.png

diff: 
diff_tout_render_bounds_2d100

nicolemarsaglia:
Instead of dealing with negative values when they are encountered in the image compositor, I'm dealing with them in the Image initialization by taking the abs value. Hopefully that works.

cyrush:
Awesome, can we add a baseline check for the reproducer I added in this PR:

https://github.com/Alpine-DAV/ascent/pull/1005/

nicolemarsaglia:
> Awesome, can we add a baseline check for the reproducer I added in this PR:
> 
> #1005

Good call. Will do!

cyrush (approved):
Nice work sorting this out!

PR #1028, opened 2022-09-27
test suite report: copy images and baselines for use via relative pat…

cyrush:
…h (fix downloaded reports from CI)


cyrush:
I verified the azure artifact HTML reports look good.

Issue #1032, opened 2022-10-12
Linker warning about weak symbols


ibaned:
Hi There!

We're getting the following linker warning when linking our application, which uses all static libraries for Ascent and its dependencies:
```
ld: warning: direct access in function 'std::__1::__shared_ptr_pointer<std::__1::__empty_state<char>*, std::__1::shared_ptr<std::__1::__empty_state<char> >::__shared_ptr_default_delete<std::__1::__empty_state<char>, std::__1::__empty_state<char> >, std::__1::allocator<std::__1::__empty_state<char> > >::__get_deleter(std::type_info const&) const' from file '/Users/daibane/capp-alegra-ng/flavor/skylake/install/ascent/lib/libascent_mpi.a(ascent_actions_utils.cpp.o)' to global weak symbol 'typeinfo name for std::__1::shared_ptr<std::__1::__empty_state<char> >::__shared_ptr_default_delete<std::__1::__empty_state<char>, std::__1::__empty_state<char> >' from file '/Users/daibane/capp-alegra-ng/flavor/skylake/install/mu/lib/libmu.a(mu_datafile.cpp.o)' means the weak symbol cannot be overridden at runtime. This was likely caused by different translation units being compiled with different visibility settings.
```
Is there some visibility setting we could change in Ascent to silence this?

cyrush:
I think those style of symbols come from C++ std namespaces. 

There is nothing special going on in `ascent_actions_utils.cpp.o` so this might be hard to track down. 

We do default to hidden symbol visibility:

`set_target_properties(ascent_mpi PROPERTIES CXX_VISIBILITY_PRESET hidden)`

https://github.com/Alpine-DAV/ascent/blob/d7f41651f8251ffffb2b74845489e7fb4eb3fcae/src/ascent/CMakeLists.txt#L526

Any ideas of how we could narrow down the culprit? 

Do we also need `CMAKE_VISIBILITY_INLINES_HIDDEN`?

ibaned:
Would an option to not do hidden visibility make any sense? Or would that break Ascent in some major way?

cyrush:
We could try -- but we prefer hidden symbol visibility for VTK-m, else we get a ton of symbols exposed.

ibaned:
Is it possible to leave this up to the user so that your builds would specify
```
-DCMAKE_CXX_VISIBILITY_PRESET=ON
```
and each user can choose what visibility their whole build uses?

ibaned:
[...]

PR #1033, opened 2022-10-14
Optional symbol visibility

ibaned:
in order to avoid linker warnings
    in static application builds
    (see #1032) it is useful to
    provide users with an option
    to disable hidden visibility.
    ENABLE_HIDDEN_VISIBILITY will
    be ON by default but can be
    turned OFF by users who are
    experiencing issues with
    mixed levels of visibility.

### DIFF
--- a/src/libs/vtkh/compositing/CMakeLists.txt
+++ b/src/libs/vtkh/compositing/CMakeLists.txt
@@ -40,7 +40,9 @@ if (ENABLE_SERIAL)
 
     # build time only defs
     target_compile_definitions(vtkh_compositing PRIVATE ASCENT_EXPORTS_FLAG)
-#    set_target_properties(vtkh_compositing PROPERTIES CXX_VISIBILITY_PRESET hidden)
+    if(ENABLE_HIDDEN_VISIBILITY)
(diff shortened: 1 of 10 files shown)


ibaned:
This resolves #1032

src/libs/vtkh/compositing/CMakeLists.txt
```diff
@@ -98,7 +100,7 @@ if (MPI_FOUND)
 
     # build time only defs
     target_compile_definitions(vtkh_compositing_mpi PRIVATE ASCENT_EXPORTS_FLAG)
-    set_target_properties(vtkh_compositing_mpi PROPERTIES CXX_VISIBILITY_PRESET hidden)
+#    set_target_properties(vtkh_compositing_mpi PROPERTIES CXX_VISIBILITY_PRESET hidden)
```
cyrush: Looks good! one suggestion -- We need the conditional logic for the mpi variants of a few libs that were comment out in the first pass.

ibaned: Whoops! Let me see what that was... definitely a mistake on my part

ibaned:
I think I got the missing ones covered now!

cyrush:
Looks good, thanks for making these changes!

PR #1049, opened 2022-11-29
move ascent to src/libs

cyrush:


cyrush:
- [x] update src path mentions in docs

PR #1056, opened 2022-11-30
2022 11 warnings pass

### LINKED ISSUE #1021: remove #warning from source code
Hello,

There is a preprocessor warning line in the source code:
```
ascent/runtimes/ascent_expression_eval.cpp:1554:  #warning "Need a way to delete the intermediate results during execution"
```

Could this please be removed so it doesn't cause warnings in all user builds?

It seems like it is essentially a "TODO" note for the developers, should it perhaps be converted to an issue in an issue tracker?

Thank you!

cyrush:
resolves #1021


cyrush:
- [x] need to check the way we are using the new `std::shuffle` statement -- I expect we may be creating an engine that seeds at 0 every iteration (which would not be random :-) )

cyrush:
I checked the shuffle change -- we still get successive random sequences, so initialization strategy seems ok.

PR #1063, opened 2022-12-03
add caliper - uberenv, cmake, and smoke test

cyrush:


cyrush:
- [ ] need to update ci containers to include caliper

cyrush:
will update CI containers as part of a follow on PR.

PR #1066, opened 2022-12-28
Automatic Slice using entropy

nicolemarsaglia:
Users provide a normal, a field, and the number of levels. The levels specify the number of equally spaced slices in the direction of the normal that will be considered. The chosen slice is the one with the highest entropy. 

Serial Unit Test
MPI Unit Test
Docs


nicolemarsaglia:
Just noticed the "squash and merge" option. That's handy.

cyrush (approved):
Looks awesome, can you add a short note to the changelog and then we should merge.

nicolemarsaglia:
> Looks awesome, can you add a short note to the changelog and then we should merge.

Sounds good! Ahh the changelog, I knew I forgot something!

PR #1067, opened 2022-12-29
add vtkm check for Ascent default types

nicolemarsaglia:
check for VTKm_USE_DEFAULT_TYPES_FOR_ASCENT if this is set then in the vtkm build the use doubles is on and 64bit ids is off. 

Needs a vtkm merge to reveal this flag


cyrush:
We planned to include the default type cmake check for 0.9.2, now that we are using VTK-m 2.0, however it turns out the CMake logic isn't fully there yet. So we will need to wait for a new VTK-m patch release and address after 0.9.2

cyrush:
@nicolemarsaglia  -- should we be able to get this in with VTK-m 2.1?

nicolemarsaglia:
Yeah, let me revisit this!

nicolemarsaglia:
@cyrush if you could just double check this I would appreciate it. Should be good, but this pr is very old and cmake changes scare me.

cyrush (approved):
Looks good, let's merge!

PR #1073, opened 2023-01-07
use conduit relay mesh bp filters instead of old internal logic

cyrush:
requires a conduit release beyond 0.8.5 to pick up bug fixes in https://github.com/LLNL/conduit/pull/1062


src/libs/ascent/runtimes/flow_filters/ascent_runtime_relay_filters.cpp
```diff
@@ -77,103 +79,68 @@ namespace filters
...
 
-    NodeConstIterator itr = gather_idx.children();
-    while(itr.has_next())
+    // make sure they all have the same cycle
+    for(int i = 0; i < num_local_domains; ++i)
     {
-        const Node &curr = itr.next();
-        index_out.update(curr);
+        conduit::Node &dom = domains.child(i);
+        dom["state/cycle"] = cycle;
```
cyrush: - [x] Need to consider set external implications

PR #1074, opened 2023-01-07
Fix power/xl compile errors

tomstitt:
* POWER's char is unsigned, which caused an error: `src/libs/dray/filters/internal/marching_cubes_lookup_tables.hpp:77:41: error: constant expression evaluates to -1 which cannot be narrowed to type 'int8' (aka 'char')`
* `math.h` doesn't include the `std` variant, which caused `error: no member named 'pow' in namespace 'std';`


cyrush:
thanks @tomstitt !

cyrush (approved):

PR #1078, opened 2023-01-12
0.9.0 build updates

cyrush:


cyrush:
@nicolemarsaglia FYI: I added the `enable_find_mpi` env var option and the logic to avoid re-configure/build if the install exists to build_ascent.

cyrush:
merging successful forward progress

cyrush:
FYI: there are still more to issues to address

Issue #1079, opened 2023-01-13
build_ascent_cuda script fails to build ascent


jfavre:
I have Ampere nodes, so I could use the script as-is, with the following modules loaded:
cray-python/3.9.13.1 PrgEnv-gnu/8.3.3 /8.3.3  gcc/11.2.0 cray-mpich/8.1.21 cudatoolkit/22.3_11.6

`env enable_mpi=ON enable_openmp=OFF enable_tests=OFF build_jobs=24 enable_python=ON /users/jfavre/Projects/Ascent/build_ascent_cuda.sh`

compilations goes well for all dependencies, but Ascent itself fails to build with many lines of error similar to 

```
[  6%] Building CUDA object libs/dray/CMakeFiles/dray.dir/dray.cpp.o
[...]

PR #1087, opened 2023-01-26
fix for cray issue with raja vector intrinsics

cyrush:


cyrush:
@jfavre this build script change should fix the problem you were seeing in https://github.com/Alpine-DAV/ascent/issues/1079

I reproduced the issue on NERSC's permutter and then was able to build with this fix.

PR #1090, opened 2023-01-30
refactoring on the path to windows support

cyrush:


cyrush:
This set of work refactors ascent so we have one png util lib shared across all components.

More windows work to follow.

PR #1096, opened 2023-02-08
windows posix fixes

cyrush:


cyrush:
Note: The resource header for ascent web is now huge, due the windows strategy. 

May need to look into RLE then Base64.

Issue #730, opened 2021-05-03
template inst issue with DataWrapper<conduit::Node>


cyrush:
macOS gcc 10, this statement causes multiple symbols undermine static builds:

https://github.com/Alpine-DAV/ascent/blob/132bd6264e1e915ab1c4b8b4365ba1d2765d6fbc/src/flow/flow_data.hpp#L174

This was added when we switched to hide symbols to fix an error where python flow failed to detect types. 

Ideally we shouldn't need this, but I wonder if there is an issue with symbol visibility in the client using ascent. 

Haven't been able to duplicate, tried gcc 10 on ubuntu 20.10.

mlohry:
Hi @cyrush , I actually just ran into this on latest develop branch on MacOS with gcc-12. Any idea for a workaround?

```
[2/123] Linking CXX shared library lib/libascent_flow.dylib
FAILED: lib/libascent_flow.dylib 
[...]

PR #1111, opened 2023-02-27
explore new template symbol vis strategy

cyrush:


cyrush:
Empty decoration for base template DataWrapper may be enough to solve windows symbol vis quandary and, https://github.com/Alpine-DAV/ascent/issues/730.

If not, we will try with extra dllimport strategy.

cyrush:
closing in favor of:
#1112

PR #1114, opened 2023-03-06
Replace VTKh Vector Magnitude worklet with VTKm Vector Magnitude filter

nicolemarsaglia:
Replace VTKh Vector Magnitude worklet with VTKm Vector Magnitude filter
Add 2D vector magnitude test to unit test for vector operations


cyrush (approved):
Looks good!

PR #1116, opened 2023-03-07
Replace VTKh GhostStripper with VTKm GhostCellRemove

nicolemarsaglia:
Replace the GhostStripper worklets in VTKh with VTKm's GhostCellRemove filter


cyrush (approved):

cyrush:
Looks good!

PR #1125, opened 2023-04-07
More windows support work

cyrush:


cyrush:
oneapi compiler was failing to build vtk-m 1.9 (unrelated to windows).

I think the github actions runner ub-latest, changed to use ub 22.04 (vs 20.04). I switched back to use 20.04 which healed that test.

PR #1144, opened 2023-05-17
move to more yaml output and fix for sparse field filtering

cyrush:
- use yaml instead of json in several places
- improve blueprint verify fail error messages
- support sparse field cases for field filtering


cyrush:
resolves an issue with Nyx using field filtering

PR #1148, opened 2023-05-25
Add script to build ascent on polaris

siramok:
This script builds ascent on [polaris](https://www.alcf.anl.gov/polaris) and is based on the other HPC build scripts in scripts/build_ascent.

Of note: polaris uses nvc++ as its default C++ compiler, which fails to build conduit. Swapping from NVHPC to GNU compilers per [these instructions](https://docs.alcf.anl.gov/polaris/compiling-and-linking/compiling-and-linking-overview/#compilers-provided-by-cray-programming-environments) works around this issue and allows the build to complete as expected.


cyrush:
Thanks @siramok !

If you still have the details of the conduit vs nvc++ issue, could you share those and maybe we can resolve them in the future.

cyrush (approved):

siramok:
Sure!

I started with the default environment and compilers (using cc, CC, and ftn). Running the script hits an error when building conduit, specifically because one of conduit's dependencies (rapidjson) makes use of uint128 resulting in `error: identifier "uint128" is undefined` with nvc++.

Per [this forum thread](url), it seemed like maybe passing -Mint128 as a compiler option could help, but doing so runs into the same error.

The other thing I tried was loading newer nvhpc modules, since the default nvc++ on polaris is v21.9-0. Loading nvhpc/23.1 seems to hit a different error with a different conduit dependency (fmt-7.1.0):
```
NVC++-S-1207-Long double in GPU code is unsupported
NVC++-F-0155-Compiler failed to translate accelerator region (see -Minfo messages): Long double in GPU code is unsupported
```
Finally, loading nvhpc/23.3 fails during cmake init, stating `nvc-Error-CUDA version 11.0 is not available in this installation.`

cyrush:
@siramok merged the script updates, thanks for the details on the nvhpc compilers issues!

PR #1150, opened 2023-05-30
 vtkh stats to vtkm stats

nicolemarsaglia:
Update VTKh stats to use VTKm stats filter rather than worklets 

Relies on VTKm post 2.0 release


cyrush:
@nicolemarsaglia  should we be able to get this in once we have VTK-m 2.1?

nicolemarsaglia:
@cyrush yes we should!

cyrush (approved):

PR #1154, opened 2023-06-05
Use Cray compiler wrappers in manual Polaris builds

siramok:
Improves upon #1148, the Cray compiler wrappers are now used to build on Polaris instead of using gcc, g++, and gfortran directly. It turns out that the ALCF documentation I originally referenced for how to swap from NVHPC to GCC was not correct.

Just one caveat here: compiling RAJA with the updated script runs into [this error](https://github.com/LLNL/RAJA/issues/1464#issue-1642591604). The fix is to set `-D RAJA_ENABLE_VECTORIZATION=OFF` when building RAJA, which defaults to `ON`.

There are several ways to get around this, and I wasn't sure what the preferred way would be. I settled on making it an option that leaves the default alone, but lets me disable it for Polaris builds specifically. Would appreciate your thoughts, I'm happy to change it however you think is best.


cyrush (approved):

cyrush:
Looks good, let me know if there any more changes needed, if not I will merge.

siramok:
Cool! In that case, it doesn't need any more changes. Thanks!

cyrush:
great -- merged -- thanks for confirming,

PR #1162, opened 2023-06-26
Improvements for accessing ascent-catalyst, plus examples

siramok:
After going down a rabbit hole to get ascent-catalyst to build and subsequently render an image, this PR should make the process much easier in case others want to try it out themselves.

There are two contributions here:

1. build_ascent.sh has been modified such that catalyst can be curled and built from source like ascent's other optional dependencies. I disabled catalyst builds by default, but it can be enabled by setting `build_catalyst` to true.
2. Examples have been added that show how to use ascent-catalyst for both the non-MPI and MPI cases. These are just modified versions of the `using-with` examples, not my own original examples.

Things worth noting:

1. I move `libcatalyst-ascent.so` from ascent's install directory into catalyst's install directory at the end because that's where cmake expects to find it when linking `catalyst::catalyst`. Not performing this step causes the examples to fail. Perhaps there's some cmake magic that could be done to avoid it?
2. Custom catalyst implementations require conduit's [C API](https://github.com/LLNL/conduit/blob/1ac3851c5cf30df1cb63ea7d1491a1a4d5feaed9/src/libs/conduit/c/conduit_node.h), which unfortunately is undocumented. Fortunately it's very easy to use conduit in the usual C++ way and then cast those nodes into the correct type (seen in the examples).
3. It's not currently possible to define actions in code and pass them to ascent-catalyst as you can normally do with ascent. Only external actions files will work, perhaps an area of improvement?
4. Using ascent-catalyst + MPI currently prints on every MPI rank.


cyrush:
This is great, thank you for the build_ascent infrastructure and the examples!

This will be great for folks wanting a clear example and we will be able to add a CI test to keep this path tested.

siramok:
Sweet, thanks!

PR #1167, opened 2023-07-07
exprs refactor and reorg part 1

cyrush:


cyrush:
Provides renames for both class names and filter graph names for most expression filters.
Reorg of implementation order with better logical grouping of functionality.

cyrush:
@nicolemarsaglia FYI: Reorg with name changes, etc. 

Post this PR I will make more extensive additions.

src/libs/ascent/runtimes/ascent_expression_eval.cpp
```diff
@@ -879,27 +961,57 @@ initialize_functions()
...
 
   //---------------------------------------------------------------------------
-
+  // topo()
+  //---------------------------------------------------------------------------
   conduit::Node &topo_sig = (*functions)["topo"].append();
-  topo_sig["return_type"] = "topo";
-  topo_sig["filter_name"] = "topo";
+  topo_sig["return_type"] = "topology";
+  topo_sig["filter_name"] = "expr_mesh_topology";
```
nicolemarsaglia, line 980: Can this have the same filter name as topolgoy_sig below it (line 977)?

cyrush: yes, think of these as two aliases that map to the same filter. 

However, I typo below -- we need to set params for:

`topology_sig` instead of `topo_sig`, or else `topology_sig` isn't inited. 

The original function name was only `topo`, but I wanted us to be able to call the more explicit `topology` as well.

The return type for both has to be `topology`, b/c that is the new name I deemed moving forward for the expression language object that holds the topology.

src/libs/ascent/runtimes/ascent_expression_eval.cpp
nicolemarsaglia, line 1: Saw your "?" comments on two of the args. Not sure if you wanted to change those now or wait 
1389: field to scalar
1273: scalar to field

cyrush: Yes -  thanks, I am still working on those two.

nicolemarsaglia (approved):
the renaming is so so so so helpful

cyrush:
Getting this chunk in to open up the floor to more changes.

PR #1173, opened 2023-07-14
add blueprint extract hdf5 opts

### LINKED ISSUE #902: wire up conduit relay i/o hdf5 options to the relay extract
See: https://github.com/LLNL/conduit/issues/455

Conduit provides several options, connect compression level control, heuristic thresholds, etc 

add tests to demonstrate.

cyrush:
resolves https://github.com/Alpine-DAV/ascent/issues/902

### DIFF
--- a/src/libs/ascent/runtimes/flow_filters/ascent_runtime_relay_filters.cpp
+++ b/src/libs/ascent/runtimes/flow_filters/ascent_runtime_relay_filters.cpp
@@ -23,6 +23,7 @@
 #include <conduit_blueprint.hpp>
 #include <conduit_blueprint_mesh.hpp>
 #include <conduit_relay_io_blueprint.hpp>
+#include <conduit_relay_io_hdf5.hpp>
 
 //-----------------------------------------------------------------------------
 // ascent includes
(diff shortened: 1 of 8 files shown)


src/libs/ascent/runtimes/flow_filters/ascent_runtime_relay_filters.cpp
```diff
@@ -330,13 +330,85 @@ verify_io_params(const conduit::Node &params,
...
+                            params_hdf5_opts,
+                            info,
+                            false);
+
+        res &= check_string("chunking/compression/method",
+                            params_hdf5_opts,
+                            info,
+                            false);
+
+        res &= check_numeric("compact_storage/level",
```
nicolemarsaglia: Based on the example above and test case: 
compact_storage/level -> chunking/compression/level ?

cyrush: great catch, thank you!

nicolemarsaglia (approved):
Other than the one param things looked good.

wilsonCernWq:
Should the HDF5 path be guarded by `#if defined(ASCENT_HDF5_ENABLED)`?

cyrush:
yes, we may need some guards -- thanks for mentioning this as CI does not test Conduit w/o HDF5.

PR #1175, opened 2023-07-17
WIP: Shell commands and function callbacks

siramok:
Adds the following new capabilities to ascent:

1. Execute arbitrary shell commands, defined as ascent actions.
2. Register simulation functions with ascent so that they can be executed later as callbacks.
3. Execute arbitrary callbacks (have to be registered first), defined as ascent actions.

Example shell command usage (subject to change):
```yaml
-
  action: "add_shell_commands"
  shell_commands:
    c1:
      params:
        command: 'echo This shell command was executed by ascent > test1'
    c2:
      params:
        command: 'echo This shell command was also executed ascent > test2'
```

Example callback usage (subject to change):
```c++
void test_func() {
  ...
}

int main() {
  ...
  Ascent a;
  a.open();
  a.register_callback("test_func", test_func);
  ...
}
```
```yaml
-
  action: "add_callbacks"
  callbacks:
    c1:
      params:
        name: "test_func"
```

Still to do, we discussed:

- Allowing users to define multi-line shell commands under a single `command:`.
- Finding a different place for m_callback_map to live, other than flow_workspace.
- Allowing functions that return bool to be registered as callbacks (only void functions with no arguments will work right now).
- Dependent on the previous, letting callbacks that return bool be used directly as trigger conditions (maybe simpler if we only allow callbacks that return bool and not void in the first place?).


cyrush:
Awesome, thank you for creating this is a nice foundation!

Here are some ideas to discuss  how we can evolve:

How about "commands" as an abstraction for either a shell command or a callback?

For example:

```
-
  action: "add_commands"
  commands:
    c1:
      params:
        shell_command: 'echo This shell command was executed by ascent > test1'
    c2:
      params:
        callback: 'calbackname'
```

This approach would also work if we want to use either shell cmds/callbacks as trigger inputs in the future, for example:

``` 
-
  action: "add_triggers"
  triggers:
    t1:
      params:
        shell_command: 'echo This shell command was executed by ascent > test1'
        actions_file : "my_trigger_actions.yaml"
```

``` 
-
  action: "add_triggers"
  triggers:
    t1:
      params:
        callback: 'calbackname'
        actions_file : "my_trigger_actions.yaml"
```


I agree callback should be of the form `bool func()` to simplify both the command-like and trigger-like cases.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_callback_filters.cpp
```diff
@@ -0,0 +1,161 @@
...
+//-----------------------------------------------------------------------------
+
+#include "ascent_runtime_callback_filters.hpp"
+
+//-----------------------------------------------------------------------------
+// thirdparty includes
+//-----------------------------------------------------------------------------
+
+#include <mpi.h>
+
```
cyrush: to heal CI issues, need to guard including mpi with:

```
#ifdef ASCENT_MPI_ENABLED
...
#endif
```

cyrush:
For the callback table:

We can move callback table into the Callback filter (or a combined `Command` class that implements (Shell Cmds + Callbacks), so it does not need to be supported by the Data Flow Workspace API.

You should be able to move the static table instance and methods right over from Workspace to the Callback (or Command) class, and wire up their use in the main ascent runtime. 

Let me know how this sounds and if you have questions.

siramok:
Thanks for the comments, I like the idea of combining them into a single filter. I'll make some progress this week and ping you here when it's ready to be looked at again.

cyrush:
> Thanks for the comments, I like the idea of combining them into a single filter. I'll make some progress this week and ping you here when it's ready to be looked at again.

Sounds good!

siramok:
Ok, the PR is ready for re-review! I'm unsure why curl is failing to download conduit in the CI, but that's unrelated to my changes.

Summary:
1. The Callback and ShellCommand filters were merged into a single Command filter.
2. Callback registration and execution are now handled in Command - the Data Flow Workspace API changes were reverted.
3. Registered callbacks must now return bool.
4. Multi-line shell commands and callbacks are now supported, via the different yaml multi-line syntaxes.

I haven't yet done the work to use these directly as trigger conditions, but think that could be done as a follow-up PR. 

Multi-line shell command example:
```yaml
-
  action: "add_commands"
  commands:
    c1:
      params:
        shell_command: "echo A single-line shell command"
    c2:
      params:
        shell_command: >
          echo A single shell command
          defined on multiple lines
    c3:
      params:
        shell_command: |
          echo The first of a multi-line shell command
          echo The second of a multi-line shell command
    c4:
      params:
        shell_command: "echo The first of a \
          multi-line shell command\necho The \
          second of a multi-line shell command"
```

Multi-line callback example:
```yaml
-
  action: "add_commands"
  commands:
    c1:
      params:
        callback: "test_func1"
    c2:
      params:
        callback: |
          test_func1
          test_func2
```

Mixed example, with mpi behavior controls:
```yaml
-
  action: "add_commands"
  commands:
    c1:
      params:
        shell_command: |
          echo This is the first shell command
          echo This is the second shell command
        mpi_behavior: "all"
    c2:
      params:
        callback: |
          test_func1
          test_func2
        mpi_behavior: "root"
```

mvictoras:
Does it make sense to also add parameters to the callbacks?

siramok:
After trying to implement some use cases, I realize that callbacks should probably be allowed to return either void or bool (as opposed to only allowing one or the other). Void for cases where you want to perform an operation (like checkpointing) and bool for cases where you want to compute something and use the result in a trigger condition. It feels a bit odd to add `return true;` to a function that doesn't compute anything only so that ascent will let it be registered as a callback.

What do you think?

cyrush:
> After trying to implement some use cases, I realize that callbacks should probably be allowed to return either void or bool (as opposed to only allowing one or the other). Void for cases where you want to perform an operation (like checkpointing) and bool for cases where you want to compute something and use the result in a trigger condition. It feels a bit odd to add `return true;` to a function that doesn't compute anything only so that ascent will let it be registered as a callback.
> 
> What do you think?

Sounds good, we would just need to support the two signatures and have two maps, and logic to make sure registered name is a unique name across both.

cyrush:
@mvictoras for passing params to callbacks:

We could have a callback that passes a conduit node as a generic mechanism -- but the question of how to select what we put in that Node from Ascent isn't clear. 

Another option - if they want info from ascent, we could pass the ascent instance in the callback.

That use case could still be tricky b/c the info details won't be fully complete until all of the current actions are executed.  People could also do bad things if they try to execute new actions in the middle of current execution.

cyrush:
Everything looks good.

We should add unit tests that exercise the system command and call back cases.

I did have an idea to simplify overall: Registered callbacks are effectively static and shared across ascent instances.

Should we just make this part of the main interface as static methods?

That would cut out the plumbing and the extra logic for other runtime.


It also helps with the jupyter use cases b/c you don't need an ascent instance, it could be a static method as well (like `ascent.about()`

If we do this, it's more rework (I am happy take it on), but i wonder if what folks think.

@nicolemarsaglia  ideas?

siramok:
Moving callback registration/execution into the main interface as static methods does significantly simplify the PR, but I had trouble exposing `execute_callback` in Jupyter afterwards. I didn't bother committing that attempt, but feel free to give it a go yourself. Assuming that the Jupyter piece gets resolved, I think the rework should be a positive change.

For testing, I committed a bunch of unit tests for `register_callback` to get the ball rolling. They don't currently assert anything since `register_callback` either silently succeeds or throws an ASCENT_ERROR. Do you have advice for what to assert in that scenario?

Finally, I'll work on creating tests that actually execute some simple shell commands and callbacks. I'll also be sure to create (at least) one that demonstrates using a bool callback as a trigger condition.

cyrush:
@siramok thanks for the updates!

The basic strategy for testing callbacks is close to what you have. Use a global var that gets changed when the call back is exec-ed, and then assert we have the changed value.

Also, I am happy to take over the refactoring and get the final steps done, just let me know when you are done with your updates.

siramok:
I couldn't help myself and attempted the refactor again. I had better luck this time around and found the issue that I was having with the Python/Jupyter interface, so I committed the work!

The `register_callback` API has been moved to the top level (on par with ascent::about) and does not depend on an Ascent instance. An `execute_callback` API has been created at the top level, also independent of an Ascent instance, moving that logic out of the command filter.

Callback invocation from Jupyter will now require importing ascent explicitly, which is perfectly fine:
```python
import ascent
import conduit
ascent.execute_callback("myCallback", conduit.Node(), conduit.Node())
```

In addition to getting that refactor done, I added tests for executing callbacks, one for executing a shell command, and one that uses a bool callback as a trigger condition. The tests now reflect the API changes as well.

siramok:
I gave the changes another look and found a few minor things that needed to be cleaned up (basically just imports made unnecessary by the refactor).

Besides that, I feel that everything is good to go. The only notably missing piece I can think of is a Python test for `execute_callback`. The challenge there being that `execute_callback` requires a callback to first be registered using `register_callback`, which itself is not exposed to Python. If you have an idea for how to workaround that, please let me know.

Finally, I see that CI is failing again due to a third party dependency issue. We'll obviously need to get CI passing again, but in the meantime you're good to give it another review @cyrush. :)

Edit: worth considering, I overloaded `execute_callback` to handle both void and bool callbacks. It works, but I would understand if you'd rather have them be separate and explicit. Let me know what you think.

src/libs/ascent/ascent.hpp
```diff
@@ -66,13 +66,31 @@ class ASCENT_API Ascent
     conduit::Node  m_info;
 };
 
+// Callback maps
+static std::map<std::string, void (*)(conduit::Node &, conduit::Node &)> m_void_callback_map;
+static std::map<std::string, bool (*)(void)> m_bool_callback_map;
```
cyrush, line 71: we should define these in the cpp file, otherwise they could be multiple defined by users calling ascent.

src/libs/ascent/python/ascent_module/ascent_mpi_python.cpp
```diff
@@ -465,6 +488,72 @@ PyAscent_MPI_about()
...
+static PyObject *
+PyAscent_MPI_execute_callback(PyObject *self,
+                              PyObject *args)
+{
+    char *callback_name;
+    PyObject *py_params = NULL;
+    PyObject *py_output = NULL;
+
+    if (!PyArg_ParseTuple(args,
+                          "sOO",
```
cyrush, line 503: we can use `s|00` to make all the args optional (maybe to support the bool callback case)

src/libs/ascent/python/ascent_module/ascent_python.cpp
```diff
@@ -596,6 +596,72 @@ PyAscent_about()
...
+static PyObject *
+PyAscent_execute_callback(PyObject *self,
+                          PyObject *args)
+{
+    char *callback_name;
+    PyObject *py_params = NULL;
+    PyObject *py_output = NULL;
+
+    if (!PyArg_ParseTuple(args,
+                          "sOO",
```
cyrush, line 611: also `s|OO` possible here

src/tests/ascent/t_ascent_commands.cpp
```diff
@@ -0,0 +1,473 @@
...
+    // Run Ascent
+    //
+    Ascent ascent;
+    Node ascent_opts;
+    // default is now ascent
+    ascent_opts["runtime/type"] = "ascent";
+    ascent.open(ascent_opts);
+
+    // An error should be thrown due to not including a name
+    ascent::register_callback("", void_callback_1);
```
cyrush, line 81: we can wrap this with a google test, `EXPECT_THROW` to make sure an error is thrown

```suggestion
    EXPECT_THROW(ascent::register_callback("", void_callback_1),conduit::Error);
```

src/tests/ascent/t_ascent_commands.cpp
```diff
@@ -0,0 +1,473 @@
...
+
+    Node actions;
+    std::string msg = "An example of registering a void callback";
+                      " without a callback name.";
+    ASCENT_ACTIONS_DUMP(actions, std::string("register_no_name_void"), msg);
```
cyrush, line 86: FYI: we don't need to use this marco here b/c we aren't dumping any actions.
We use this to help seed our docs.

cyrush (approved):
Awesome, the outer level interface was exactly what I was hoping for, nice refactoring! 

I changed the target of the branch from develop to a branch in the ascent repo that we can both work on. 
I have some minor suggestions but I can tackle those. 

I will also explore how we can register python callbacks ...

cyrush:
@siramok FYI: I invited you to join the ascent repo as a contributor, keep an eye out for github request.

PR #1177, opened 2023-07-19
better quick start guidance

cyrush:
that includes build_ascent.sh


cyrush:
@nicolemarsaglia  can you review to check my new words for quick start

PR #1181, opened 2023-07-26
expand blueprint to vtk-m zero copy to support more cases (compatible strides)

cyrush:


cyrush:
closing since complete changes are in: #1161

PR #1193, opened 2023-08-22
Shell commands and function callbacks

cyrush:
Work from @siramok.

* Initial callback and shell command implementations

* Initial work to combine commands and callbacks

* Move callback registration and execution into Command

* Catch the case where neither action is defined

* Allow defining multiline shell commands and callbacks

* Minor formatting tweaks

* Initial implementation of triggers with callbacks

* Improved organization, more error catching

* Make sure that triggers either have a condition or callback

* Make void callbacks take conduit node parameters

* Initial attempt at exposing callbacks through ascent-jupyter-bridge

* Let void callbacks return arbitrary data via conduit nodes

* Disallow anonymous callbacks, add some tests

* Add tests for shell commands and callbacks

* Refactor callback API into the main API, adjusts tests to reflect the change

* Minor cleanups that I missed

* Fix minor bug with the trigger + callback test


cyrush:
@siramok  I fixed the third party lib build issue (we needed a new zlib download url)

I'll get to work on other tweaks soon.

siramok:
Awesome. I saw your feedback on the previous PR and thought that the suggestions were great.

Let me know if you'd like me to take on any of those tasks.

siramok:
Just noticed that the command tests are failing on MSVC: my guess is that it's due to `touch <file_path>` not creating a file (due to `touch` not being a thing), and subsequently failing to assert that the file was created. I believe we already have a conditional compilation directive for when the arch is Windows? A possible fix would be to use the Windows equivalent for `touch <file_path>` ifdef Windows, else use the existing command.

siramok:
Upon investigating, I realized that the tests were failing due to several things:

- A side effect of callbacks being stored in static maps was that subsequent tests would still have callbacks registered from previous tests. To fix this, I added a way to reset the callback maps and then had each test reset the callbacks before executing.
- Functions that we expect to fail are now wrapped with `EXPECT_ANY_THROW` so that the "failure" tests ultimately pass.
- I had directly copied one of the trigger tests to test using callbacks as trigger conditions, but wasn't careful to check whether the assertions were still applicable. It turns out that evaluated expressions have their results available via an info extract, but the same is not true for callbacks.
- I made the `touch` change for Windows that I mentioned above. I don't have Windows setup for building ascent, so I'll rely on the CI to tell me if it was sufficient or not.

cyrush:
@siramok thanks for the updates, I should have time next week to review further and get these features in!

src/libs/ascent/python/ascent_module/py_src/mpi/ascent_mpi.py
```diff
@@ -24,6 +24,13 @@ def about():
         raise ImportError('failed to import ascent_mpi_python, was Ascent built with mpi support?')
     return None
 
+def execute_callback(callback_name, params, output):
+    try:
+        from .ascent_mpi_python import execute_callback as ascent_execute_callback
+        return ascent_execute_callback(callback_name, params, output)
+    except ImportError:
+        raise ImportError('failed to import ascent_mpi_python, was Ascent built with serial support ENABLE_SERIAL=ON ?')
```
cyrush:
```suggestion
        raise ImportError('failed to import ascent_mpi_python, was Ascent built with MPI support ENABLE_MPI=ON ?')
```

src/libs/ascent/ascent.hpp
```diff
@@ -66,13 +66,34 @@ class ASCENT_API Ascent
     conduit::Node  m_info;
 };
 
+// Callback maps
+static std::map<std::string, void (*)(conduit::Node &, conduit::Node &)> m_void_callback_map;
+static std::map<std::string, bool (*)(void)> m_bool_callback_map;
```
cyrush: We should move this to the ascent.cpp file to avoid possible issues with multiple definitions when ascent.hpp is included by users.

src/tests/ascent/t_ascent_commands.cpp
```diff
@@ -0,0 +1,408 @@
...
+    ascent::register_callback("void_callback_1", void_callback_1);
+
+    //
+    // Run Ascent
+    //
+    Ascent ascent;
+    Node ascent_opts;
+    // default is now ascent
+    ascent_opts["runtime/type"] = "ascent";
+    ascent.open(ascent_opts);
```
cyrush: for this test case we aren't actually using the ascent instance, so we can remove creating it.

src/tests/ascent/t_ascent_commands.cpp
```diff
@@ -0,0 +1,408 @@
...
+        has_output = output["param_was_passed"].to_uint8();
+    }
+    EXPECT_TRUE(has_output);
+
+    Node actions;
+    std::string msg = "An example of invoking void callbacks directly";
+    ASCENT_ACTIONS_DUMP(actions, std::string("direct_void_callback_invocation"), msg);
+
+    ascent.close();
+}
```
cyrush, line 183: we can remove this section as well b/c we aren't using ascent instance.

cyrush (commented):
Looks great, made a few minor suggestions (which I can take care of if you like)

One Q:
I don't think we have the logic to trigger based on a shell command result yet?
If so no worries, we can tackle in a follow on PR -- just want to make sure I am reading the current code correctly.

Thanks for your effort on this, these are going to be great features!

siramok:
@cyrush Correct, it's not possible to use shell commands as trigger conditions right now.

Since your suggestions were pretty easy, I just went ahead and implemented them.

src/libs/ascent/python/ascent_module/py_src/mpi/ascent_mpi.py
siramok: I went ahead and made this change elsewhere in the file for the sake of consistency.

cyrush:
Awesome, thanks for the quick turn around!

cyrush:
CI Failure was a github fetch fluke, unrelated to changes.

cyrush:
Thanks again for this!

siramok:
Thanks!

PR #1196, opened 2023-08-31
global unique id check

nicolemarsaglia:
check for local and global unique ids


src/libs/ascent/runtimes/ascent_main_runtime.cpp
```diff
@@ -616,24 +615,77 @@ AscentRuntime::EnsureDomainIds()
...
+    {
+      if(domains_per_rank[i] > max_num_domains)
+        max_num_domains = domains_per_rank[i];
+    }
     delete[] domains_per_rank;
 #endif
+
+    std::unordered_set<int> local_unique_ids;
+
+    int *domain_ids = new int[max_num_domains];
```
cyrush: This new may be mis matched with free if an exception is thrown.

We could  use a conduit node to hold these arrays so they are automatically cleaned up for any error cases.

src/libs/ascent/runtimes/ascent_main_runtime.cpp
```diff
@@ -616,24 +615,77 @@ AscentRuntime::EnsureDomainIds()
...
+      else
+      {
+         domain_ids[i] = dom["state/domain_id"].to_int32();
+         local_unique_ids.insert(domain_ids[i]);
+      }
+    }
+
+    if(local_unique_ids.size() != num_domains)
+    {
+      ASCENT_ERROR("Local Domain IDs are not unique ");
```
cyrush: In the mpi case, this would hang if one mpi task has bad local ids. (before getting deeper into mpi logic)

We need to add mpi error checking (flag all is ok, etc)

src/libs/ascent/runtimes/ascent_main_runtime.cpp
```diff
@@ -616,24 +615,77 @@ AscentRuntime::EnsureDomainIds()
...
+#ifdef ASCENT_MPI_ENABLED
+
+    int total_domains = 0;
+    MPI_Allreduce(&num_domains, &total_domains, 1, MPI_INT, MPI_SUM, mpi_comm);
+    for(int i = num_domains; i < max_num_domains; i++)
+    {
+      domain_ids[i] = 0;
+    }
+
+    int *domain_ids_per_rank = new int[comm_size*max_num_domains];
```
cyrush: Is max_num_domains - the max domains per mpi rank?

Could we just directly count number of domain trees per rank and alloc exactly? 
We could then:
- init an array to all -1s
- use per mpi task offset and fill with local domain ids
- use an MPI MAX all reduce to get the full set of domain ids

Then we could unordered set (or even an in place sort + scan) to detect dups.

If we are in bad shape, we should provide which rank(s) have duplicate ids in the error message.

cyrush (commented):
@nicolemarsaglia  a few questions and ideas.

Also, we should add test cases that trigger both local and global error.

src/libs/ascent/runtimes/ascent_main_runtime.cpp
nicolemarsaglia: max_num_domains is the max number of domains of any rank, since there might be a mismatch of number of domains between ranks, I wanted to standardize a passing size. 

But your way works better, I'll change it.

src/libs/ascent/runtimes/ascent_main_runtime.cpp
nicolemarsaglia: I think I ended up rewriting the code enough that I don't hit this case anymore. But for future reference, is this what you meant?

```
node n;
int *darray = new int[number];
n["array"].set_external(darray);
if(case) ASCENT_ERROR("info"); //hit this, no delete
delete[] darray;

```
darray will still get cleaned up by conduit. Is my understanding correct?

cyrush: Here is what I would suggest:

```
Node n;
n.set(DataType::c_int(number));
int *dptr = n.value();
// use dptr
// ptr will be cleaned up when n goes out of scope
```

src/tests/ascent/t_ascent_mpi_unique_ids.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+    Ascent ascent;
+
+    Node ascent_opts;
+    ascent_opts["runtime"] = "ascent";
+    ascent_opts["mpi_comm"] = MPI_Comm_c2f(comm);
+    ascent.open(ascent_opts);
+
+    EXPECT_THROW(ascent.publish(data),conduit::Error);
+}
```
nicolemarsaglia, line 84: Lol. It throws an error but doesn't catch it. @cyrush have any guesses? I tested out throwing other errors from EnsureDomains, like consistent IDs and it is also not catching them https://github.com/Alpine-DAV/ascent/blob/e52b7bb8c9fd131f2fd49edf58037cc5ef77a166/src/libs/ascent/runtimes/ascent_main_runtime.cpp#L554

```

[ RUN      ] ascent_partition.test_indiv_rank_non_unique
[Error] Ascent::publish
file: /sw/summit/ums/ums010/2023_01/frontier/debug_build/ascent/src/libs/ascent/runtimes/ascent_main_runtime.cpp
line: 670
message:
Local Domain IDs are not unique on ranks: 0

Testing local non unique IDs/sw/summit/ums/ums010/2023_01/frontier/debug_build/ascent/src/tests/ascent/t_ascent_mpi_unique_ids.cpp:85: Failure
Expected: ascent.publish(data) throws an exception of type conduit::Error.
  Actual: it throws nothing.
[  FAILED  ] ascent_partition.test_indiv_rank_non_unique (1207 ms)
```

```
[ RUN      ] ascent_partition.test_indiv_rank_non_unique
[Error] Ascent::publish
file: /sw/summit/ums/ums010/2023_01/frontier/debug_build/ascent/src/libs/ascent/runtimes/ascent_main_runtime.cpp
line: 614
message:
Inconsistent domain ids: all domains must either have an id or all domains do not have an id

Testing local non unique IDs/sw/summit/ums/ums010/2023_01/frontier/debug_build/ascent/src/tests/ascent/t_ascent_mpi_unique_ids.cpp:85: Failure
Expected: ascent.publish(data) throws an exception of type conduit::Error.
  Actual: it throws nothing.
[  FAILED  ] ascent_partition.test_indiv_rank_non_unique (1205 ms)
```

cyrush: By default, Ascent will not throw beyond the interface (we don't want to crash client codes).

You need to use:
```
ascent_opts["exceptions"] = "forward";
```

To allow them to be thrown outside of Ascent.

nicolemarsaglia: Ahh makes sense! Thanks!

cyrush:
@nicolemarsaglia awesome, thanks for getting this in!

Issue #1155, opened 2023-06-07
PNG Metadata


WillTrojak:
Would it be possible to add something to the scene API that allows metadata to be written to PNGs? Additionally, would it be possible for a flag that causes a map from value to colourmap colour to be written to the metadata? 

The idea of this is twofold, it would allow for PNG files to be more accurately post-processed, with the colour map being handy for training ML/dataset curation. The second thing would be to improve reproducibility, for example in PyFR we would like to write parts of the config/inifile and stats to the png.

WillTrojak:
A similar feature is now supported in Paraview: https://www.kitware.com/paraview-state-embedded-in-png-screenshots/

nicolemarsaglia:
This is a great idea! I had previously looked into adding more info to PNGs, but it was not as straightforward as I had hoped, so I tabled it. 

I will look into this again!

nicolemarsaglia:
Looking into this, in the meantime, one thing Ascent does have in an info() call in the API that populates a conduit node with metadata rather than the PNG, and can be used in a similar way: https://ascent.readthedocs.io/en/latest/AscentAPI.html?highlight=info#info

brugger1:
I added the ability to add meta data to PNGs with the following PR.

https://github.com/Alpine-DAV/ascent/pull/746

If you add specific conduit nodes under the state, it will add those strings to the PNG meta data. Specifically:

state/software
state/source
state/title
state/info
state/comment

nicolemarsaglia:
Thanks @brugger1 

edit: Definitely needs to be documented.

brugger1:
Yes, the PR didn't have any documentation associated with it. :(

WillTrojak:
This is great, thanks @brugger1. I'll give it a go!

cyrush:
@WillTrojak  if you are creating images for ML datasets, you might be interested in scalar rendering -- which you can use to save an image out to a HDF5 file and preserve the floating point data. 


https://ascent.readthedocs.io/en/latest/Actions/Examples.html?highlight=scalar#an-example-of-scalar-rendering

WillTrojak:
I was looking into this for partitioned domains and found that the metadata implemented [here](urlhttps://github.com/Alpine-DAV/ascent/blob/e52b7bb8c9fd131f2fd49edf58037cc5ef77a166/src/libs/ascent/runtimes/ascent_main_runtime.cpp#L511) only works for flat data structures ie those such as:
```
state:
[...]

PR #1199, opened 2023-09-08
Added generalisation of png metadata handler

WillTrojak:
This is in answer to #1155.

The intention is to generalise the handling of setting metadata in the PNG files. I am open to changes in the syntax that I've used or which domain is chosen to set the metadata from.

At the moment, the syntax that I've used is that for the first child domain, any child of the `state` node that has a child called `keyword` and `data` will create a png metadata keyword. ie
```
domain_0:
   state:
      domain_id: 0
      time: 
         keyword: 'Time'
         data: 0.1
domain_1:
   state:
      domain_id: 1
      comment: 
         keyword: 'Comment'
         data: 'some comment'
```

Will give the png metadata:
```
Time: 0.1
```

This seemed like the cleanest way of handling this.

### DIFF
--- a/src/libs/ascent/runtimes/ascent_main_runtime.cpp
+++ b/src/libs/ascent/runtimes/ascent_main_runtime.cpp
@@ -1219,20 +1219,22 @@ AscentRuntime::PopulateMetadata()
   const int num_domains = m_source.number_of_children();
   int cycle = -1;
   float time = -1.f;
+  bool metadata_set = false;
 
   for(int i = 0; i < num_domains; ++i)
   {
     const conduit::Node &dom = m_source.child(i);
-    if (i == 0 && dom.has_path("state"))
+    if (!metadata_set && dom.has_path("state"))
     {
       m_comments.reset();
       conduit::NodeConstIterator itr = dom["state"].children();
 
       while(itr.has_next())
       {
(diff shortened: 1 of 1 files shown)


cyrush:
@WillTrojak  good catch and nice suggestion!

They way ascent is currently plumbed, PNG comments are only used MPI task 0, (since task 0 writes the final pngs)

For the general case, it's possible that task 0 may have no domains at all. This is rare but we need to support it. 

So we will need an extension to your code to find the first task that has this info, and then get that info to Task 0 (or all tasks).

That will require some MPI coordination. I can help implement this, but won't have time until later next week.

FreddieWitherden:
> For the general case, it's possible that task 0 may have no domains at all. This is rare but we need to support it.

This is quite likely in the PyFR use case as we often only pass Ascent a small subset of the total computational domain.  As such it is not at all unusual for ranks to be excluded and have no data of their own (but they all make collective calls to Ascent).

WillTrojak:
I added a simple test so that the first domain that has metadata is used. Let me know if you forsee any issues with this.

cyrush (approved):
Yes, this looks good. I think we should get this in now. 

Sorry for the delay on the MPI version of the checks. 
I can follow up with those changes later.

PR #1217, opened 2023-10-31
add support for bp implicit points topo

cyrush:


jfavre:
Thanks @cyrush . Just tested the new implicit points topology and it works for our needs.

cyrush:
Thanks for trying it out @jfavre 

VTK-m doens't have a way to make this truly implicit - so at the end of the day the we are creating the connectivity in Ascent. But I wanted to make sure to support this since Blueprint provides the option.

Issue #1221, opened 2023-11-04
zlib linking issue on macOS: dyld[28998]: Library not loaded: libz.1.dylib


BenWibking:
zlib does not appear to be linked correctly when building with the `build_ascent.sh` script. At runtime, I get errors like this:
```
dyld[28998]: Library not loaded: libz.1.dylib
  Referenced from: <B0A9E967-42DD-36D3-83DD-46915447138D> /Users/benwibking/athenapk/build/bin/athenaPK
  Reason: tried: 'libz.1.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OSlibz.1.dylib' (no such file), 'libz.1.dylib' (no such file), '/Users/benwibking/athenapk/libz.1.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/Users/benwibking/athenapk/libz.1.dylib' (no such file), '/Users/benwibking/athenapk/libz.1.dylib' (no such file)
```

This can be worked around by adding the path to zlib to `DYLD_LIBRARY_PATH`, e.g.:
```
 export DYLD_LIBRARY_PATH=$HOME/ascent_build/install/zlib-1.2.13/lib:$DYLD_LIBRARY_PATH
```

I built with:
```
env enable_mpi=ON ./build_ascent.sh
```
with
```
Apple clang version 15.0.0 (clang-1500.0.40.1)
Target: arm64-apple-darwin23.1.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

cyrush:
Thank you, and sorry for this build issue.  I just hit this last week on my new mac & I am looking into how to solve.

-Cyrus
________________________________
From: Ben Wibking ***@***.***>
Sent: Saturday, November 4, 2023 12:13:30 PM
To: Alpine-DAV/ascent ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [Alpine-DAV/ascent] zlib linking issue on macOS: dyld[28998]: Library not loaded: libz.1.dylib (Issue #1221)


zlib does not appear to be linked correctly when building with the build_ascent.sh script. At runtime, I get errors like this:

dyld[28998]: Library not loaded: libz.1.dylib
  Referenced from: <B0A9E967-42DD-36D3-83DD-46915447138D> /Users/benwibking/athenapk/build/bin/athenaPK
[...]

PR #1222, opened 2023-11-06
rework zlib cmake logic, add smoke tests, update build_ascent to use …

cyrush:
…zlib 1.3


cyrush:
resolves #1221, need to track down windows zlib smoke test issue

cyrush:
windows issue, I found errors in the smoke test (not windows specific) and resolved those, now we are in good shape

PR #1224, opened 2023-11-10
container refactor

cyrush:
update container naming to better manage multiple architectures (x86_64 and amr6))


cyrush:
rocm containers download cmake binaries, we need to add logic to select proper binary (x86_64 vs aarch64)

https://github.com/Alpine-DAV/ascent/blob/task/2023_11_container_updates/scripts/ci/docker/alpinedav_ubuntu_20.04_rocm_5.1.3_devel_tpls/docker_uberenv_build.sh

PR #1238, opened 2023-12-13
raja binning fixes

cyrush:
fix raja binning implementation and switch to use it 
add more tests


cyrush:
Screenshot 2024-04-19 at 11 46 04 AM

RAJA Binning is not creating the proper input samples,  it seems like the centroids are way off. 
Above is a `count` case, elements in each output column should have the same value. The outer columns should match and the inner column should have higher values. 

Input is a 6x6x6 cube. Purple shows the bin boundaries. Output bins are a 3x6x1 mesh.

cyrush:
Screenshot 2024-04-19 at 12 13 14 PM

cyrush:
sample points are bogus

cyrush:
Screenshot 2024-04-19 at 12 15 35 PM

trying different ordering, still bogus

cyrush:
there are still issues with the new raja implementation, but this is good forward progress.  Old binning implementation is wired up with new tests.

PR #1239, opened 2023-12-21
add tubes for rendering streamlines

nicolemarsaglia:
Apply the VTK-m Tube filter to the output of the Streamline filters to allow for rendering


src/libs/ascent/runtimes/flow_filters/ascent_runtime_vtkh_filters.cpp
```diff
@@ -3831,47 +3916,389 @@ VTKHParticleAdvection::execute()
...
+              dy = dist_y/(num_seeds_y-1);
+	    else
+              dy = dist_y/num_seeds_y;
+	  if(num_seeds_z != 0)
+	    if(num_seeds_z != 1)
+              dz = dist_z/(num_seeds_z-1);
+	    else
+              dz = dist_z/num_seeds_z;
+ 
+	  double epsilon = 0.1;
```
nicolemarsaglia: I was struggling to render seeds that are directly on the boundary, so I've been offsetting all my points my epsilon. 

Note for later for picking a better epsilon: 
      vtkm::Float64 lx = coordBounds.X.Length();
      vtkm::Float64 ly = coordBounds.Y.Length();
      vtkm::Float64 lz = coordBounds.Z.Length();
      vtkm::Float64 mag = vtkm::Sqrt(lx * lx + ly * ly + lz * lz);
      // same as used in vtk ospray
      constexpr vtkm::Float64 heuristic = 1000.;
      radius = static_cast<vtkm::Float32>(mag / heuristic);
      m_tube_size = radius;

cyrush: So I understand: Is it that the box is exactly aligned a data set boundary causing numeric sadness?

If so, the user could adjust the box to not be aligned.   Box params being correct is something user can control.

nicolemarsaglia: @cyrush 

Exactly, anything on the boundary needs to be moved slightly or else we lose the particles. 

I guess the check should be if the particle is on the boundary and then adjust, rather than adjusting ALL points like I am. 

Here are boundary points that are not and are adjusted using epsilon, respectively. 
box_boundary_uniform_0000
box_boundary_uniform_0001

cyrush:
looks great, can we change the file names of the tests to be a more descriptive:

for example, from:

`src/tests/ascent/line_seeds.cpp`

to

`src/tests/ascent/t_ascent_particle_advection_line_seeds.cpp`

nicolemarsaglia:
> looks great, can we change the file names of the tests to be a more descriptive:
> 
> for example, from:
> 
> `src/tests/ascent/line_seeds.cpp`
> 
> to
> 
> `src/tests/ascent/t_ascent_particle_advection_line_seeds.cpp`

oops,  I meant to delete these. But they should also pass the windows tests no problem. I'll add them . Good catch!

src/tests/vtkh/t_vtk-h_warpx_streamline.cpp
```diff
@@ -180,15 +155,34 @@ TEST(vtkh_particle_advection, vtkh_serial_particle_advection)
...
+//  vtkh::RayTracer tracer;
+//  tracer.SetInput(outWSL);
+//  tracer.SetField("streamlines");
+//  std::string fieldName = "streamlines";
+//
+//  vtkh::Scene scene;
+//  scene.AddRender(render);
+//  scene.AddRenderer(&tracer);
+//  scene.Render();
+//
```
cyrush, line 187: Should we turn this back on?
If so, lets add a ticket to track.

src/tests/vtkh/t_vtk-h_warpx_streamline_par.cpp
```diff
@@ -0,0 +1,195 @@
...
+//                                         "tout_warpx_streamline_render");
+//
+//  vtkh::RayTracer tracer;
+//  tracer.SetInput(outWSL);
+//  tracer.SetField("streamlines");
+//
+//  vtkh::Scene scene;
+//  scene.AddRender(render);
+//  scene.AddRenderer(&tracer);
+//  scene.Render();
```
cyrush, line 191: Should we turn this back on?

cyrush (approved):
Looks good, lets get this merged!

I added two comments, if we want to follow up on those lets create a ticket to remember.

Thanks for getting this all working!

nicolemarsaglia:
> Looks good, lets get this merged!
> 
> I added two comments, if we want to follow up on those lets create a ticket to remember.
> 
> Thanks for getting this all working!

Sounds good! I'll make a ticket for the warpx tests

PR #1243, opened 2024-01-04
Only use HDF5 if conduit was compiled against it

szellmann:
Fixes compilation w/o HDF5.


cyrush:
@szellmann thanks for these fixes! Running CI and pending that we can get them in.

cyrush (approved):

PR #1247, opened 2024-01-23
Add VTKm Probe filter for regular grid resampling

nicolemarsaglia:


src/libs/vtkh/vtkm_filters/vtkmProbe.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+  std::string name = "coords";
+  vtkm::Id3 dims = 3;
+  if(m_dims[2] == 0)
+    dims = 2;
+
+  //move origin to lower left corner
+  m_origin[0] = m_origin[0] - (m_dims[0]/2);
+  m_origin[1] = m_origin[1] - (m_dims[1]/2);
+  m_origin[2] = m_origin[2] - (m_dims[2]/2);
```
cyrush: I want to make sure I understand the origin shift. 
Does VTK-m create Uniform Datasets from the center of the Volume?

src/libs/vtkh/filters/SampleGrid.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+void
+SampleGrid::DoExecute()
+{
+  this->m_output = new DataSet();
+
+  const int num_domains = this->m_input->GetNumberOfDomains();
+
+  for(int i = 0; i < num_domains; ++i)
+  {
```
cyrush: I think we need more logic here to support the distributed memory case. 

If all ranks are doing the resample to the same grid, we would have do a reduction to compose the results on each task. 

Also, we may want to distribute the output grid across MPI tasks as well for high res resampling.

src/libs/vtkh/vtkm_filters/vtkmProbe.cpp
nicolemarsaglia: VTKm creates uniform datasets from the lower left corner, whereas conduit starts in the middle (and I'm copying that format), so I shifted middle to lower left. Does that make sense? 

https://vtk-m.readthedocs.io/en/latest/dataset.html#creating-uniform-grids

cyrush: In general, Conduit also starts with an origin + dx,dy,dz. Which is lower left corner with respect to an overall coordinated system. 

That said the blueprint example you are using may be centering things at 0 and have volume that covers both negative and positive parts of the overall coordinate system.

Want to make sure we aren't confusing a specific case for testing with what we need for the general case.

nicolemarsaglia: Ahh ok, that's good to know! Must have just saw this special case and assumed it was the default. Let me add some tests with different input data.

src/libs/vtkh/filters/SampleGrid.cpp
nicolemarsaglia: Do we have to compose the results? If each rank is responsible for it's own portion of the data, they will either have a point that gets sampled or they won't, and in the end they are still in charge of their own portion of the new grid. 

I have a concern about needing to do a large resample, right  now we have to create the global grid on each rank.

nicolemarsaglia: I will add a parallel test

cyrush: yes - I think so.  Let's chat about the cases high bandwidth. 

To test this you can use the 2D spiral mulit-domain blueprint example and dish out domains to different mpi tasks.

nicolemarsaglia:
Will require updated VTK-m. (2.2 or 2.1.1)

src/libs/vtkh/filters/UniformGrid.cpp
```diff
@@ -0,0 +1,759 @@
...
+  /// for each domain `d`:
+  ///   reset local_res_tmp (including making info)
+  ///   execute probe filter on `d` with output in `local_res_tmp`
+  ///   combine results from `local_res_tmp` into `local_res`
+  /// if mpi parallel:
+  ///.  Use global reduce w/ `local_res` to create `global_res`
+  ///
+  ///.  (We know that all ranks will have something to reduce, b/c
+  ///.   even if they have no domains, the still created the local
+  ///    output grid)
```
cyrush, line 975: @nicolemarsaglia  see note here of how we can accomplish this w/o MergeDataset -- allowing us to use vtk-m 2.1
Also, this would be more memory efficient b/c merging will create a fused unstructured grid which can require a lot more space if the input was uniform (explicit coords and connectivity, maybe copying and repacking of field arrays)

cyrush:
@nicolemarsaglia  this looks good to me, we can discuss today!

PR #1250, opened 2024-02-07
Add support for customizable installation directory

mvictoras:
This pull request modifies the existing bash script for compiling and installing Ascent to introduce flexibility in the installation directory. Previously, the script installed Ascent at `${root_dir}/install` by default. However, with this change, users can now define their own installation directory using the environment variable `${install_dir}`.

By allowing users to specify their preferred installation path, this update enhances the usability of the script and accommodates diverse user requirements. Whether installing Ascent for personal projects or system-wide use, users now have the freedom to choose where Ascent is installed, providing greater control over their development environments.

### DIFF
--- a/scripts/build_ascent/build_ascent.sh
+++ b/scripts/build_ascent/build_ascent.sh
@@ -132,9 +132,13 @@ fi
 
 cd ${root_dir}
 
+# install_dir is where we will install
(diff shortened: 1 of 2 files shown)


cyrush (approved):
looks good, thanks for the improvement

PR #1256, opened 2024-02-22
add output path for cinema databases

nicolemarsaglia:
caveat: output path needs to exist prior to running


src/libs/ascent/runtimes/flow_filters/ascent_runtime_rendering_filters.cpp
```diff
@@ -1171,6 +1171,11 @@ DefaultRender::execute()
 	  if(render_node.has_path("output_path"))
 	  {
             output_path = render_node["output_path"].as_string();
+            // create a folder if it doesn't exist
+            if(!conduit::utils::is_directory(output_path))
+            {
+              conduit::utils::create_directory(output_path);
+            }
```
cyrush, line 1183: +1

But we need to make to only do this on rank-0 (don't want the mpi tasks to fight to create the dir :-) )

nicolemarsaglia: right right! thanks for the reminder

cyrush (approved):
looks good!

MinSizeRel seemed to get us under the space ceiling.

PR #1277, opened 2024-04-17
add low order refine option to the relay extract

cyrush:


src/tests/ascent/t_ascent_relay.cpp
```diff
@@ -988,6 +988,77 @@ TEST(ascent_relay, test_relay_sparse_topos)
...
+    std::cout << actions.to_yaml() << std::endl;
+
+    //
+    // Run Ascent
+    //
+    Ascent ascent;
+    ascent.open();
+    ascent.publish(data);
+    ascent.execute(actions);
+    ascent.close();
```
nicolemarsaglia, line 1036: Do we want a EXPECT_TRUE file check for the first extract?

cyrush: thanks, done!

nicolemarsaglia (approved):
Looks good!

PR #1286, opened 2024-05-09
simplify kokkos init logic

### LINKED ISSUE #1127: are we double initing?
https://github.com/Alpine-DAV/ascent/blob/86ff9aafa89b859457d0850b5732a7115489f5b5/src/libs/ascent/runtimes/ascent_main_runtime.cpp#L212

cyrush:
resolves #1127


nicolemarsaglia (approved):
Ooo real nice, I like we can select a device now!

PR #1303, opened 2024-05-22
Enable kokkos with cuda

nicolemarsaglia:


cyrush (approved):
Looks good, the test failure is unrelated to your changes. 

I opened #1304  to track that failure.

nicolemarsaglia:
Current super-fun-time, no-good error:
[lulesh_cuda_break.txt](https://github.com/Alpine-DAV/ascent/files/15408541/lulesh_cuda_break.txt)

cyrush:
Maybe we are missing some sources that need to be tagged as device sources?

https://github.com/Alpine-DAV/ascent/blob/baa7ec2af97d074d28c4e1bb3e3a7586de481027/src/libs/ascent/CMakeLists.txt#L248

I thought we cover all of those cases via our static cuda build, but maybe using kokkos with VTK-m opens up new cases?

nicolemarsaglia:
> Maybe we are missing some sources that need to be tagged as device sources?
> 
> https://github.com/Alpine-DAV/ascent/blob/baa7ec2af97d074d28c4e1bb3e3a7586de481027/src/libs/ascent/CMakeLists.txt#L248
> 
> I thought we cover all of those cases via our static cuda build, but maybe using kokkos with VTK-m opens up new cases?

`/usr/tce/packages/gcc/gcc-9.3.1/rh/usr/bin/../libexec/gcc/ppc64le-redhat-linux/9/ld: ../../../lib/libascent.so: undefined reference to '__cudaRegisterLinkedBinary_56_tmpxft_0001a5d5_00000000_7_ascent_memory_manager_cpp1_ii_5b8ab178'`

If it was about missing sources then this file wouldn't be failing, right? And the files the errors reference seem like a (deterministic) grab bag.

PR #1324, opened 2024-07-08
bugfix for vtk-m ray culling issue

cyrush:
Adds vtk-m patch for a bug causing rays to be incorrectly culled before tracing.

The issue is that the near and far planes of the camera are not properly used in the culling logic. 

This could be encountered when ray tracing very small spatial domains from a camera far away.

Adds test case that to ensure that exercises the issue.


cyrush:
FYI: Azure test cases will fail new test until we update docker containers with patched vtk-m.

PR #1326, opened 2024-07-09
VTK-m MIR

nicolemarsaglia:
currently does not support zero-copy


src/libs/ascent/runtimes/flow_filters/ascent_runtime_vtkh_filters.cpp
```diff
@@ -5246,6 +5247,133 @@ VTKHVTKFileExtract::execute()
...
+    std::string topo_name = collection->field_topology(ids_name);
+
+    vtkh::DataSet &data = collection->dataset_by_topology(topo_name);
+    if(!graph().workspace().registry().has_entry("is_discrete"))
+    {
+      
+      int *is_discrete = new int();
+      *is_discrete = 1;
+      graph().workspace().registry().add<int>("is_discrete",
+                                               is_discrete,
```
cyrush: wondering if there is another way to do this hand shake?

CHANGELOG.md
```diff
@@ -38,13 +38,16 @@ and this project aspires to adhere to [Semantic Versioning](https://semver.org/s
 - Added a `vtk` extract that saves each mesh domain to a legacy vtk file grouped, with all domain data grouped by a `.visit` file.
 - Added particle advection for streamline and related rendering support.
 - Added WarpX Streamline filter that uses charged particles.
+- Added Uniform Grid Sampling filter.
+- Added Material Interface Reconstruction (`mir`) filter which produces `cellMat` output field.
```
cyrush: Instead of cell mat, could we use the matset name as the field name?

If there are multiple matsets, one name could lead to a collision.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_vtkh_filters.cpp
nicolemarsaglia: Stripped this out and now it's a param for pseudocolor plots

CHANGELOG.md
nicolemarsaglia: Got this changed

cyrush (approved):
Looks good, awesome to get this wired up!

PR #1330, opened 2024-07-11
Fixed mpi communication from MPI_COMM_WORLD to the one defined in ascent

mvictoras:
I believe the MPI communicator here should be mpi_comm and not MPI_COMM_WORLD.


nicolemarsaglia:
Good catch! Thanks for the fix @mvictoras

cyrush:
@mvictoras  thank you , I did a pass to fix a group of these, but I guess I didn't catch them all - thanks for the PR!

Issue #1202, opened 2023-10-12
read the docs search is not working


cyrush:
Same deal as conduit setup:

https://github.com/LLNL/conduit/issues/1176

cyrush:
resolved in #1334

PR #1334, opened 2024-07-18
rtd config: add jquery to req extensions

cyrush:


cyrush:
resolves #1202 -- heals read the docs so that search works.

Solution was simple, should have tackled sooner.

PR #1344, opened 2024-07-30
build_ascent caliper support

cyrush:


cyrush:
Note: Windows will need patches for caliper.

We are going to focus on linux + rocm first.

PR #1351, opened 2024-08-03
add siramok terminal steering example 

cyrush:


cyrush:
@siramok  FYI: I am merging your example into develop, we can always refine more in the future.

PR #1359, opened 2024-08-13
build_ascent caliper + hip support

cyrush:


cyrush:
@nicolemarsaglia @emily-howell  after we merge this PR, you can use `scripts/build_ascent/build_ascent_hip_rzvernal.sh` to create a build on rzvernal that includes caliper + rocm support.

PR #1366, opened 2024-08-21
Replay parse actions before loading data

### LINKED ISSUE #1317: replay -- parse actions before reading data
best to know if you have malformed actions before you spend effort doing I/O.

emily-howell:


emily-howell:
I'm not 100% sure that this is the direction that #1317 intended. I know there is the "CheckForSettingsFile" function in ascent.cpp, I was hesitant to expose that function but alternatively could and just use that.

A few questions on this code:
1) Are there additional steps that should be added if ASCENT_REPLAY_MPI is defined. I am a little rusty on my MPI so some feedback there would be appreciated.
2) There aren't any existing tests for the replay functionality as far as I can see. Should I add something for this somewhere?

src/utilities/replay/replay.cpp
```diff
@@ -156,6 +156,66 @@ void trim(std::string &s)
...
+    return actions;
+  }
+  MPI_Comm mpi_comm = MPI_Comm_f2c(mpi_comm_id);
+  MPI_Comm_size(mpi_comm, &comm_size);
+  MPI_Comm_rank(mpi_comm, &rank);
+#endif
+  
+  bool has_file = false;
+
+  if(conduit::utils::is_file(file_name)){
```
cyrush: We should only check if the file exists on rank 0.

src/utilities/replay/replay.cpp
```diff
@@ -156,6 +156,66 @@ void trim(std::string &s)
...
+#ifdef ASCENT_MPI_ENABLED
+    MPI_Bcast(&actions_file_valid, 1, MPI_BOOL, 0, mpi_comm);
+#endif
+
+  if(!actions_file_valid){
+    // Raise Error
+    ASCENT_ERROR("Failed to load actions file: " << file_name
+                  << "\n" << file_load_error_msg);
+  }
+
```
cyrush, line 222: If successful loading the file on rank 0, we need to broadcast the actions to all other ranks.
Conduit's relay MPI allows us to broadcast a node.

src/utilities/replay/replay.cpp
```diff
@@ -156,6 +156,66 @@ void trim(std::string &s)
              s.end());
 }
 
+conduit::Node load_actions(const std::string &file_name, int mpi_comm_id){
+  conduit::Node actions;
```
cyrush: We prefer passing in a ref to the node and having node filled out instead of returning a node.

This is b/c returning a node involves more copies.

src/utilities/replay/replay.cpp
```diff
@@ -197,16 +257,16 @@ main(int argc, char *argv[])
...
 #if defined(ASCENT_REPLAY_MPI)
   ascent_opts["mpi_comm"] = MPI_Comm_c2f(MPI_COMM_WORLD);
 #endif
 
   //
-  // Blank actions to be populated from actions file
+  // Populate actions with actions file
   //
-  conduit::Node actions;
+  int mpi_comm = ascent_opts.has_child("mpi_comm") ? ascent_opts["mpi_comm"].to_int() : 0;
```
cyrush: Should the case were we don't have a comm be `-1`, to trigger the right error conditions?

cyrush (commented):
You are on the right track. We don't want to expose the internal method, but we want a similar set of logic here. 

The goal is to read the actions on rank 0, and then broadcast to others so all tasks have the same actions passed to `ascent.execute`.  You have most of the logic in place -- added suggestions for next steps. 


We don't have any tests that run replay wired up into our build system.

We should explore adding those, I can help you with the steps.

src/utilities/replay/replay.cpp
```diff
@@ -156,9 +156,7 @@ void trim(std::string &s)
              s.end());
 }
 
-conduit::Node load_actions(const std::string &file_name, int mpi_comm_id){
-  conduit::Node actions;
-
+void load_actions(conduit::Node &actions, const std::string &file_name, int mpi_comm_id){
```
cyrush: Unwritten rule: prefer out params last :-)
```suggestion
void load_actions(const std::string &file_name, int mpi_comm_id, conduit::Node &actions)
{
```

cyrush (approved):
Looks good!

PR #1367, opened 2024-08-22
Renaming replay_ser to ascent_replay and replay_mpi to ascent_replay_mpi

emily-howell:


emily-howell:
@cyrush Let me know if you want to do these name changes on their own or along with the addition of tests.

cyrush:
@emily-howell Looks good -- can you also add an entry to the changelog that mentions the name change?

Post changelog update, lets get this merged & we can think about replay tests after this PR.

Thanks!

PR #1383, opened 2024-09-09
fix for replay mpi case (fix defs)

cyrush:


cyrush:
Applies  replay fix tested in frontier branch.

Frontier branch still needs work, but want to get this in asap.

PR #1397, opened 2024-10-10
Fixing Nan errors in Camera Frustum Mesh

### LINKED ISSUE #1390: investigate nans in camera frustrum mesh
@emily-howell  can you take a look at this case for the frustum mesh and see why nans are arriving?

```
    camera:
      position: [0.0342999994754791, 0.00480000022798777, 0.0581200011074543]
      look_at: [0.0342999994754791, 0.00480000022798777, 0.0]
      up: [0.0, 1.0, 0.0]
      zoom: 2.0
      fov: 60.0
      near_plane: 0.0363688580691814
      far_plane: 3.63688564300537
      camera_frustum_mesh:
        coordsets:
          scene_bounds_coords:
            type: "explicit"
            values:
              x: [-0.0584625, -0.0584625, -0.0584625, -0.0584625, 0.2376, 0.2376, 0.2376, 0.2376]
              y: [-0.131195885166494, -0.131195885166494, 0.0797536816207084, 0.0797536816207084, -0.131195885166494, -0.131195885166494, 0.0797536816207084, 0.0797536816207084]
              z: [0.0, 0.0108, 0.0108, 0.0, 0.0, 0.0108, 0.0108, 0.0]
          camera_frustum_coords:
            type: "explicit"
            values:
              x: ["-nan", "nan", "nan", "-nan", "-nan", "nan", "-nan", "-nan", 0.0342999994754791, 0.0342999994754791, "-nan"]
              y: ["-nan", "-nan", "-nan", "-nan", "-nan", "-nan", "-nan", "-nan", 0.00480000022798777, 0.00480000022798777, "-nan"]
              z: ["-nan", "-nan", "-nan", "-nan", "-nan", "-nan", "nan", "-nan", 0.0217511430382729, -3.57876564189792, "-nan"]
        topologies:
          scene_bounds_topo:
            type: "unstructured"
            coordset: "scene_bounds_coords"
            elements:
              shape: "line"
              connectivity: [0, 1, 1, 2, 2, 3, 3, 0, 4, 5, 5, 6, 6, 7, 7, 4, 0, 4, 1, 5, 2, 6, 3, 7]
          camera_frustum_topo:
            type: "unstructured"
            coordset: "camera_frustum_coords"
            elements:
              shape: "line"
              connectivity: [0, 1, 1, 2, 2, 3, 3, 0, 4, 5, 5, 6, 6, 7, 7, 4, 0, 4, 1, 5, 2, 6, 3, 7, 8, 9, 8, 10]
          clipping_planes_topo:
            type: "unstructured"
            coordset: "camera_frustum_coords"
            elements:
              shape: "quad"
              connectivity: [0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 5, 4, 2, 3, 7, 6, 0, 3, 7, 4, 1, 2, 6, 5]
```

emily-howell:
Issue was that when normalized, the look vector had a slight floating point imprecision that caused this to try and calculate a new up vector. Because the up vector was being rotated based on that floating point imprecision, it was trying to rotate around a (0,0,0) axis which caused nan values


cyrush:
@emily-howell  can we add a test that exercises one of these cases?

Worried about the divide + mult impact on other cases.

Is it a case where the look at and position are very clost?

Is it possible to fix this before the normalize?

emily-howell:
For sure. Let me put together some test cases for this

emily-howell:
@cyrush To answer your above questions:

1) Q: can we add a test that exercises one of these cases?
    A: I just pushed up changes with a test added

2) Q: Worried about the divide + mult impact on other cases.
    A: I don't think it would have negatively impacted other cases. Since it was after the normalization step, all of the values being divided and multiplied are [0,1] which helps avoid potential overflow. That being said, I modified the handling of this case to avoid any mult/div/rounding.

3) Q: Is it a case where the look at and position are very clost?
    A: Not completely, it is a case where the vtkm Normalization of the look vector has some additional floating point values. E.g. instead of normalizing to [0.0, 0.0, 1.0] it is normalizing to [0.0, 0.0, 1.000000000003]. Which is a problem because when  you try to find the angle between the forward vector [0.0, 0.0, -1.0] and that look vector you get a non-zero angle. But because it is such a small non zero angle  the 3D vector transform of the up vector is trying to transform by essentially 0 degrees which causes nans.

4) Q: Is it possible to fix this before the normalize?
    A: No because it is a byproduct of that normalization step being imperfect. But there are plenty of ways of handling that imperfection and I pushed up an alternative approach just now that you can look at.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rendering_filters.cpp
```diff
@@ -1922,8 +1922,8 @@ void generate_camera_meshes(conduit::Node &image_data){
   vtkm::Vec<vtkm::Float64,3> forward(0,0,-1);
   double angle_between = vtkm::ACos(vtkm::Dot(forward, vtkm_look)) / vtkm::Pi() * 180;
 
-  // If the look vector has been rotated by a certain angle, ajust the camera up vector to match
-  if (angle_between != 0.0) {
+  // If the look vector has been rotated by a certain angle, adjust the camera up vector to match
+  if (vtkm::Abs(angle_between) >= 0.001) {
```
cyrush, line 1926: Awesome, this looks like a robust approach.

cyrush (approved):
Looks great, thanks!

cyrush:
Thanks @emily-howell !

PR #1398, opened 2024-10-10
Aurora scripts

mvictoras:
Added SYCL and Aurora compile scripts.


cyrush (approved):
Thanks!

scripts/build_ascent/build_ascent.sh
```diff
@@ -81,6 +82,14 @@ if [[ "$enable_hip" == "ON" ]]; then
     ROCM_PATH="${ROCM_PATH:=/opt/rocm/}"
 
     # NOTE: this script only builds kokkos when enable_hip=ON
+    build_kokkos="${build_kokkos:=true}"
+elif [[ "$enable_sycl" == "ON" ]]; then
+    echo "*** configuring with SYCL support"
+
+    CC=`which icx`
+    CXX=`which icpx`
+    FTN=`which ifx`
```
cyrush, line 91: Can you pass these in the aurora script instead of here b/c we won't be able to override with this way of finding the compilers.

cyrush: I am going to merge this, and then move these` which`s to the auroa specific script.

cyrush (approved):

scripts/build_ascent/build_ascent.sh
mvictoras: I apologize, I completely missed this! Thanks for fixing it

PR #1402, opened 2024-10-17
Fix for saving metadata bug on cscs alps

### LINKED ISSUE #1389: Saving metadata bug on CSCS Alps
Hi,

I have come across a bug when running Ascent on CSCS Alps, originally when running [PyFR](https://www.pyfr.org/) but I can recreate the issue by adding 2 lines to the "[First Light](https://ascent.readthedocs.io/en/latest/Tutorial_Intro_First_Light.html#render-a-sample-dataset-using-ascent)" example in the introduction to Ascent documentation:

```
#include <iostream>

#include "ascent.hpp"
#include "conduit_blueprint.hpp"

using namespace ascent;
using namespace conduit;

int main(int argc, char **argv)
{
    // echo info about how ascent was configured
    std::cout << ascent::about() << std::endl;

    // create conduit node with an example mesh using 
    // conduit blueprint's braid function
    // ref: https://llnl-conduit.readthedocs.io/en/latest/blueprint_mesh.html#braid

    // things to explore:
    //  changing the mesh resolution

    Node mesh;
    conduit::blueprint::mesh::examples::braid("hexs",
                                              50,
                                              50,
                                              50,
                                              mesh);
    mesh["/state/test/keyword"] = "TEST";
    mesh["/state/test/data"] = 1.0;
    // create an Ascent instance
    Ascent a;

    // open ascent
    a.open();

    // publish mesh data to ascent
    a.publish(mesh);

    //
    // Ascent's interface accepts "actions"
    // that to tell Ascent what to execute
    //
    Node actions;
    Node &add_act = actions.append();
    add_act["action"] = "add_scenes";

    // Create an action that tells Ascent to:
    // add a scene (s1) with one plot (p1)
    // that will render a pseudocolor of 
    // the mesh field `braid`
    Node & scenes = add_act["scenes"];

    // things to explore:
    //  changing plot type (mesh)
    //  changing field name (for this dataset: radial)
    scenes["s1/plots/p1/type"] = "pseudocolor";
    scenes["s1/plots/p1/field"] = "braid";
    // set the output file name (ascent will add ".png")
    scenes["s1/image_name"] = "out_first_light_render_3d";
    
    // view our full actions tree
    std::cout << actions.to_yaml() << std::endl;

    // execute the actions
    a.execute(actions);

    // close ascent
    a.close();
}
```

This works and produces a PNG with TEST = 1.0 in the metadata on my laptop but on Alps I get the following error:

```
[Error] Ascent::execute 
file: /capstor/scratch/cscs/[USERNAME]/code/manual-ascent-build/conduit-v0.9.2/src/libs/conduit/conduit_node.cpp
line: 15455
message: 
Node::as_string() const -- DataType empty at path [1] does not equal expected DataType char8_str
```

Using gdb I traced it back to here, specifically when getting the node as a string: https://github.com/Alpine-DAV/ascent/blob/c11da70c958bd9cdcc5d49eacbbc13c95a48ecef/src/libs/ascent/runtimes/flow_filters/ascent_runtime_rendering_filters.cpp#L348-L358

When the 'data' is already a string, i.e. `mesh["/state/test/data"] = "1.0";`, there is no issue on Alps so it appears to be an issue when casting the double to a string (but weirdly only on Alps).

On both Alps and my laptop I am using GCC 13.2.0. I have tried building Ascent using the `build_ascent.sh` script and also by building it (along with the dependencies) manually - both produce the same error.

A workaround for PyFR is to cast the metadata to a string before passing it to Ascent but it would be great to fully understand the issue here. If there is any further information I can provide or if there are any suggestions on how to investigate this further please let me know.

Thanks,
Toby

emily-howell:


cyrush (approved):
thanks for the change and for testing!

PR #1403, opened 2024-10-18
add ascent transform filter

cyrush:
adds a transform filter that supports:

- translate
- scale
- rotate
- transform via an arbitrary 4x4 transformation matrix


src/libs/vtkh/filters/PointTransform.cpp
```diff
@@ -5,61 +5,111 @@
...
   m_transform = vtkm::MatrixMultiply(m_transform, matrix);
 }
 
-void PointTransform::SetTransform(const vtkm::Matrix<double, 4, 4>& mtx)
+//---------------------------------------------------------------------------//
+void
+PointTransform::SetTransform(const double *matrix_values)
+{
+  // TODO: row vs col vs matrix vs array, what is the best
+  //       order for users to provide flat values to matrix?
```
cyrush, line 56: Note: this order seems good.

I googled some example transform matrices and threw them in easily.

JustinPrivitera: scale then rotate then translate

_T * R * S_

I didn't look at the code but is that what this is about? https://gamedev.stackexchange.com/questions/16719/what-is-the-correct-order-to-multiply-scale-rotation-and-translation-matrices-f

nicolemarsaglia (changes requested):
I like the unit test images! But it looks like you forgot to add the actual unit tests (:

src/libs/vtkh/filters/PointTransform.cpp
```diff
@@ -5,61 +5,111 @@
...
+
+  m_transform[2][0] = matrix_values[8];
+  m_transform[2][1] = matrix_values[9];
+  m_transform[2][2] = matrix_values[10];
+  m_transform[2][3] = matrix_values[11];
+
+  m_transform[3][0] = matrix_values[12];
+  m_transform[3][1] = matrix_values[13];
+  m_transform[3][2] = matrix_values[14];
+  m_transform[3][3] = matrix_values[16];
```
nicolemarsaglia: m_transform[3][3] = matrix_values[**15**];

PR #1409, opened 2024-10-25
Empty pipeline error when no filter

### LINKED ISSUE #1406: error not forwarded as a exception
The following actions are malformed:

```
- 
  action: "add_pipelines"
  pipelines: 
    pl1: 
      f1:
- 
  action: "add_scenes"
  scenes: 
    s1: 
      plots: 
        p1: 
          type: "pseudocolor"
          field: "pl"
          pipeline: "pl1"
- 
  action: "save_info"

```

However the error will not trigger an exception when we try to forward:

```
Node opts;
opts["exceptions"]="forward";
ascent.open(opts)
```

This will:

Sounds like we need to check for empy pipeline

```
- 
  action: "add_pipelines"
  pipelines: 
    pl1: 
      f1:
         type: "garbage"
- 
  action: "add_scenes"
  scenes: 
    s1: 
      plots: 
        p1: 
          type: "pseudocolor"
          field: "pl"
          pipeline: "pl1"
- 
  action: "save_info"

```

emily-howell:


cyrush (approved):
Thanks for fixing + and for the test!

PR #1412, opened 2024-10-28
Update Partition filter & docs

nicolemarsaglia:
Updated docs -- feedback welcome

Add more mpi partition unit tests


src/libs/ascent/runtimes/flow_filters/ascent_runtime_blueprint_filters.cpp
```diff
@@ -359,29 +388,70 @@ BlueprintPartition::execute()
...
     }
 #else
     conduit::blueprint::mesh::partition(*n_input,
                                         n_options,
-                                        *n_output);
+                                        tmp);
 #endif
+    if(target == 1)
+    {
+      n_output->append() = tmp;
```
cyrush:
```suggestion
      n_output->append().move(tmp);
```

src/libs/ascent/runtimes/flow_filters/ascent_runtime_blueprint_filters.cpp
```diff
@@ -359,29 +388,70 @@ BlueprintPartition::execute()
 
     conduit::Node *n_output = new conduit::Node();
     
-    conduit::Node n_options = params();
+    conduit::Node n_options;
```
cyrush, line 394: we have already validated the input in verify params, we can simply pass the options along as is -- we don't need to pick out the individual params and copy them.

src/docs/sphinx/Actions/Pipelines.rst
```diff
@@ -1089,33 +1089,26 @@ among many simulation ranks. Ascent utilizes the ``partition()`` functions provi
...
-appropriate to contain the inputs.
-
-In parallel, the ``partition()`` function will make an effort to redistribute data across MPI
+  conduit::Node pipelines;
+  // pipeline 1
+  pipelines["pl1/f1/type"] = "partition";
+  //params optional
+  pipelines["pl1/f1/params/target"] = 1;
+  const char* fields[3] = {"pink", "pony", "club"};
+  pipelines["pl1/f1/params/fields"].set(fields,3);
```
cyrush: I think this will only end up wrapping the first 3 letters of the char *
For a list of strings, they need to be a conduit list, where each entry is a string.

cyrush:
This was failing on windows b/c of the embedded path issue with conduit vs windows:
```
IOHandle does not support opening paths with subpaths specified: "D:\a\ascent\ascent\build\tests\_output/D:\a\ascent\ascent\build\tests\_output\tout_partition_multi_dom_serial.cycle_000000.root"
```

cyrush:
@nicolemarsaglia  merging this to get important fixes into develop.

We still need more tests to demonstrate and validate the partition use cases, we can tackle those in the future.

PR #1434, opened 2024-12-19
Fixed ascent clone path

mvictoras:
The path is incorrect if script is standalone.
This PR fixes the path in the case that the script needs to clone ascent.


cyrush (approved):

cyrush:
thanks @mvictoras !

PR #1442, opened 2025-01-15
Adding actions for opening, flushing, and closing the logs

### LINKED ISSUE #1431: add actions that control logging options
Actions that control logging to give users more control than just params passed via `ascent::open`

Examples:

start logging at any time:

```
action: open_log # open ascent's logging
file_pattern: (support all params in new logging options added in #1426)
```

flush logs so they can be seen before ascent is closed out:

```
action: flush_log # flush current log streams to disk
```

stop logging and flush current logs:

```
action: close_log # close current log streams
```

emily-howell:
This currently only adds some simple actions for opening, flushing, and closing the logs. I wonder if it would be useful to also add an action for modifying the logging threshold?


cyrush:
this looks great!

Can you add a note to the CHANGELOG.md file about these actions and a short section in the docs?

We can think about the mpi relevant options `rank`, etc in the future. 

We could also add support for `log_threshold` to `open_log` so folks can use one action to setup everything up.

The separate actions for log and echo thresholds are a good addition.

emily-howell:
Yes, I can add a note to the change log.

I have a draft of the MPI relevant options but I don't have tests written for it yet so I didn't push them.

I like the idea of adding the log threshold to the options for open_log. That should be a quick change so I will add that too.

emily-howell:
@cyrush I circled back to this and updated the MPI thresholding to be Debug for rank 0 and Warnings for all other ranks. Does that work?

cyrush (approved):

cyrush:
@emily-howell  sounds good, thanks!

cyrush:
Merging, thanks again for this!

PR #1446, opened 2025-01-28
Add conversion for VisIt color table to Ascent actions color table

nicolemarsaglia:
Python script that converts a .ct file to ascent actions color table


cyrush:
can we change `session_to_color_table.py` to `visit_color_table_to_ascent_color_table.py`?

cyrush (approved):
Thanks!

Issue #1438, opened 2024-12-20
2d camera controls


cyrush:
make setting views easier for 2d cases

cyrush:
`dataset_bounds` provides this already for normal renders.

We also want this for `project2d` which we used for scalar rendering.

PR #1454, opened 2025-02-04
scalar renderer (project 2d) add option to select fields

cyrush:


cyrush:
also working on #1438

PR #1461, opened 2025-02-11
uniform grid fixes

nicolemarsaglia:
Rework Uniform Grid Sampling to utilize built-in masks from vtkmProbe and change way to break ties across domains with valid points: locally by highest domain ID and globally by highest rank.


src/libs/vtkh/filters/UniformGrid.cpp
```diff
@@ -283,17 +294,18 @@ class GlobalReduceField
         {
           for(int i = 0; i < num_points; ++i)
           {
-            if(g_mask[i] == 1)
+            if(global_field[i] == 0) std::cerr << "POINT IS ZERO: " << i  << std::endl;
```
cyrush: should we remove this debug bring?

src/libs/vtkh/filters/UniformGrid.cpp
```diff
@@ -323,13 +336,13 @@ class GlobalReduceField
...
-            }
-            if(g_valid[i] > 1)
+            if(g_rank_mask[i] == -1)
             {
-                    global_field[i] = global_field[i]/g_valid[i];
+              global_field[i] = m_invalid_value;
+#if _DEBUG 
+              global_field[i] = par_rank*10;
+#endif
+	      std::cerr << "INVALID VALUE>>>>>>>>>>>>>>>>>>>>>>>>>" << std::endl;
```
cyrush: same, do we need this cerr?

nicolemarsaglia: thank you! Thought I caught all of these. We should only have cerr within the _DEBUGs. I'll do another look through as well.

cyrush (approved):
Nice work, I checked with my test data with 4 tasks and all looks good now!

PR #1463, opened 2025-02-20
Fix issues with umpire in ascent_config.mk.in and dray_config.mk.in.

brugger1:
Umpire stores its libraries in a `lib64` directory and the `*config.mk.in` files had `lib`. I changed the paths in the `*config.mk.in` files.

There was a typo in `ascent_config.mk.in` where `ASCENT_UMPIRE_LIB_FLAGS` was used instead of `ASCENT_CAMP_LIB_FLAGS`. I assume this was from a cut, paste and modify error.

There was a typo in `dry_config.mk.in` where `DRAY_RAJA_DIR` was used instead of `DRAY_UMPIRE_DIR`. I assume this was from a cut, paste and modify error.


cyrush:
thank you, I fixed this for the cmake case, but missed the config.mk case!

https://github.com/Alpine-DAV/ascent/blob/86d8b77700ab4785c09efb76c6df1c1804559fdb/src/config/ascent_setup_deps.cmake#L258

cyrush (approved):

brugger1:
@cyrush , the CI now passes. I tested with both `lib` and `lib64` for umpire and both worked. It looks like even though I changed the code after your review, your review is still sufficient for me to checkin. Do you want to take a quick look?

cyrush:
@brugger1 feel free to merge, thanks for this fix!

PR #1472, opened 2025-03-07
add ascent lib version details to info

### LINKED ISSUE #1451: add ascent version and git sha to info()
so we can use `save_info` to get a reliable read on version of ascent compiled into any code

cyrush:
resolves #1451


cyrush:
I verified that info has the new entries.

cyrush:
added explicit test to check for the entries

PR #1480, opened 2025-03-17
relay blueprint -- add generated root file as path in extract info

### LINKED ISSUE #1465: info is missing extract file path for relay extract
```
- 
  action: "add_extracts"
  extracts: 
    e1: 
      type: "relay"

      params: 
        path: "out_export_braid_all_fields"
        protocol: "blueprint/mesh/hdf5"
```

Resulting info:

```
- 
  type: "relay"
  protocol: "blueprint/mesh/hdf5"
  path: ""
```

cyrush:
wip to resolve #1465

We are going to address this with a new helper in conduit that creates the root file name, given the mesh, path, and options.


src/libs/ascent/runtimes/flow_filters/ascent_runtime_relay_filters.cpp
```diff
@@ -988,6 +998,7 @@ RelayIOSave::execute()
     }
 #endif
 
+    std::string mapped_protocol = protocol;
```
cyrush: this needs to be empty

src/libs/ascent/runtimes/flow_filters/ascent_runtime_relay_filters.cpp
```diff
@@ -578,13 +586,15 @@ verify_io_params(const conduit::Node &params,
...
-                         int num_files,
-                         const Node &extra_opts,
-                         std::string &root_file_out)
+void
+mesh_blueprint_save(const Node &data,
+                    const std::string &path,
+                    const std::string &file_protocol,
+                    int num_files,
+                    const Node &extra_opts,
+                    std::string &root_file_out)
```
JustinPrivitera, line 595: I guess `root_file_out` wasn't used before, but was just a hook here for when we added the generate filename helper?

Now that it is getting set and used somewhere else in the case where we can generate the filename, do you handle the case where we cannot generate the filename? Does it still get used somewhere else? Does that make sense?

src/tests/ascent/t_ascent_relay.cpp
```diff
@@ -82,15 +89,22 @@ TEST(ascent_relay, test_relay_hdf5)
...
+    ascent.open();
     ascent.publish(data);
     ascent.execute(actions);
+    ascent.info(info);
     ascent.close();
 
+    std::cout << info.to_yaml();
+
+#ifdef CONDUIT_HAS_ROOT_FILE_NAME_GEN
+    // if conduit 0.9.4 provides util that allows us to
```
JustinPrivitera:
```suggestion
    // conduit 0.9.4 provides util that allows us to 
```

src/libs/ascent/runtimes/flow_filters/ascent_runtime_relay_filters.cpp
JustinPrivitera: I guess I don't see where `root_file_out` is plumbed to, but you should make sure that wherever it goes it is robust to the case where it does not exist.

JustinPrivitera (approved):

src/libs/ascent/runtimes/flow_filters/ascent_runtime_relay_filters.cpp
cyrush: It was being used before but was always blank. The upshot was that the extracts path entry in the `ascent.info` output was always an empty string when saving meshes.

JustinPrivitera: do we want that behavior to remain? It might be more useful to add a message there saying that it couldn't generate the output filename b/c reasons. Or maybe it should be omitted in that case?

cyrush: I think an empty path is best for now, it's easier to check for that than to parse an error string.

JustinPrivitera: sounds good

PR #1482, opened 2025-03-18
Include Yaml

### LINKED ISSUE #1435: a few actions improvements ideas
color tables entry: allow, r,g,b,a, position arrays for a more compact way to specify a color table
explore an `include: myfile.yaml` directive that can pull in extra settings

emily-howell:


src/libs/ascent/ascent.cpp
```diff
@@ -77,6 +77,91 @@ check_for_file(const std::string &file_name,
...
+            catch(conduit::Error &e)
+            {
+                include_file_valid = 0;
+                emsg = e.message();
+            }
+        }
+
+#ifdef ASCENT_MPI_ENABLED
+        // make sure all ranks error if the parsing on rank 0 failed.
+        MPI_Bcast(&include_file_valid, 1, MPI_INT, 0, mpi_comm);
```
cyrush, line 141: If we have an error, we should broadcast the error message from rank 0 to all other ranks so their exceptions match

src/libs/ascent/ascent.cpp
```diff
@@ -242,11 +327,16 @@ CheckForSettingsFile(std::string file_name,
         ASCENT_ERROR("Failed to load actions file: " << file_name
                      << "\n" << emsg);
     }
+
 #ifdef ASCENT_MPI_ENABLED
     relay::mpi::broadcast_using_schema(node, 0, mpi_comm);
 #endif
+
+    detail::load_included_files_in_node_tree(node, mpi_comm_id);
+    node.print();
```
cyrush: Comment out the node.print()

cyrush (commented):
Looks great, to small suggestions to wrap up the PR.

emily-howell:
@cyrush Thank you for the feedback. I addressed your comments

src/libs/ascent/ascent.cpp
```diff
@@ -77,6 +77,98 @@ check_for_file(const std::string &file_name,
...
+#ifdef ASCENT_MPI_ENABLED
+        // make sure all ranks error if the parsing on rank 0 failed.
+        MPI_Bcast(&include_file_valid, 1, MPI_INT, 0, mpi_comm);
+
+        // Pass the error to all ranks so the error message matches
+        int line_size = emsg.size();
+        MPI_Bcast(&line_size, 1, MPI_INT, 0, mpi_comm);
+        if (rank != 0)
+            emsg.resize(line_size);
+        MPI_Bcast(const_cast<char*>(emsg.data()), line_size, MPI_CHAR, 0, mpi_comm);
```
cyrush: it's scary to write into an std::string as a buffer, how about:

```
Node n_emsg;
if(rank == 0)
{
   n_emsg.set(emsg);
}

relay::mpi::broadcast_using_schema(n_emsg, 0, mpi_comm);
emsg = n_emsg.as_string();
```

emily-howell: That is a much better way of doing that. Thank you

cyrush (approved):
Looks good, thanks!

PR #1487, opened 2025-03-19
Task/2025 3 18 1413 camera zoom of 0 causes hang

### LINKED ISSUE #1413: camera zoom of 0 causes hang
Accidentally had camera zoom set to 0  (which doesn't make much sense) and noticed that my render hung. 
We should provide an error, not hang.

emily-howell:
Added an Ascent Error to let the user know they should fix the zoom


src/tests/ascent/t_ascent_render_3d.cpp
```diff
@@ -3275,6 +3275,88 @@ TEST(ascent_render_3d, test_render_3d_bentgrid_example)
...
+        ascent.execute(actions);
+    }
+    catch(...)
+    {
+        error = true;
+    }
+
+    ascent.close();
+
+    EXPECT_TRUE(error);
```
cyrush, line 3352: 👍

cyrush (approved):

PR #1496, opened 2025-03-28
Task/2025 3 27 1428 path enhancements rollup issue fmt path strings

### LINKED ISSUE #1428: path enhancements rollup issue
This issue links to several file path and naming issues.

They may need simple one off fixes and reviewing could inform need for a holistic approach to dealing with paths:

#1400
#1289
#1285 
#1368

emily-howell:


src/libs/ascent/utils/ascent_string_utils.cpp
```diff
@@ -39,39 +50,133 @@ void split_string(const std::string &s,
...
+
+template<typename T>
+std::string expand_generic_variable(const std::string& path_string, 
+                                    const std::regex& pattern, 
+                                    const T value)
+{
+    std::smatch match;
+    std::string result_string = path_string;
+
+    std::regex valid_int_format(R"(^\d*d$)");
```
cyrush: these look good to me, as a gift to our future selves, can you add a comment  about the types of patterns these will catch?

cyrush (commented):
This is great!


Can we add some tests for path gen strings in:

https://github.com/Alpine-DAV/ascent/blob/develop/src/tests/ascent/t_ascent_utils.cpp

There we can try for various patterns without a full ascent pipeline execution. 
We can also test what happens with mangled pattern (which folks will give us)

A new place where we need path expansion in the legacy vtk extract:

https://github.com/Alpine-DAV/ascent/blob/0f24f6eddb25fade184f475634292f72a2d9ae31/src/libs/ascent/runtimes/flow_filters/ascent_runtime_vtkh_filters.cpp#L5995

For the family case, we should think about how to handle if the file exists on disk.
That is the case we want to check only with rank 0.

I think `s_file_family_map` will be the same on all tasks for our use cases.

emily-howell:
@cyrush I think I have addressed all your feedback but this is still a WIP:
- [x] Added a bunch of tests to t_ascent_utils.cpp
- [x] Added path expansion to legacy vtk extract
- [x] Checking if the file pattern exists already in the directory and determine the family value from that
- [x] Added some comments about what the regular expressions do

Outstanding:
- [ ] Some tests are failing still so I will have to hunt down what is happening in each of them

src/libs/ascent/utils/ascent_string_utils.cpp
```diff
@@ -39,39 +50,243 @@ void split_string(const std::string &s,
...
+  }
+
+  // Use the defined pattern to make the final regular expression
+  // Adding a ^ to lock the pattern to the start of the file_name and a pattern for a file extension
+  // at the end to the pattern
+  search_pattern_str = "^" + search_pattern_str + R"(\.[a-zA-Z]+$)";
+  std::regex search_pattern(search_pattern_str);
+
+  // Checking the directory contents for any filenames that match
+  std::vector<std::string> contents;
```
cyrush: we only want to check the dir contents on rank 0

src/libs/ascent/utils/ascent_string_utils.cpp
```diff
@@ -39,39 +50,243 @@ void split_string(const std::string &s,
...
+}
+
+int get_family_value(const std::string& path_string, int family_value)
+{
+  
+  std::string modified_path_string = path_string;
+
+  int rank = 0;
+#ifdef ASCENT_MPI_ENABLED
+  MPI_Comm mpi_comm = MPI_Comm_f2c(flow::Workspace::default_mpi_comm());
```
cyrush: lets pass in `int mpi_comm_id`, instead of using comm form the workspace, that way we can use it generally

src/libs/ascent/utils/ascent_string_utils.cpp
```diff
@@ -39,39 +50,243 @@ void split_string(const std::string &s,
...
+                        << format_spec 
+                        << "'. Inserting value without formatting.");
+          result_string.replace(match.position(0), match.length(0), std::to_string(value));
+        }
     }
+
+    return result_string;
+}
+
+int check_directory_for_family_value(const std::string& path_string, int family_value) {
```
cyrush: we should pass `int mpi_comm_id` id here. 

We can use a strategy like this to handle mpi vs non mpi:

https://github.com/Alpine-DAV/ascent/blob/7cb169c0ddf1f5217458cde2d6179ad9a67e9223/src/libs/ascent/utils/ascent_actions_utils.cpp#L368

src/libs/ascent/utils/ascent_string_utils.cpp
```diff
@@ -39,39 +50,243 @@ void split_string(const std::string &s,
...
+    family_value = s_file_family_map[path_string] + 1;
+    s_file_family_map[path_string] = family_value;
   }
-  return result;
+
+  return family_value;
+}
+
+std::string expand_path_special_variables(const std::string &path_string,
+                                          int counter,
```
cyrush, line 262: we may have to pass mpi comm here as well

cyrush (commented):
Looks good thanks for all the new tests!
Some suggestions about how to support the mpi case.

src/libs/ascent/utils/ascent_string_utils.cpp
emily-howell: I opted to protect that entire chunk where contents is used for just rank 0 and then afterwords I broadcast the updated family value. Alternatively, I could set the contents variable and then pass that into a conduit node and broadcast that node. Let me know if you want me to make that change :)

cyrush (approved):
Looks good, thanks for the adjustments.

emily-howell:
> @cyrush I think I have addressed all your feedback but this is still a WIP:
> 
>     * [x]  Added a bunch of tests to t_ascent_utils.cpp
> 
>     * [x]  Added path expansion to legacy vtk extract
> 
>     * [x]  Checking if the file pattern exists already in the directory and determine the family value from that
> 
>     * [x]  Added some comments about what the regular expressions do
> 
> 
> Outstanding:
> 
>     * [ ]  Some tests are failing still so I will have to hunt down what is happening in each of them

With respect to the outstanding failing tests:
- [x] ascent_runtime_options.test_timings_tear_updown is failing because when execute is called it is rendering twice even though it is only supposed to render once. This wasn't a problem before but now that the family variable in the output path is checking the file directory to prevent overwriting existing files it is now producing both renders with different names. 
- [x] ascent_runtime_options.test_actions_file appears to only be failing because ascent_runtime_options.test_timings_tear_updown is failing. They use the same output file names and the extra file left over from ascent_runtime_options.test_timings_tear_updown is causing ascent_runtime_options.test_actions_file to increment the family value it uses for it's output file.
- [x] ascent_pipeline_failure.test_bad_chaining and ascent_pipeline_failure.test_bad_chaining_topo both failing. Cause of failure unclear but they also use the same output file names (which is also used in ascent_pipeline_failure.test_bad_field) so potentially if tests are run in parallel that could cause the family name to increment. Issue is only seen when running all tests together. Not when tests are run individually.

emily-howell:
@cyrush The final remaining failing test is failing because vtk extracts is writing files on multiple ranks, not just rank 0. So when it calls "check_directory_for_family_value" on rank 1 it is not handling the family setting properly. Should I maybe add a flag to allow for multiple ranks to do file I/O if passed?

cyrush:
@emily-howell multiple ranks to need to write in parallel.  All ranks should get the same family answer -- it's just that rank 0 should be the only one that asks questions to the file system.

emily-howell:
Sorry, maybe I am miss understanding. I thought that the family value was a unique value that could be used to differentiate between files of the same name. Speficically in vtk that is the "domain_{family:08d}.vtk". Rank 0 in the "ascent_vtk_file_extract.basic_mulit_domain" test rank 0 is outputting `domain_00000000.vtk`, `domain_00000001.vtk`, `domain_00000002.vtk` while rank 1 is supposed to output `domain_00000003.vtk`, `domain_00000004.vtk`, `domain_00000005.vtk`, and `domain_00000006.vtk`.

While writing that I am wondering if I missunderstood your intention and instead should have left the `domain_{:08d}.vtk` statements as they were and instead just used the updated file path handling for the `std::string output_visit_file = output_base + ".visit";`?

cyrush:
In the case of the vtk files, we must write one file out for each domain.

They stomp on each other if we don't have a unique base name, so this is what we are aiming for:

```
myoutput_{family:08d}/domain_00000000.vtk
myoutput_{family:08d}/domain_00000001.vtk
myoutput_{family:08d}/domain_{n-1}.vtk
```

cyrush:
The individual ranks know the domain ids, it's the root we need to be unique.

emily-howell:
Okay, that makes more sense 😅

emily-howell:
@cyrush Looks like I resolved all the tests that were failing. I think this is good for another review

CHANGELOG.md
```diff
@@ -35,6 +35,9 @@ and this project aspires to adhere to [Semantic Versioning](https://semver.org/s
 - Added a `declare_fields` action, that allows users to explicitly list the fields to return for field filtering. This option avoids complex field parsing logic.
 - Added a 2d camera mode (`camera/2d: [left, right, bottom, top]`) to scene render cameras and the `project_2d` (scalar rendering) filter cameras.
 - Added support for `include` keyword to include children from yaml files in an input node trees
+- Added support for special keyword formatting for output paths. Current supported keywords include
+`cycle`, `family`, and `time`.
```
cyrush, line 39: `ohana`?

src/libs/ascent/runtimes/ascent_metadata.hpp
```diff
@@ -29,7 +31,7 @@ namespace ascent
 // PopoulateMetaData fills this inside the main runtime
 struct Metadata
```
cyrush: I think ASCENT_API should be attached to the overall struct.

src/libs/ascent/utils/ascent_string_utils.cpp
```diff
@@ -39,39 +50,269 @@ void split_string(const std::string &s,
...
+    //
+    // The different supported floating point number types are:
+    //    f: decimal floating point
+    //    F: decimal floating point
+    //    e: scientific notation using e to indicate the exponent
+    //    E: same as e except uses E to indicate the exponent
+    //    g: Uses the shortest notation (either e or f)
+    //    G: Uses the shortest notation (either E or F)
+    //
+    // e.g. 3.14f or 2F or 023g or 10.4e
```
cyrush, line 101: great comments to document this here, thank you for adding!

src/libs/ascent/utils/ascent_string_utils.cpp
```diff
@@ -39,39 +50,269 @@ void split_string(const std::string &s,
...
+                                     int mpi_comm_id,
+                                     int family_value)
+{
+  // Initialized the MPI variables if needed
+  int rank = 0;
+#ifdef ASCENT_MPI_ENABLED
+  if(mpi_comm_id == -1)
+  {
+      // report as failure to load
+      return false;
```
cyrush: return `0`? We can discuss

src/libs/ascent/utils/ascent_string_utils.cpp
```diff
@@ -39,39 +50,269 @@ void split_string(const std::string &s,
...
+  std::regex other_fmts_pattern(R"(\{[a-zA-Z]*:[a-zA-Z0-9.]*\})");
+
+  // This pattern is used to match to numbers.
+  // It is looking for integers, decimal, and scientific notation values
+  // Explanation:
+  //      [+-]?              - Optional + or - symbol
+  //      \d+                - At least one digit
+  //      (:?\.\d+)?         - Optional decimal value in a non-capturing group
+  //      (?:[eE][+-]?\d+)?  - Optional handling for scientific notation (e.g. e+02)
+  std::string number_pattern = R"([+-]?\d+(?:\.\d+)?(?:[eE][+-]?\d+)?)";
```
cyrush, line 176: again, awesome comments!

cyrush (approved):
Looks great!

Lets add a ticket to add follow up docs and to update our tutorials to all use this.

src/libs/ascent/utils/ascent_string_utils.cpp
emily-howell: Good catch, that would not be the desired behavior. I can change that to just return the family_value that was supplied originally. Or would it be better in this case to throw an ASCENT_ERROR? I have seen other places in the code that say "do nothing, an error will be thrown later so we can respect the exception handling" which was the mentality here too.

CHANGELOG.md
emily-howell: No Ohana was added. I can but I didn't want to muddy the code

PR #1497, opened 2025-04-01
Non existant directory check

### LINKED ISSUE #1400: ascent crash while trying to write image to non existent directory
We think that using:

```
image_prefix: "mydir/mypng"
```

Where `mydir` doesn't exist, lead to a crash.

We should error, not crash.

emily-howell:


src/libs/ascent/runtimes/flow_filters/ascent_runtime_rendering_filters.cpp
```diff
@@ -1052,8 +1052,8 @@ DefaultRender::verify_params(const conduit::Node &params,
                              conduit::Node &info)
 {
     info.reset();
-    bool res = check_string("image_name",params, info, false);
-    res &= check_string("image_prefix",params, info, false);
+    bool res = check_path("image_name",params, info, false);
+    res &= check_path("image_prefix",params, info, false);
```
cyrush: I think we should check this on execution, not on verify params.
We should also only check on rank 0 :-)

emily-howell:
@cyrush This is ready for review again :) 

It will conflict with my other PR since I touch `DefaultRender::execute` in both so I would appreciate this one going in first.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_utils.cpp
```diff
@@ -56,6 +62,39 @@ std::string output_dir(const std::string file_name)
...
+
+bool check_dir_path_exists(const std::string file_path, int mpi_comms_id, conduit::Node &err_msg)
+{
+  int rank = 0;
+#ifdef ASCENT_MPI_ENABLED
+  MPI_Comm mpi_comm = MPI_Comm_f2c(mpi_comms_id);
+  MPI_Comm_rank(mpi_comm, &rank);
+#endif
+
+  int res = true;
```
cyrush: minor suggestion:
Use 1/0 b/c the type is an int. 

Using true and false is more readable, but it makes me think we are using bools, which we don't want to do for MPI calls.

cyrush (approved):
Looks good, on minor suggestion which we can discuss if you think we should keep as is.

PR #1499, opened 2025-04-10
Fix uniform grid to account 2D slices along all axis

nicolemarsaglia:
Previously only took into account slices along the Z-axis, now we can do slices along any axis.


emily-howell:
Should a new test be added to exercise this change?

nicolemarsaglia:
@emily-howell ok ready for some eyes when you get a chance

src/tests/ascent/t_ascent_uniform_grid.cpp
```diff
@@ -33,6 +33,188 @@ using namespace ascent;
...
+    add_scenes["action"] = "add_scenes";
+    add_scenes["scenes"] = scenes;
+    //add the extracts
+//    conduit::Node &add_extracts= actions.append();
+//    add_extracts["action"] = "add_extracts";
+//    conduit::Node &extracts = add_extracts["extracts"];
+//    extracts["e1/type"] = "relay";
+//    extracts["e1/pipeline"] = "pl1";
+//    extracts["e1/params/path"] = output_file + "_blueprint";
+//    extracts["e1/params/protocol"] = "hdf5";
```
emily-howell, line 198: Would maybe clean this up

```suggestion
```

src/tests/ascent/t_ascent_uniform_grid.cpp
```diff
@@ -33,6 +33,188 @@ using namespace ascent;
...
```
emily-howell, line 107: Also here

```suggestion
```

emily-howell (approved):
Would be nice to clean up the commented out sections in the code but otherwise this looks good to me.

Probably can't merge until the CI is fixed

src/tests/ascent/t_ascent_uniform_grid.cpp
nicolemarsaglia: @emily-howell ya, it's handy when you need to debug, but could do with only one copy for reference. I'll clean it up at some point, probably the next time I need to fix uniform grid.

PR #1500, opened 2025-04-15
DRAFT: First pass at a clang formatter

emily-howell:
Can be played around with locally by calling:
```
find ./src/libs/ascent/ -regex '.*\.\(c\|cpp\|h\|hpp\)' -exec clang-format -i {} +
```

I don't know that I want all of the formatting to go in in one big MR like this. This is more a vessel to visualize the types of changes it will make


emily-howell:
@cyrush @nicolemarsaglia this is a first pass a what formatting could look like. My changes are the `.clang-format` file. Everything else is the result of running the formatter on Ascent code. The goal here is not to merge this in but rather show what changes would occur if this was added. 

**Please criticize and tear this apart!! The goal here is to spur discussion if it is needed.**

Eventually, this would need to be built into the CI and also be run on the tests and other files.

src/libs/ascent/hola/ascent_hola.hpp
```diff
@@ -4,22 +4,20 @@
...
 /// file: ascent_hola.hpp
 ///
 //-----------------------------------------------------------------------------
 
 #ifndef ASCENT_HOLA_HPP
-#define ASCENT_HOLA_HPP
-
-#include <ascent_config.h>
-#include <ascent_exports.h>
+#    define ASCENT_HOLA_HPP
```
cyrush: In general I like intended defs, however for the header file case there is always a indef / define that wraps the whole file to avoid multiple definitions.

It would be good to not indent the normal includes for this case.

src/libs/ascent/python/ascent_module/ascent_python_exports.h
```diff
@@ -18,28 +18,26 @@
...
-#else
-#define ASCENT_PYTHON_API __declspec(dllimport)
-#endif
-#if defined(_MSC_VER)
+#    if defined(ASCENT_PYTHON_EXPORTS) || defined(ascent_python_EXPORTS)
+#        define ASCENT_PYTHON_API __declspec(dllexport)
+#    else
+#        define ASCENT_PYTHON_API __declspec(dllimport)
+#    endif
+#    if defined(_MSC_VER)
```
cyrush: indents do improve readability here

src/libs/ascent/runtimes/ascent_data_object.hpp
```diff
@@ -4,91 +4,96 @@
...
+#    include <ascent.hpp>
+#    include <conduit.hpp>
+#    include <memory>
 
 //-----------------------------------------------------------------------------
 // -- begin ascent:: --
 //-----------------------------------------------------------------------------
 // forward declare
-#if defined(ASCENT_DRAY_ENABLED)
+#    if defined(ASCENT_DRAY_ENABLED)
```
cyrush: def indent nesting doesn't work well for this case

src/libs/ascent/runtimes/ascent_empty_runtime.cpp
```diff
@@ -86,32 +77,19 @@ EmptyRuntime::Initialize(const conduit::Node &options)
 }
 
 //-----------------------------------------------------------------------------
-void
-EmptyRuntime::Info(conduit::Node &out)
-{
-    out.set(m_info);
-}
+void EmptyRuntime::Info(conduit::Node &out) { out.set(m_info); }
```
cyrush: I prefer multi line even for these simple guys

src/libs/ascent/runtimes/ascent_empty_runtime.cpp
```diff
@@ -121,33 +99,27 @@ EmptyRuntime::Publish(const conduit::Node &data)
...
-    if(verify_ok)
+    if (verify_ok)
         n_src = (int)0;
     else
         n_src = (int)1;
 
-    conduit::relay::mpi::sum_all_reduce(n_src,
-                                        n_reduce,
-                                        mpi_comm);
+    conduit::relay::mpi::sum_all_reduce(n_src, n_reduce, mpi_comm);
```
cyrush, line 116: I also like multi line in these cases, not sure if it is possible to retain them.
For longer functions multi line allows you to add comments after each of the arguments.

.clang-format
```diff
@@ -0,0 +1,47 @@
...
+
+
+#############################
+### Alignment and Spacing ###
+#############################
+
+# Basic spacing and line length limits
+ColumnLimit: 79
+PenaltyExcessCharacter: 1000000 # Allows for strings to overflow past column
+AlignArrayOfStructures: Left    # Alignes columns of structures so they all line up clearly
```
JustinPrivitera:
```suggestion
AlignArrayOfStructures: Left    # Aligns columns of structures so they all line up clearly
```

.clang-format
```diff
@@ -0,0 +1,47 @@
...
+AlignConsecutiveAssignments: false # Don't bother lining up `=` symbols in consecutive lines
+AlignConsecutiveDeclarations: false # Don't bother lining up `=` symbols in consecutive lines
```
JustinPrivitera, line 15: does this enforce this
```
int a = 1;
float b = 2.0;
double c = 3.0;
```
versus this?
```
int a    = 1;
float b  = 2.0;
double c = 3.0;
```
or does it just relax the restriction so users can do whatever?

Also, as far as I know, `AlignConsecutiveDeclarations` is not for lining up `=` symbols, it is aligning declarations like this:
```
int    a;
float  b;
double c;
```

.clang-format
```diff
@@ -0,0 +1,47 @@
...
+
+# For lines that get wrapped, this forces one parameter per line and aligns them
+BinPackParameters: false        # One param per line if wrapped
+BinPackArguments: false         # One argument per line if wrapped
+AlignAfterOpenBracket: Align    # Align wrapped args/params under the first
+AllowAllParametersOfDeclarationOnNextLine: false
```
JustinPrivitera, line 21:
```suggestion
AllowAllParametersOfDeclarationOnNextLine: true
```
I think this should be true. Sometimes namespace::method() names get so long that it is useful to put parameter declarations on the following lines.

.clang-format
```diff
@@ -0,0 +1,47 @@
...
+
+# Indentation
+IndentWidth: 4
+UseTab: Never
+IndentPPDirectives: AfterHash   # don't indent things like `#ifdef` and `#define` unless nested
```
JustinPrivitera:
```suggestion
IndentPPDirectives: Always   # don't indent things like `#ifdef` and `#define` unless nested
```

I prefer `Always` if we have to do this at all. With extreme nesting this can potentially be burdensome.

.clang-format
```diff
@@ -0,0 +1,47 @@
...
+### Includes ###
+################
+SortIncludes: false             # Don't auto-sort #includes
+IncludeBlocks: Preserve         # Keeps include groups separate
+
+
+##################
+### Namespaces ###
+##################
+AccessModifierOffset: -4        # Access modifiers (public:, private:, protected:) should be inset
```
JustinPrivitera, line 46:
```suggestion
AccessModifierOffset: 0        # Access modifiers (public:, private:, protected:) should be inset
```
🙃

src/libs/ascent/runtimes/ascent_data_object.hpp
JustinPrivitera: The setting Emily has now indents after the hash. There is another setting to indent the hash too. Is that what you didn't like, or you just didn't want it indented at all?

cyrush: All our header files look like:

```
#ifndef HDR_FILE_NAME
#define HDR_FILE_NAME
...
#include <myheader1>
#include <myheader2>
#include <myheader3>
#include <myheader4>

...
#endif
```

The current setup transforms them all to:

```
#ifndef HDR_FILE_NAME
#    define HDR_FILE_NAME

#    include  <myheader1>
#    include  <myheader2>
#    include  <myheader3>

...
#endif
```

Not a fan of this, I don't think it is an improvement.

cyrush: in this case `HDR_FILE_NAME` is to avoid duplicate defs, but its not otherwise important to the structure or semantics of the header file

JustinPrivitera: I see

.clang-format
emily-howell: `AlignConsecutiveDeclarations` will turn

```
int a = 1;
float bbb = 2.0;
double c = 3.0;
```

into

```
int    a   = 1;
float  bbb = 2.0;
double c   = 3.0;
```

And `AlignConsecutiveAssignments` applies to assignments both in declarations and not
```
a = 1;
bbb = 2.0;
c = 3.0;
```

into
```
a   = 1;
bbb = 2.0;
c   = 3.0;
```

It does not relax the requirement.

I personally don't want these to be set to true. You end up with a lot of changes like:
```
T identity = std::numeric_limits<T>::lowest();
const int size = accessor.m_size;

using for_policy = typename Exec::for_policy;
using reduce_policy = typename Exec::reduce_policy;
```

becoming 
```
T         identity = std::numeric_limits<T>::lowest();
const int size     = accessor.m_size;

using for_policy_thing    = typename Exec::for_policy;
using reduce_policy_thing = typename Exec::reduce_policy;
```
Which looks wonky to me personally

emily-howell: It does look like the default behavior is to not align things

.clang-format
emily-howell: What I don't like is that when you make that change you get:

```
conduit::Node ASCENT_API field_reduction_max(
    const conduit::Node &field, const std::string &component = "");
```

rather than when it is false it gives:
```
conduit::Node ASCENT_API
history_gradient_range(const conduit::Node &y_values,
                       const conduit::Node &dx_values);
```

.clang-format
emily-howell: `Always` isn't an option but I can change it to `BeforeHash` which will also indent the hash.

.clang-format
emily-howell: Hmm. I don't know that I agree with this. That change would make header files significantly harder to read.

`AccessModifierOffset: -4`
```
class ASCENT_API ExpressionEval
{
protected:
    DataObject m_data_object;
    flow::Workspace w;
    static Cache m_cache;
    void jit_root(conduit::Node &root, const std::string &expr_name);

public:
    ExpressionEval(DataObject &dataset);
    ExpressionEval(conduit::Node *dataset);
    DataObject &data_object();
}
```


`AccessModifierOffset: 0`
```
class ASCENT_API ExpressionEval
{
    protected:
    DataObject m_data_object;
    flow::Workspace w;
    static Cache m_cache;
    void jit_root(conduit::Node &root, const std::string &expr_name);

    public:
    ExpressionEval(DataObject &dataset);
    ExpressionEval(conduit::Node *dataset);
    DataObject &data_object();
```

src/libs/ascent/runtimes/ascent_empty_runtime.cpp
emily-howell: If you add comments it won't put them on the same line anymore.

src/libs/ascent/runtimes/ascent_empty_runtime.cpp
emily-howell: Updated to only allow single line for empty functions

src/libs/ascent/hola/ascent_hola.hpp
emily-howell: Unfortunately, there isn't a ton of knobs to turn with respect to the preprocessor directives. The only thing I really found was to indent or not to indent. I personally really like the indentation as it greatly improves readability but I do agree that having a blanket policy for it means that some things have slightly wonky indentation.

src/libs/ascent/runtimes/ascent_empty_runtime.cpp
JustinPrivitera: Yikes! I rely on that many places in Conduit: https://github.com/LLNL/conduit/blob/8f5a32ae6540eaead82d287f231fe84d0cd8540a/src/libs/relay/conduit_relay_io_silo.cpp#L4854

.clang-format
JustinPrivitera: Does it let you put the arguments on multiple lines when set to `true`?

.clang-format
JustinPrivitera: Hah! Thanks Livchat. `BeforeHash` is the right answer then :)

.clang-format
JustinPrivitera: I blame Livchat again. I like the `-4`. I (mistakenly) thought `-4` and `0` would have the opposite effect of what they actually have.

src/libs/ascent/runtimes/ascent_empty_runtime.cpp
emily-howell: ^ It would support that case Justin

The comments make it wrap, even on short lines

.clang-format
emily-howell: It does, but it seems to prioritize grouping things in a way I don't love. When there are enough arguments or the named are long enough both true and false will wrap the arguments onto their own lines

.clang-format
emily-howell: LivChat really does hallucinate a lot on this topic unfortunately. It is especially disappointing when I ask for a setting to change something and it makes up something only for that setting to not really exist :(.

src/libs/ascent/runtimes/ascent_empty_runtime.cpp
JustinPrivitera: > If you add comments it won't put them on the same line anymore.

I'm confused. So comments will be on the same line?

Oh I get it! When you said "them" you meant the arguments, not the comments for those arguments. 🙃

emily-howell: Yes, sorry, that wasn't clear

It will absolutely respect that comments are placed next to the arguments and it will wrap all of the arguments onto their own lines.

JustinPrivitera: No worries. I was reading too fast.

PR #1502, opened 2025-04-16
Task/2025 4 15 1275 replay parallel multi timestep mode

### LINKED ISSUE #1275: replay parallel multi-timestep mode
given mpi env, split into sub coms and run parallel in time actions as well as dist-memory parallel.

emily-howell:
First pass at adding parallel multi timesteps. 

Something we might want to change is the default number of sub comm groups created


emily-howell:
## TLDR: 
comm_id values are actually opaque handles that do not necessarily have a correlation to the comm of a thread. Instead, we have to pass the "color" value used to split the threads into the sub_comm in order to differentiate ranks from different communicators.

----
## Full write up: (for documentation sake)
One of the requests for this MR was to make the log file names generated to be unique. The problem being that if the threads are separated into two groups that there would be duplicate rank values so rank alone can no longer be used as a unique value. Example being if there are 4 threads (ranks 0, 1, 2, and 3) split evenly into two sub_comm groups then each group would have two threads ranked 0 and 1. This was causing file collisions in the logging outputs since two threads were trying to write to the same value.The first idea was to use the comm_id value in combination with the rank to create unique file names. However, a few simple test cases identified the following behavior:

```
Rank 0 is now sub_rank 0 using color 0 on sub_comm_id -1006632960
Rank 1 is now sub_rank 0 using color 1 on sub_comm_id -2080374780
Rank 2 is now sub_rank 0 using color 2 on sub_comm_id -2080374780
Rank 3 is now sub_rank 1 using color 0 on sub_comm_id -2080374780
Rank 4 is now sub_rank 1 using color 1 on sub_comm_id -2080374780
Rank 5 is now sub_rank 1 using color 2 on sub_comm_id -2080374780
Rank 6 is now sub_rank 2 using color 0 on sub_comm_id -2080374780
Rank 7 is now sub_rank 2 using color 1 on sub_comm_id -2080374780
Rank 8 is now sub_rank 2 using color 2 on sub_comm_id -2080374780
Rank 9 is now sub_rank 3 using color 0 on sub_comm_id -2080374780
```

The expected behavior was that the comm_id values would be unique to the communicator. However, the above shows that not only did threads the same communicator not have the same comm_id but further that threads from multiple communicators could have the same comm_id.

After some investigation into how groups, communicators, and handles work in MPI (see: [info about communicators](https://hpc-tutorials.llnl.gov/mpi/management_routines/) and [info about handles and opaque objects](https://www.mpi-forum.org/docs/mpi-3.1/mpi31-report/node25.htm)) it was concluded that the above assumption that those values would be unique was incorrect. Rather, comm_id is an integer value known as an opaque handle that is used by MPI to identify where in memory to look for the information about the thread's communicator. The handle can not be used by the programmer to identify the communicator themselves.

src/utilities/replay/replay.cpp
```diff
@@ -271,21 +284,39 @@ main(int argc, char *argv[])
...
+  int rank, sub_rank = 0;
+  int num_process_groups = 1;
 #if defined(ASCENT_REPLAY_MPI)
   MPI_Init(NULL,NULL);
   MPI_Comm_size(MPI_COMM_WORLD, &comm_size);
   MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+
+  // Determine how many sub_comms we should be creating
+  num_process_groups = options.m_num_groups > 0 ? options.m_num_groups 
+                                                : static_cast<int>(std::floor(std::sqrt(comm_size)));
```
cyrush: Does this mean we always do a sub-com split?

I think we should only to parallel in time if they user asks for groups.

emily-howell: Okay, I got myself all straightened out. I added a flag variable that controls if parallel in time is being used. It only gets turned on if `--groups` or `--groups=[number]` is used. Otherwise replay will act as normal.

cyrush (approved):
Looks good, thanks for getting this feature in!

PR #1511, opened 2025-05-08
Fix conflicting family value found in output test file names

emily-howell:
Temporary fix to get the tests working again. Solves the problem that when tests are run multiple times that they start to fail because of naming conflicts even with different extensions. Will revert this change when file name extensions fix is added


cyrush (approved):
Thanks!

PR #1523, opened 2025-05-13
update min cmake version to 3.23

### LINKED ISSUE #1501: CMake 4.0 incompatibility
When building an application code with CMake 4.0, it complains that the Ascent CMake configs are incompatible with CMake 4.0:

```
CMake Error at /Users/benwibking/spack/opt/spack/darwin-m1/ascent-develop-rlanj2cqymzd46b2bswskf7eniydoyxe/lib/cmake/ascent/AscentConfig.cmake:16 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
Call Stack (most recent call first):
  extern/amrex/Tools/CMake/AMReXThirdPartyLibraries.cmake:107 (find_package)
  extern/amrex/Src/CMakeLists.txt:49 (include)
```

cyrush:
resolves #1501


cyrush:
Default cmake on llnl cts2 systems is 3.23.

We need 3.24+ for HIP builds, llnl toss4 amd systems have 3.23 -- but I am going to back off strict requirement of 3.24 to 3.23 to avoid extra hassle.

Issue #1522, opened 2025-05-13
Running tests repeatedly is still causing filenames to not line up


JustinPrivitera:
related to #1518 

I am seeing successive runs of the tests yield the following file names

```
tout_rover_xray_multi_curv3d48.png
tout_rover_xray_multi_curv3d49.png
tout_rover_xray_multi_curv3d50.png
```

The cycle number in multi_curv3d is 48.

JustinPrivitera:
Ascent should overwrite the file each time instead of making a new one with the wrong cycle number.

cyrush:
we need to separate the cycle case from the family case

cycle case should always overwrite the cycle

family case should do the increment logic

cyrush:
@emily-howell  I think we need a few tweaks and maybe more tests to the new path logic. 
#1518  also related

JustinPrivitera:
We also have the `filename{cycle}` yielding `filename{cycle}.png` issue. I am not sure how that is related or if that is something that is supposed to work already.

emily-howell:
This was addressed in #1511 and #1524. 

The issue was that family value was being used as the default for formatting rather than cycle. Additionally, a default format for keyword values is now supported so `filename{cycle}` should work now.

Please let me know if there are any additional issues encountered with filename formatting.

Issue #1518, opened 2025-05-13
image output filenaming not following the cycle number


jfavre:
given, for example, a standard trigger condition "cycle() % 5 == 0", images are named:

density.00005.png  density.00010.png  density.00015.png  density.00020.png, etc..

ever since commit efbdebd5 , this is no longer true and images are instead named:

density.00005.png  density.00006.png  density.00007.png  density.00008.png. etc.

i.e. the first filename is correct (using the first cycle meeting the trigger condition), and all subsequent filenames are numbered sequentially.

cyrush:
Thanks for the report -- we are making a few changes that will give direct family, cycle, and time options in the path you provide.
There will be some slight adjustments needed, and we may have a few bugs. 

@jfavre can you share your yaml so we can test your direct case?

jfavre:
here is a sample yaml
- 
  action: "add_triggers"
  triggers: 
    t1: 
      params: 
        condition: "cycle() % 5 == 0"
        actions: 
          - 
            action: "add_pipelines"
            pipelines: 
              pl_threshold_thin_clip_z: 
                f1: 
                  type: "threshold"
                  params: 
                    field: "z"
                    min_value: 0.12425
                    max_value: 0.12575
          - 
            action: "add_scenes"
            scenes: 
              s1: 
                plots: 
                  p1: 
                    type: "pseudocolor"
                    field: "Temperature"
                    pipeline: "pl_threshold_thin_clip_z"
                renders: 
                  r1: 
                    image_prefix: "datasets/Temperature.%05d"

emily-howell:
Apologies for the changes to the file naming conventions. This should be fixed in #1524. 

Cycle is now the default value for non-keyword formatting again and is automatically added to the end of filenames with no formatting. Please note that for filenames with no formatting specified that do not end in an underscore (`_`), that the cycle value will now be appended with one and that the default cycle value has been expanded to 6 digits. E.g. `./filename` will become `./filename_000001`. If that is not the desired behavior, the format can be modifies using either basic formatting `"./filename_%06d"` or with the `cycle` keyword `"./filename_{cycle:%06d}"`.

Please let me know if you are still experiencing unexpected behavior.

PR #1524, opened 2025-05-13
DRAFT: Task/7 5 2025 fmt string followup

emily-howell:
Looking to address some of the issues uncovered with the original implementation. 

Reverts #1511 and actual implementation for extension checking is added here

#1522 and #1518 document issues with the default value using family and not cycle which is also addressed here.

DRAFT subject to changes and discussion on requirements for path formatting


src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -335,11 +345,11 @@ RoverXRay::execute()
       bov_filename = output_dir(bov_filename);
       if(cycle != -1)
       {
-        tracer.save_bov(expand_path_special_variables(bov_filename, mpi_comm_id, cycle));
+        tracer.save_bov(expand_path_special_variables(bov_filename, ".png", mpi_comm_id, cycle));
```
cyrush: this case would be `".bov"`

emily-howell:
The current state of this PR includes the following changes:
- [x] Reverts the patch fix made in #1511 
- [x] Tracks the file path extensions in family value checks
  - [x] This also fixes the failing tests caused by file conflicts
- [x] Default value is cycle again rather than family value
- [x] Family value has been decoupled from cycle. Can now pass a family value seed as an option that gets stored in the metadata
- [x] Patterns not including a format such as "filename{cycle}" and "filename{cycle:}" are now supported
- [x] Fixed extension for bov files 

Outstanding:
- [x] Using the family value as a backup if cycle not available for default formatting. Concern that this might be confusing to users still stands.
- [x] Underscore added to default
- [x] default should follow format set by table
- [ ] Improve documentation
- [ ] Add a set of tests that does everything in the "table" and documentation

src/libs/ascent/utils/ascent_string_utils.cpp
```diff
@@ -50,6 +50,28 @@ void split_string(const std::string &s,
 
 } // namespace detail
 
+std::vector<std::string> split(const std::string &s, char delim)
+{
+  std::vector<std::string> elems;
+  detail::split_string(s, delim, elems);
+  return elems;
+}
```
cyrush, line 58: You can use Conduit's string splitter util

https://github.com/LLNL/conduit/blob/b8f4ad0858980e430661b2d381645bd1de5c0e4b/src/libs/conduit/conduit_utils.hpp#L350

emily-howell:
@cyrush I think this has all the requests (minus robust documentation which you said should go in a future PR). I added the family value as a fallback, default formats (`06d` for cycle and family `g` for time) and an `_` in front of the default fmt value iff the path does not already end with a `_` character.

As a note, those changes, both the default format of `06d` and the `_`, changed how the vast majority of our output images from tests are named so there are a lot of renamed baseline_images but none of the images themselves have been changed.

I also got annoyed with how `check_test_image` and `remove_test_image` worked so I added an overload option for both that can take a cycle number in as an int and format it. This is nice because it means that we can make changes to the formatting and only have to change that one function rather than every instance of the function call. I did notice that a number of places that use a custom value for `check_test_image` don't also pass that value to `remove_test_image` so that should get fixed. For the `ASCENT_ACTIONS_DUMP` I don't think you can overload macros? so I made a few different options that also allow you to specify the cycle and format.

src/tests/t_utils.hpp
```diff
@@ -682,16 +705,25 @@ add_matset_to_spiral(Node &n_mesh, const int ndomains)
...
   out<<actions_str; \
   out.close();
 
+// Macro to save ascent actions file with a custom cycle number
+#define ASCENT_ACTIONS_DUMP_CYCLE(actions,name,msg,num) \
+  ASCENT_ACTIONS_DUMP_CYCLE_FMT(actions,name,msg,num,"%06d")
+
+// Macro to save ascent actions file with default cycle and formatting
+#define ASCENT_ACTIONS_DUMP(actions,name,msg) \
+  ASCENT_ACTIONS_DUMP_CYCLE(actions,name,msg,100)
```
cyrush, line 726: these macros are nice

cyrush (approved):
Looks good, thanks for working through this!

PR #1527, opened 2025-05-16
Update Uniform Sample to accept and output multiple fields

nicolemarsaglia:
Now takes in a list of fields.


src/libs/ascent/runtimes/flow_filters/ascent_runtime_vtkh_filters.cpp
```diff
@@ -3794,17 +3805,40 @@ VTKHUniformGrid::execute()
...
+        bool throw_error = false;
+        detail::field_error(field_selection[i], this->name(), collection, throw_error);
+        // this creates a data object with an invalid soource
+        set_output<DataObject>(new DataObject());
+        return;
+      }
     }
 
-
+    std::string field = field_selection[0];
```
cyrush, line 3857: is this now unused?

src/libs/ascent/runtimes/flow_filters/ascent_runtime_vtkh_filters.cpp
```diff
@@ -3794,17 +3805,40 @@ VTKHUniformGrid::execute()
     }
     std::shared_ptr<VTKHCollection> collection = data_object->as_vtkh_collection();
 
-    std::string field = params()["field"].as_string();
-    if(!collection->has_field(field))
+    //std::string field = params()["field"].as_string();
```
cyrush: I think we want to support both `field` and `fields`, since we have folks using the `field` form.
If only `field` is provided, we can add it to the string list.
If both are provided, we can just use the `fields` value.

src/libs/vtkh/filters/UniformGrid.hpp
```diff
@@ -21,7 +21,7 @@ class VTKH_API UniformGrid : public Filter
   void Dims(const Vec3f dims);
   void Origin(const Vec3f origin);
   void Spacing(const Vec3f spacing);
-  void Field(const std::string field);
+  void Field(const std::vector<std::string> fields);
```
cyrush: Lets change this to `Fields` to reflect it accepts a vector of names.

src/tests/ascent/t_ascent_uniform_grid.cpp
```diff
@@ -33,6 +33,16 @@ using namespace ascent;
...
+//- 
+//  action: "add_extracts"
+//  extracts: 
+//    e1: 
+//      type: "relay"
+//      pipeline: "pl1"
+//      params: 
+//        path: "uni_grid_sample_in_x"
+//        protocol: "blueprint/mesh/hdf5"
+
```
cyrush: we can remove this commented out block

cyrush (commented):
A few final suggestions

src/libs/ascent/runtimes/flow_filters/ascent_runtime_vtkh_filters.cpp
nicolemarsaglia: We're still using it to grab the topo and then data

nicolemarsaglia:
> A few final suggestions

Thanks! I think I've hit them all.

cyrush (approved):
Looks good, thanks!

Issue #1513, opened 2025-05-09
X Ray Filter Enhancements


JustinPrivitera:
## High-Level Goals
- Achieve feature parity between the Ascent X-Ray Extract (aka Rover) output and VisIt’s X-Ray Image Query Blueprint Output, including:
    - Meshes representing the camera frustum, spatial extents, and image extents
    - 1D spectra curves
    - Metadata describing the camera/ray trace setup
    - Fix issues in the X-Ray Extract.
    - Considerations for performance and memory utilization at scale.
- Develop tests for the upgraded X-Ray Extract.
    - Refactor the serial and MPI tests to use shared utility functions.
    - Refactor the diff tests to compute what the baseline node should have based on the input actions. 
- Write comprehensive documentation for the upgraded X-Ray Extract.
   - Including code comments where they make sense.

### Output and Feature Parity with VisIt
- [x] Match Blueprint Output shape to VisIt (remove cube stretching) - #1529
- [x] Improve default color table for the X-Ray Extract - #1529
- [x] Add tests and baselines for provided test files - #1529
- [x] Add tests covering the full parameter space of X Ray Filter arguments - #1541
- [x] Basic Mesh Output - #1535
- [x] Match dimension ordering (`z` for energy groups as in VisIt)
- [x] Use the same topology and coordinate set as VisIt
- [ ] Comparable Ascent X Ray Filter arguments to VisIt X Ray Image Query arguments ([reference](https://visit-sphinx-github-user-manual.readthedocs.io/en/develop/using_visit/Quantitative/XRayImageQuery.html#query-arguments))
    - [x] Add `divide_emis_by_absorb` flag
    - [ ] Add support for other standard arguments (need to be more specific and see what we're still missing)
- [ ] Ensure output filenames and directories are handled the same as in VisIt
- [ ] Support all output types
    - [x] Conduit Blueprint
    - [ ] PNG (not tested)
    - [ ] BOV (not tested)
    - [ ] Support parallelized outputs?
- [ ] Support for passing field units through to the output.
- [ ] Camera specification (use complete camera spec, consider near/far clipping planes)
    - [x] Support for specifying non-default camera params: near and far clipping planes, position, look at, etc
    - [ ] Support for passing VisIt camera params and mapping those onto ascent's VTK-m camera
- [x] Ensure optical depth and intensities outputs align with VisIt
- [ ] Add/verify metadata fields:
    - [ ] #1598
    - [x] view - #1541
    - [x] query - #1551
[...]

PR #1529, opened 2025-05-20
X Ray Filter First Enhancements

siramok:
Resolves bullets in #1513
 - Adds test data to the repo.
 - Adds a test for multi_curv3d and a baseline image.
 - Fixes the "cool2warm" color table warning by using "Cool to Warm" instead
 - Adds skeleton tests for new data
 - Undoes rover's mesh scaling


JustinPrivitera:
Let's do the following and then merge:
- [x] figure out how to make the output not a box (and check results in VisIt) (and update baselines as needed)
- [x] add a test for `curv3d` with a baseline
- ~~[ ] add a test for `tire` with a baseline~~

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -274,8 +276,7 @@ RoverXRay::execute()
 
     if(params().has_path("blueprint"))
     {
-
-
+      // TODO: validate that protocol is a valid string in verify_params
       std::string protocol = params()["blueprint"].as_string();
```
JustinPrivitera, line 280: we also may want to check that the blueprint protocol is one of the allowed types "hdf5", "yaml", and "json", and error if it is not one of those. Not sure if that is handled elsewhere.

src/libs/rover/rover_types.hpp
```diff
@@ -11,19 +11,21 @@
...
 
 namespace rover {
+
 // this could be ray tracing(surface) / volume rendering / energy
 enum RenderMode
 {
   volume, // standard volume rendering (3d)
   energy, // radiography with enegry bins (3d)
-  surface //
+  surface // ?
```
JustinPrivitera: I don't think we care about whatever the rover surface render mode is. I don't recall seeing any code for this anywhere else.

src/libs/rover/rover_types.hpp
```diff
@@ -69,8 +68,7 @@ struct RenderSettings
   //
   // Default settings
   //
-  RenderSettings()
-    : m_color_table("cool2warm")
+  RenderSettings() : m_color_table("Cool to Warm")
```
JustinPrivitera, line 71: Thanks for this fix 💪

src/libs/rover/scheduler.cpp
```diff
@@ -634,15 +634,13 @@ void Scheduler<FloatType>::to_blueprint(conduit::Node &dataset)
   n_coords["origin/y"] = 0;
   n_coords["origin/z"] = 0;
 
-  // scale the group spacing so we get more squarish data set
-  n_coords["spacing/dx"] = float(width) / float(num_channels);
+  n_coords["spacing/dx"] = 1.f;
```
JustinPrivitera, line 637: good find

src/tests/ascent/t_ascent_rover.cpp
```diff
@@ -19,70 +19,307 @@
...
+{
+    //
+    // Create the actions.
+    //
+
+    Node scenes;
+    scenes["s1/plots/p1/type"] = "pseudocolor";
+    scenes["s1/plots/p1/field"] = field_name;
+    scenes["s1/renders/r1/image_prefix"] = output_file;
+    // TODO: Revisit this once we figure out why/where the data is rotated in the first place
```
JustinPrivitera, line 49: 👍

src/tests/ascent/t_ascent_rover.cpp
```diff
@@ -19,70 +19,307 @@
...
+
+bool is_vtk_disabled(const Node &about)
+{
+    bool vtk_disabled = about["runtimes/ascent/vtkm/status"].as_string() == "disabled";
+    if (vtk_disabled)
+    {
+        ASCENT_INFO("vtkm support disabled, skipping test\n");
+    }
+    return vtk_disabled;
+}
```
JustinPrivitera: should probably be named `is_vtkm_disabled`. `vtk` is a different library.

src/tests/ascent/t_ascent_rover.cpp
```diff
@@ -19,70 +19,307 @@
...
+    const std::string root_file = 
+        conduit::utils::join_file_path(std::string(ASCENT_T_DATA_DIR),
+                                       "curv3d_blueprint.cycle_000048.root");
 
-    EXPECT_TRUE(conduit::blueprint::mesh::verify(data,verify_info));
+    // test files:
+    //  - curv3d_blueprint.cycle_000048.root
+    //  - multi_curv3d_blueprint.cycle_000048.root
+    //  - tire_blueprint.cycle_000000.root
+    //  - curv2d_blueprint.cycle_000048.root - we are not sure what will happen for 2D inputs.
```
JustinPrivitera: This comment was mainly for you so you know what is available. Now that you have added tests and skeleton tests for these we don't need a note about the options here. There will be more test files coming soon too.

src/tests/ascent/t_ascent_rover.cpp
```diff
@@ -19,70 +19,307 @@
...
+    Node data, verify_info;
+    const std::string root_file = 
+        conduit::utils::join_file_path(std::string(ASCENT_T_DATA_DIR),
+                                       "multi_curv3d_blueprint.cycle_000048.root");
+
+    // test files:
+    //  - curv3d_blueprint.cycle_000048.root
+    //  - multi_curv3d_blueprint.cycle_000048.root
+    //  - tire_blueprint.cycle_000000.root
+    //  - curv2d_blueprint.cycle_000048.root - we are not sure what will happen for 2D inputs.
```
JustinPrivitera: same here - we don't need this in every test.

JustinPrivitera (changes requested):
Just a few minor things. Generally looks very good! Thanks for getting this in!

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
JustinPrivitera: We can do that later; I would just make a note.

PR #1530, opened 2025-05-20
Adding documentation that queries will not create a named output that can be referenced elsewhere

### LINKED ISSUE #1253: Queries not working when attached to pipeline
Workaround: Expressions should be used as queries. 

Expressions used as pipelines to make derived fields are losing the derived field for later use. i.e. make an image

emily-howell:


src/docs/sphinx/Actions/ExpressionsOverview.rst
```diff
@@ -277,6 +279,12 @@ of a contour mesh which is output by a pipeline.
 Queries like the one above will act on the data published to Ascent. Queries
 are also capable of acting on the results of pipelines.
 
+.. note::
+    `add_queries` does not create a named output that can be referenced by other
+    execution blocks. Rather, queries are a way to summarize and output data.
```
nicolemarsaglia: `add_queries` does not create a named output that can be referenced by other
    execution blocks. Rather, queries are a way to summarize and output data. -> `add_queries` does not create mesh fields that can be used in scenes or filters. It only produces scalar results for analysis or logic. To generate fields usable in visualizations or pipelines, use a derived field expression inside a pipeline

emily-howell: That's great! Thank you

nicolemarsaglia (approved):
Thanks for getting this added!

cyrush (approved):
Looks good, thanks!

PR #1535, opened 2025-05-21
Initial work for getting basic blueprint mesh output parity between rover and visit

siramok:
Addresses bullets in #1513
- Output meshes are now defined with a rectilinear coordset (what visit outputs) instead of a uniform coordset
- Added, removed, and reordered various Conduit fields to improve parity with VisIt
- A few additional rover options (blueprint, width, height) are now validated in verify_params before executing
- Cleanup/better formatting of ascent_runtime_rover_filters.cpp
- Dimension ordering now matches VisIt
- Striding now matches VisIt more closely
- The default output size has been reduced to 200x200 from 1024x1024 to match VisIt


src/libs/rover/scheduler.cpp
```diff
@@ -619,27 +619,34 @@ void Scheduler<FloatType>::to_blueprint(conduit::Node &dataset)
...
+
+  // Fill each vector with ascending values from [0, vector_size]
+  std::iota(x_coords.begin(), x_coords.end(), 0);
+  std::iota(y_coords.begin(), y_coords.end(), 0);
+  std::iota(z_coords.begin(), z_coords.end(), 0);
+
+  // Can't use set_external here since these vectors will go out of scope
+  n_coords["values/x"].set(x_coords);
+  n_coords["values/y"].set(y_coords);
+  n_coords["values/z"].set(z_coords);
```
JustinPrivitera: There is a better way to set the coordinates, as discussed over teams.

src/libs/rover/scheduler.cpp
```diff
@@ -619,27 +619,34 @@ void Scheduler<FloatType>::to_blueprint(conduit::Node &dataset)
 
   const int num_channels = m_result.get_num_channels();
 
-  conduit::Node &n_topo = dataset["topologies/"+topo_name];
+  conduit::Node &n_coords = dataset["coordsets/" + coord_name];
+  n_coords["type"] = "rectilinear";
+  
+  std::vector<float> x_coords(num_channels + 1);
+  std::vector<float> y_coords(width + 1);
+  std::vector<float> z_coords(height + 1);
```
JustinPrivitera: We will want to change the order of these dimensions as discussed.

src/libs/rover/scheduler.cpp
```diff
@@ -651,31 +658,25 @@ void Scheduler<FloatType>::to_blueprint(conduit::Node &dataset)
       conduit::Node &n_int = dataset["fields/intensities"];
       n_int["topology"] = topo_name;
       n_int["association"] = "element";
+      n_int["units"] = "intensity units";
```
JustinPrivitera: For now this is fine. In reality, the units will be passed through the filter as arguments and they will show up here. Later on we will want to make this happen.

src/libs/rover/scheduler.cpp
```diff
@@ -651,31 +658,25 @@ void Scheduler<FloatType>::to_blueprint(conduit::Node &dataset)
...
-      n_int["labels"].append() = "width";
-      n_int["labels"].append() = "height";
     }
 
     if(m_result.has_optical_depth(0))
     {
       conduit::Node &n_op = dataset["fields/optical_depth"];
       n_op["topology"] = topo_name;
       n_op["association"] = "element";
+      n_op["units"] = "path length metadata";
```
JustinPrivitera: Same thing as the "intensity units". We will also have to decide if this is a good default value.

src/libs/rover/scheduler.cpp
```diff
@@ -651,31 +658,25 @@ void Scheduler<FloatType>::to_blueprint(conduit::Node &dataset)
       conduit::Node &n_int = dataset["fields/intensities"];
       n_int["topology"] = topo_name;
       n_int["association"] = "element";
+      n_int["units"] = "intensity units";
       vtkm::cont::ArrayHandle<FloatType> ints = m_result.flatten_intensities();
       FloatType *ints_buffer = get_vtkm_ptr(ints);
       // can't set external since this goes out of scope
       n_int["values"].set(ints_buffer, ints.GetNumberOfValues());
-      n_int["shape"].set(shape);
       n_int["strides"].set(strides);
```
JustinPrivitera: I'm not convinced that the data is strided the way we want it to be strided.

We will need to investigate/discuss further. Perhaps we need to create some really low resolution examples (like 4x5 images) to understand how this differs from VisIt.

I think a good TODO would be to add a test that runs the Filter at a lower resolution (than 1024x1024). It would be good to know how much control we really have over the camera. What is default; what is able to be specified? We can do a deeper dive into the camera in a follow-on PR.

JustinPrivitera (changes requested):

src/libs/rover/scheduler.cpp
siramok: I added a todo for this to #1513

JustinPrivitera: We probably want most tests to be at a lower resolution than 1024x1024. I think the defaults should be changed too to match VisIt.

siramok:
The low-res test has revealed some additional parity differences between xraying in ascent and visit. I simply added optional width and height parameters to the rover filter, with the assumption that there's a more standardized approach/naming that it will need to be aligned with before merging.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -96,255 +94,315 @@ RoverXRay::~RoverXRay()
...
-      if( !params.has_path("image_params/min_value") ||
-         ! params["image_params/min_value"].dtype().is_number() )
-      {
-          info["errors"].append() = "Missing required image parameter 'min_value' must be a number";
-          res = false;
-      }
-
-      if( !params.has_path("image_params/max_value") ||
-         ! params["image_params/max_value"].dtype().is_number() )
+      info["errors"].append() = "Optional parameter 'width' must be an integer";
```
JustinPrivitera, line 135: Above you only check is it is a number, not necessarily an integer. You can ask for integer probably with `is_integer()`.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -96,255 +94,315 @@ RoverXRay::~RoverXRay()
...
+  else if (has_width && !has_height)
+  {
+    info["errors"].append() = "Optional parameter 'width' requires 'height' to also be set";
+    res = false;
+  }
+  else if (!has_width && has_height)
+  {
+    info["errors"].append() = "Optional parameter 'height' requires 'width' to also be set";
+    res = false;
+  }
```
JustinPrivitera, line 172: This is good for now but we should ultimately move towards having a bunch of default params and if you pass stuff in then it modifies the defaults. So in that case it would be ok to pass in just width and not height, etc.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -96,255 +94,315 @@ RoverXRay::~RoverXRay()
...
+  camera.ResetToBounds(dataset.GetGlobalBounds());
+
+  if(params().has_path("camera"))
+  {
+    const conduit::Node &n_camera = params()["camera"];
+    parse_camera(n_camera, camera);
+  }
+
+  int width = 1024;
+  int height = 1024;
```
JustinPrivitera: We'll want a class to have these values instead of hardcoding defaults here.

src/tests/ascent/t_ascent_rover.cpp
```diff
@@ -165,9 +165,92 @@ TEST(ascent_rover, test_xray_blueprint_braid)
...
-    // by working on the basic mesh output and changing the order of the dimensions.
-    // TODO we will need to change the baseline when we are making good renders.
+    EXPECT_TRUE(check_test_image(image_output_base, 0.01f, "100"));
+
+    std::string msg = "TODO we need a good description here";
+    ASCENT_ACTIONS_DUMP(actions, image_output_base, msg);
+}
+
+//-----------------------------------------------------------------------------
+TEST(ascent_rover, test_xray_blueprint_braid_lowres)
```
JustinPrivitera, line 175: I'll bet we could refactor more of the testing infrastructure out into helpers.

src/libs/rover/scheduler.cpp
siramok: VisIt's default is 200x200, so I updated our default to match.

src/tests/ascent/t_ascent_rover.cpp
siramok: I bet we could

siramok:
I feel that this PR is in a good place and is ready for a final review. As we discussed offline, the feedback about hardcoded default camera values will be addressed in a new PR.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
siramok: Will be fixed in a follow-up PR as discussed offline.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
siramok: Will be fixed in a follow-up PR as discussed offline.

src/libs/rover/scheduler.cpp
```diff
@@ -655,47 +656,58 @@ void Scheduler<FloatType>::to_blueprint(Node &dataset)
...
+      intensities["units"] = "intensity units";
+      vtkm::cont::ArrayHandle<FloatType> intensity_values = m_result.flatten_intensities();
+      FloatType *intensity_buffer = get_vtkm_ptr(intensity_values);
       // can't set external since this goes out of scope
-      n_int["values"].set(ints_buffer, ints.GetNumberOfValues());
-      n_int["strides"].set(strides);
+      const int num_intensity_values = intensity_values.GetNumberOfValues();
+      intensities["values"].set(intensity_buffer, num_intensity_values);
+
+      // TODO: VisIt uses int64, but using it here throws an error later on
```
JustinPrivitera: What error do you see?

src/libs/rover/scheduler.cpp
```diff
@@ -655,47 +656,58 @@ void Scheduler<FloatType>::to_blueprint(Node &dataset)
...
+      optical_depth["units"] = "path length metadata";
+      vtkm::cont::ArrayHandle<FloatType> optical_values = m_result.flatten_optical_depths();
+      FloatType *optical_buffer = get_vtkm_ptr(optical_values);
       // can't set external since this goes out of scope
-      n_op["values"].set(ints_buffer, ints.GetNumberOfValues());
-      n_op["strides"].set(strides);
+      const int num_optical_values = optical_values.GetNumberOfValues();
+      optical_depth["values"].set(optical_buffer, num_optical_values);
+      
+      // TODO: Why does VisIt only set the first two stride values?
```
JustinPrivitera: confused what you mean by this

src/libs/rover/scheduler.cpp
```diff
@@ -655,47 +656,58 @@ void Scheduler<FloatType>::to_blueprint(Node &dataset)
...
   }
 
-  //relay::io::blueprint::save_mesh(n_dataset, root_file, protocol);
-  Node info;
-  if(!blueprint::verify("mesh",dataset, info))
+  Node verify;
+  if(!blueprint::verify("mesh", data, verify))
   {
-    info.print();
+    ROVER_ERROR("to_blueprint failed to produce a valid conduit mesh");
```
JustinPrivitera: would be useful to include the `info.print()` in the error message

`<< info.to_yaml()`

src/libs/rover/scheduler.cpp
siramok: See here: https://github.com/visit-dav/visit/blob/9d51598a3c14cb4485db55e6c1cee1d6f807c593/src/avt/Queries/Queries/avtXRayImageQuery.C#L2920

JustinPrivitera: The link you sent is for the spatial energy reduced intensities, which is a field for a topo we will need to add later.

The optical depth/path length strides are set in the same spot as the intensities strides. They are identical. I used `set` to copy the values instead of specifying them again. See this line: https://github.com/visit-dav/visit/blob/9d51598a3c14cb4485db55e6c1cee1d6f807c593/src/avt/Queries/Queries/avtXRayImageQuery.C#L2852

src/libs/rover/scheduler.cpp
siramok: Here's an example:
```
[Error] Ascent::execute 
file: /usr/WS1/sewell10/ascent/tpls/source/conduit-v0.9.4/src/libs/conduit/conduit_node.cpp
line: 16712
message: 
Node::as_signed_int_array() -- DataType int64 at path [0]/fields/intensities/strides does not equal expected DataType int32

unknown file: Failure
C++ exception with description "
file: /usr/WS1/sewell10/ascent/tpls/source/conduit-v0.9.4/src/libs/relay/conduit_relay_io_blueprint.cpp
line: 2302
message: 
failed to open root file: /usr/workspace/sewell10/ascent/build/tests/_output/tout_rover_xray_query100.cycle_000100.root
" thrown in the test body.
```

JustinPrivitera: So where does this happen? Can you narrow down where the error is thrown? We need to know if it is something in Rover, or if this is an issue with loading the file with Conduit Relay, or if this is a problem with Ascent more generally.

src/libs/rover/scheduler.cpp
siramok: Makes sense, the error now includes the node contents

src/libs/rover/scheduler.cpp
siramok: Ah, I see. I'll fix that now

src/libs/rover/scheduler.cpp
siramok: Turned out to be my own mistake, using int64 for strides works as expected.

JustinPrivitera (approved):
Looks good; thanks for the iteration on this. 💪

PR #1541, opened 2025-05-28
Refactor XRay initialization + camera params

siramok:
Addresses bullets in #1513
- Adds a new test + baseline image that sets non-default camera params (minus position, which currently seems to be broken if anything non-default is set)
- Adds a description for each test, removing the placeholder
- Initial work for using a conduit node to handle all of rover's settings
    - We pass the input params to rover to be parsed
    - Currently, only camera params are wired up. Iff non-default values are set, then we adjust the camera accordingly
    - Cleaned up *_rover_filters.cpp by moving all of the camera + camera_generator code to rover's internals
- Add xray_view information to the conduit output


src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -291,33 +272,57 @@ RoverXRay::execute()
...
-  RenderSettings settings;
-  settings.m_render_mode = rover::energy;
-  settings.m_primary_field = absorption;
-    
-  // TODO: investigate how/why this is getting set, even if emission is not specified
-  // example: if absorption == "radial", why is emission also == "radial"
-  if (params().has_path("emission"))
+  vtkmCamera camera = tracer.get_camera();
+  // TODO: Resetting the camera bounds within add_data_set causes tests to fail
+  camera.ResetToBounds(dataset.GetGlobalBounds());
```
JustinPrivitera: I am curious how they fail, but perhaps that mystery can wait until after we've changed how we plumb the all the parameters.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -291,33 +272,57 @@ RoverXRay::execute()
...
-  tracer.set_render_settings(settings);
+  tracer.set_ray_generator(&camera_generator);
 
-  for(int i = 0; i < dataset.GetNumberOfDomains(); i++)
+  //
+  // Default render settings
+  //
+  RenderSettings render_settings;
+  // TODO: Changing energy to be the default render mode breaks things
+  render_settings.m_render_mode = rover::energy;
```
JustinPrivitera: We will want to find out why.

src/libs/rover/rover_types.hpp
```diff
@@ -65,16 +68,19 @@ struct RenderSettings
...
   RenderSettings() : m_color_table("Cool to Warm")
   {
+    // TODO: Why does changing the default to energy not work?
     m_render_mode     = volume;
     m_scattering_type = non_scattering;
     m_ray_scope       = global_rays;
   }
 
+  // Can we remove this?
   void print()
```
JustinPrivitera: Is it useful for debugging anything? Might not hurt to leave it in in that case.

src/tests/ascent/t_ascent_rover.cpp
```diff
@@ -337,13 +339,110 @@ TEST(ascent_rover, test_xray_blueprint_curv3d)
...
+    extracts["e1/type"] = "xray";
+    extracts["e1/params/absorption"] = "d";
+    extracts["e1/params/emission"] = "p";
+    extracts["e1/params/filename"] = query_output_file;
+    extracts["e1/params/blueprint"] = "json";
+
+    // These errors all originate from within rover
+    // TODO: Setting anything for position (e.g. 0,0,0) throws a vector range error
+    // TODO: Setting (0,0,0) for up throws a vector range error
+    // TODO: If xpan = 1 and ypan = 2, throws a vector range error
```
JustinPrivitera, line 396: We will want to solve these mysteries.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
siramok: I believe we figured out the order of operations in an offline discussion that fixes this problem.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
siramok: I removed the comment, since we determined in an offline discussion that we would rip out these setting structs in favor of using conduit throughout rover. We also decided to remove the volume renderer, making this option useless anyways.

src/libs/rover/rover_types.hpp
siramok: Same as above, we're going to remove these structs at some point anyways, ultimately making this comment + function irrelevant.

src/tests/ascent/t_ascent_rover.cpp
siramok: Added this under investigations in the meta-issue.

src/libs/rover/rover.cpp
```diff
@@ -210,15 +211,170 @@ class Rover::InternalsType
...
+
+  if (params.has_path("camera/image_width"))
+  {
+    m_settings["camera/image_width"] = params["camera/image_width"]; 
+  }
+
+  if (params.has_path("camera/image_height"))
+  {
+    m_settings["camera/image_height"] = params["camera/image_height"]; 
+  }
```
JustinPrivitera: What if instead of all of this we just said
```c++
if (params.has_child("camera"))
{
    m_settings["camera"].set(params["camera"]);
}
```
I think it is safe to trust for now. Whenever we use these parameters we should be asking if they exist before fetching them anyway.

src/libs/rover/rover.cpp
```diff
@@ -210,15 +211,170 @@ class Rover::InternalsType
...
+  {
+    float64_accessor view_up = m_settings["camera/up"].value();
+    vtkmVec3f up(view_up[0], view_up[1], view_up[2]);
+    m_camera.SetViewUp(up);
+  }
+  
+  if (m_settings.has_path("camera/fov"))
+  {
+    float64 view_angle = m_settings["camera/fov"].value();
+    m_camera.SetFieldOfView(view_angle);
```
JustinPrivitera: Let's have consistent names
```suggestion
    float64 fov = m_settings["camera/fov"].value();
    m_camera.SetFieldOfView(fov);
```

src/libs/rover/rover.cpp
```diff
@@ -210,15 +211,170 @@ class Rover::InternalsType
...
+
+void
+Rover::update_camera_generator()
+{
+  m_camera_generator.set_camera(m_camera);
+
+  int64 width = 200;
+  int64 height = 200;
+
+  if (m_settings.has_path("camera/image_width") || m_settings.has_path("camera/image_width"))
```
```suggestion
  if (m_settings.has_path("camera/image_width") || m_settings.has_path("camera/image_height"))
```

src/libs/rover/rover.cpp
```diff
@@ -210,15 +211,170 @@ class Rover::InternalsType
...
+    if (m_settings.has_path("camera/image_width"))
+    {
+      width = m_settings["camera/image_width"].value();
+    }
+
+    if (m_settings.has_path("camera/image_height"))
+    {
+      height = m_settings["camera/image_height"].value();
+    }
+  }
```
JustinPrivitera: I think the outer if statement here is redundant. Why not just check if we have width and height separately?

src/libs/rover/rover.cpp
```diff
@@ -210,15 +211,170 @@ class Rover::InternalsType
...
+
+  m_camera_generator.set_width(width);
+  m_camera_generator.set_height(height);
+  
+  Rover::set_ray_generator(&m_camera_generator);
```
JustinPrivitera: This is probably another place where names should change. Ray generator != camera generator but I guess they represent the same thing?

src/libs/rover/rover.cpp
```diff
@@ -210,15 +211,170 @@ class Rover::InternalsType
...
+    
+    if (m_settings.has_path("camera/far_plane"))
+    {
+      clipping_range.Max = m_settings["camera/far_plane"].value();
+    }
+
+    m_camera.SetClippingRange(clipping_range);
+  }
+
+  Rover::update_camera_generator();
```
JustinPrivitera: Do we need the namespacing? Or can we just call `update_camera_generator()` directly?

src/libs/rover/rover.cpp
JustinPrivitera: same namespace question here

src/libs/rover/rover.hpp
```diff
@@ -7,33 +7,43 @@
...
+  void update_settings(Node &params);
+  void print_settings();
+
   void set_mpi_comm_handle(int mpi_comm_id);
   int  get_mpi_comm_handle();
 
-  void finalize();
-
-  void add_data_set(vtkmDataSet &);
+  void add_data_set(vtkh::DataSet &);
```
JustinPrivitera, line 46: I am curious about this. Why change from vtkm to vtkh? I guess this lets you stuff the domain decomp down a layer?

src/tests/_baseline_images/tout_rover_xray_curv3d_camera_param48.png
JustinPrivitera, line 1: This is a good test case. It would be good to play with the camera params even more now that we have a handle on controlling the camera. Maybe something like this? https://visit-sphinx-github-user-manual.readthedocs.io/en/develop/_images/XRay_Query_spatial_energy_reduced_intensities_xray_side.png
https://visit-sphinx-github-user-manual.readthedocs.io/en/develop/_images/XRay_Query_spatial_intensities_side.png

src/tests/ascent/t_ascent_rover.cpp
```diff
@@ -211,8 +213,8 @@ TEST(ascent_rover, test_xray_blueprint_braid_lowres)
     extracts["e1/params/blueprint"] = "json";
 
     // Output resolution
-    extracts["e1/params/width"] = 11;
-    extracts["e1/params/height"] = 11;
+    extracts["e1/params/camera/image_width"] = 11;
+    extracts["e1/params/camera/image_height"] = 11;
```
JustinPrivitera, line 217: We might have to change these names. Let's discuss another time. Not a problem for this PR.

JustinPrivitera (changes requested):

src/libs/rover/scheduler.cpp
```diff
@@ -609,19 +609,30 @@ void Scheduler<FloatType>::save_bov(std::string file_name)
...
+  int height;
+  m_ray_generator->get_dims(width, height);
   ROVER_INFO("Saving blueprint file " << height << " "<<width);
 
   const std::string topo_name = "image_topo";
   const std::string coord_name = "image_coords";
 
   const int num_channels = m_result.get_num_channels();
 
+  Node &xray_view = data["xray_view"];
```
JustinPrivitera: should live under ["state/xray_view"]

JustinPrivitera (changes requested):
One little thing; otherwise all looks good.

src/libs/rover/rover.cpp
siramok: I agree with you, but after Cyrus said there would be a minor perf improvement I was thinking this could be a simple test case for me to familiarize myself with the fancy caliper macros to see if I measure any performance difference when going from has_path to has_child. A TODO for this has been added.

src/libs/rover/rover.cpp
siramok: Good catch.

src/libs/rover/rover.cpp
siramok: Double good catch.

src/libs/rover/rover.cpp
siramok: I agree, that will come in a future PR though.

src/libs/rover/rover.cpp
JustinPrivitera: I think we see the performance improvements at large scale. I'm guessing that just checking 10 things isn't going to impact it much. I think we will want 
```c++
if (params.has_child("camera"))
{
    m_settings["camera"].set(params["camera"]);
}
```
If you wish to do performance testing with caliper on this I would leave everything you've added but just comment it out for now, and circle back later.

src/libs/rover/rover.cpp
JustinPrivitera: sounds good

src/libs/rover/rover.cpp
siramok: Correct.

src/libs/rover/rover.cpp
siramok: True.

src/libs/rover/rover.hpp
siramok: Exactly right, enables pushing things down a layer.

src/tests/_baseline_images/tout_rover_xray_curv3d_camera_param48.png
siramok: Edited the meta-issue.

src/libs/rover/scheduler.cpp
siramok: Correct, but will require some refactoring since we currently create "state/" in *_rover_filters.cpp. I added a TODO so that this change gets done in a new PR.

JustinPrivitera: That shouldn't be a problem. You can create the state node here just by doing this:
```c++
Node &xray_view = data["state/xray_view"];
```
Then the stuff later will get added to it in the rover filter code.

src/libs/rover/rover.cpp
```diff
@@ -226,6 +226,7 @@ Rover::~Rover()
 void
 Rover::update_settings(Node &params)
 {
+  // TODO: Change this to use has_child everywhere instead
   if (params.has_path("camera/look_at"))
```
JustinPrivitera: if you're curious about examining performance of checking `has_child` versus `has_path`, that's fine and probably useful. But just calling a straight `set` is going to be faster than both of those approaches. We don't need to ask what is in the params, we can just add them directly to `m_settings`.

JustinPrivitera (changes requested):

src/libs/rover/rover.cpp
JustinPrivitera: I'm just suggesting we do these 4 lines in place of the entire function: https://github.com/Alpine-DAV/ascent/pull/1541#discussion_r2117004224

src/libs/rover/rover.cpp
siramok: Your suggestion initially went over my head. This was a wonderful change, thanks for persisting.

siramok: I'll use something else for learning caliper.

src/libs/rover/scheduler.cpp
siramok: Also didn't know that was possible, nice.

src/libs/rover/scheduler.cpp
```diff
@@ -619,10 +619,8 @@ void Scheduler<FloatType>::to_blueprint(Node &data)
 
   const int num_channels = m_result.get_num_channels();
 
-  // TODO: Should live under "state/xray_view", but requires
-  // refactoring since we currently generate "state/" in
-  // *_rover_filters.cpp
-  Node &xray_view = data["xray_view"];
+  // TODO: Plumb the other "state/" info down out of *_rover_filters.cpp
```
JustinPrivitera, line 622: We'll have to see if that's possible. I don't know where in ascent the cycle and time live. It would be nice if all could happen down here.

siramok: Fair enough.

JustinPrivitera (approved):

PR #1542, opened 2025-06-02
Revert "Refactor XRay initialization + camera params"

cyrush:
Reverts Alpine-DAV/ascent#1541


cyrush:
@siramok this will revert, and we will create a new PR with the existing branch.

PR #1544, opened 2025-06-02
Remove Rover's volume renderer

siramok:
Addresses Bullets in #1513
- Comment out all references to rover's volume renderer
    - `#if/#endif` were used to comment large blocks of code, and normal `//` comments were used for 1-3 line removals
    - If desired, the code could be properly removed instead of just commented out (if not in this PR, at a later date since I marked everything with `// removing volume renderer`)
- Minor edits to surrounding code regions
    - Lines with hardcoded usage of the volume renderer code were modified to not depend on the volume renderer
    - Indentation was fixed in cases where conditional blocks were removed

Justification: We want to make substantial changes to rover's internals, which would require updating the `RoverVolume` extract to be compatible with those changes. However, it's unclear if the existing code still works as intended, given that the `RoverVolume` tests have already been commented out for a while (presumably due to a prior breakage). Removing it should make it easier to clean up, reorganize, extend, and eventually maintain the XRay diagnostic portion of rover.


src/libs/rover/domain.cpp
```diff
@@ -10,14 +10,17 @@
...
 #include <utils/rover_logging.hpp>
 
-namespace rover {
+namespace rover
+{
+
 Domain::Domain()
 {
-  m_engine = std::make_shared<VolumeEngine>();
+  m_engine = std::make_shared<EnergyEngine>();
```
siramok, line 18: Just a note: hardcoding VolumeEngine here was the reason that the default render_mode couldn't be changed to energy in rover_types.hpp without causing a crash. It seems that the wrong engine was getting created by default, regardless of render_mode setting, and wasn't properly getting overwritten without explicitly setting render_mode. -1 to the list of mysteries

cyrush: Good find

src/libs/rover/CMakeLists.txt
```diff
@@ -63,11 +63,9 @@ set(rover_headers
       rover.hpp
       scheduler.hpp
       scheduler_base.hpp
-      static_scheduler.hpp
```
JustinPrivitera, line 66: let's delete this guy

src/libs/rover/domain.cpp
JustinPrivitera: Why do we still need EnergyEngine? If Rover always uses Energy mode.

src/libs/rover/domain.cpp
```diff
@@ -97,6 +109,8 @@ Domain::set_engine_fields()
   ROVER_INFO("Primary field: " << m_render_settings.m_primary_field);
   ROVER_INFO("Secondary field: " << m_render_settings.m_secondary_field);
 
+  // TODO: This might be redundant, surely we catch this case in verify_params?
+  // (and should if we currently don't)
   if(m_render_settings.m_primary_field == "")
```
JustinPrivitera, line 114: We want all settings to live in the conduit node - not just camera settings. We can do that in a follow up PR.

src/libs/rover/rover.cpp
```diff
@@ -62,7 +57,7 @@ class Rover::InternalsType
     // logic to create the appropriate parallel scheduler
     //
     // ray tracing = dynamic scheduler, scattering | no_scattering
-    // volume/engery = scattering + local_scope -> dynamic scheduler
+    // engery = scattering + local_scope -> dynamic scheduler
```
```suggestion
    // energy = scattering + local_scope -> dynamic scheduler
```

src/libs/rover/rover_types.hpp
```diff
@@ -58,22 +62,24 @@ struct EnergySettings
...
 {
-  RenderMode     m_render_mode;
+  // RenderMode     m_render_mode; // removing volume renderer
   ScatteringType m_scattering_type;
   RayScope       m_ray_scope;
   vtkmColorTable m_color_table;
   std::string    m_primary_field;
   std::string    m_secondary_field;
-  VolumeSettings m_volume_settings;
+  int            m_num_samples; // approximate number of samples per ray
```
JustinPrivitera, line 71: What is this for? I don't think we are doing multi-sampling.

src/libs/rover/static_scheduler.hpp
JustinPrivitera, line 1: let's just delete this file

src/libs/rover/scheduler.cpp
```diff
@@ -262,65 +269,72 @@ void Scheduler<FloatType>::composite()
...
   {
-    if(m_render_settings.m_secondary_field != "")
-    {
-      vtkh::PartialCompositor<vtkh::EmissionPartial<FloatType>> compositor;
-      compositor.set_background(m_background);
+  }
+#endif
+
+  // TODO: Secondary field should always be set for energy renders,
+  // so we should make it required at verify_params
```
JustinPrivitera: not true - we want to support the case where there is only an absorption specified.

src/libs/rover/scheduler.cpp
```diff
@@ -541,78 +553,81 @@ void Scheduler<FloatType>::save_result(std::string file_name)
...
+Scheduler<FloatType>::save_result(std::string file_name,
+                                  float min_val,
+                                  float max_val,
+                                  bool log_scale)
 {
-  int height = 0;
-  int width = 0;
-  m_ray_generator->get_dims(height, width);
-  assert( height > 0 );
-  assert( width > 0 );
```
JustinPrivitera, line 557: we might want to preserve these asserts?

JustinPrivitera (changes requested):

src/libs/rover/domain.cpp
siramok: At this point it's just vestigial. It could be renamed now, or later as other changes get made.

src/libs/rover/domain.cpp
siramok: Yes, replacing all settings with a conduit node will be my next PR.

src/libs/rover/rover_types.hpp
siramok: The reason it's still here for now is because it gets used later on in the pipeline, although I'm not sure if we ultimately need it.

src/libs/rover/domain.cpp
JustinPrivitera: we will merge engine and energy engine later

siramok: I left a TODO for this to be done later on.

src/libs/rover/scheduler.cpp
siramok: Fair. I went back and added if statements + a proper error message to at least handle this case more nicely.

src/libs/rover/CMakeLists.txt
siramok: Deleted.

src/libs/rover/scheduler.cpp
siramok: Removed this TODO.

JustinPrivitera (approved):

PR #1545, opened 2025-06-03
Replace Rover's setting structs with conduit nodes

siramok:
Addresses Bullets in [#1513](https://github.com/Alpine-DAV/ascent/issues/1513)

- Removes all of Rover's setting structs and enums in favor of conduit nodes
    - The same default values as before are preserved, but the user input params always take precedence
- Rover-specific input params are now scoped with "rover/"
    - All of the rover tests have been adjusted to reflect this change
- Removes InternalsType (was doing strange things, removing it was planned anyways, and doing so enabled a clean fix for the double precision issue) 
- Lots of clean-ups (organization, naming, redundant lines, unused functions, etc)


src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -242,62 +252,33 @@ RoverXRay::execute()
...
+  std::string absorption = params()["rover/absorption"].as_string();
   if(!collection->has_field(absorption))
   {
     ASCENT_ERROR("Absorption field name '" << absorption << "' is not in the dataset");
   }
 
+  // Fetch the dataset associated with the 'absorption' field
   std::string topo_name = collection->field_topology(absorption);
-
-  // Returns an empty dataset if topo_name doesn't exist in the collection
```
siramok, line 253: Note: I decided to remove this clarification, because immediately before this we error if 'absorption' isn't in the dataset. As far as I can tell, we should never end up with an empty dataset by this point.

src/libs/rover/domain.cpp
```diff
@@ -106,16 +105,17 @@ Domain::set_data_set(vtkmDataSet &dataset)
...
+  std::string emission = m_settings["rover/emission"].as_string();
+  std::string color_table_name = m_settings["rover/color_table"].as_string();
+  vtkmColorTable color_table(color_table_name);
+
+  ROVER_INFO("Primary (absorption) field: " << absorbtion);
+  ROVER_INFO("Secondary (emission) field: " << emission);
 
-  // TODO: This might be redundant, surely we catch this case in verify_params?
-  // (and should if we currently don't)
-  if(m_render_settings.m_primary_field == "")
```
siramok, line 114: Note: I added a check to verify_params that ensures 'absorption' is not the empty string, before rover ever executes

src/libs/rover/domain.cpp
```diff
@@ -165,9 +167,6 @@ Domain::set_composite_background(bool on)
 vtkmRange
 Domain::get_primary_range()
 {
-  // TODO: Redundant? Even if it's not, we can probably use something nicer
```
siramok, line 168: Note: I added a check to verify_params that ensures 'absorption' is not the empty string, before rover ever executes

src/libs/rover/rover.cpp
```diff
@@ -206,21 +209,37 @@ class Rover::InternalsType
...
 #ifdef ROVER_ENABLE_LOGGING
-  DataLogger::GetInstance()->WriteLog();
+  DataLogger::get_instance()->WriteLog();
 #endif
 }
 
 void
 Rover::update_settings(Node &params)
 {
+  if (params.has_child("rover"))
```
siramok, line 62: Feedback would be greatly appreciated on whether this was an acceptable way to merge incoming params with already-set default params.

siramok:
I discovered that double precision is currently broken, and that none of our current tests use it. Will want to fix that before merging.

siramok:
I have verified that double precision does work (or at least doesn't throw any error) in develop.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -104,70 +104,79 @@ bool
...
+  if (!params.has_path("rover/absorption") || !params["rover/absorption"].dtype().is_string())
   {
-    info["errors"].append() = "Missing required string parameter 'absorption'";
+    info["errors"].append() = "Missing required string parameter 'rover/absorption'";
     res = false;
   }
 
-  if (!params.has_child("filename") || !params["filename"].dtype().is_string())
+  std::string absorption = params["rover/absorption"].as_string();
+  if (absorption == "")
```
JustinPrivitera: two style/c++ best practices things:
```suggestion
  const std::string absorption = params["rover/absorption"].as_string();
  if ("" == absorption)
```
why?
1. It is best to put `const` everywhere it can possibly go. This reduces possible mistakes. If you don't think a variable is going to change then it is best to make it a constant. If it is known at compile time it is best to make it `constexpr`.
2. Despite it looking a little strange, I prefer to put literal quantities always on the left of a comparison: `(literal_quantity </<=/==/>=/>/!= variable)`. This is safer, because it is valid code to say `if (absorption = "")` by accident. Assignment statements evaluate to the result of the assignment, so saying `if (x = true)` will always evaluate to true, even if `x` was originally false. If we switch the order of the variable and quantity we are comparing, the compiler will save us if we make such a mistake. `if (true = x)` will yield a compiler error.
If you have made `absorption` `const` then you also get around this issue, since an assignment to a constant yields a compiler error. But as a general rule of thumb I try to write my if statements backwards, so to speak, since I am not always comparing `const` variables. I try to leverage the compiler as much as possible.

JustinPrivitera: You don't have to choose to do these things; they are just my suggestions for how I like to write code.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -104,70 +104,79 @@ bool
...
-  bool has_width = params.has_child("width");
-  bool has_height = params.has_child("height");
+  if (params.has_path("rover/emission") && !params["rover/emission"].dtype().is_string())
+  {
+    info["errors"].append() = "Optional parameter 'rover/emission' must be a string";
+    res = false;
+  }
+
+  bool has_width = params.has_path("rover/width");
+  bool has_height = params.has_path("rover/height");
```
JustinPrivitera: so both of these could be `const bool` as well.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -195,30 +204,31 @@ RoverXRay::verify_params(const conduit::Node &params,
...
     {
-      info["errors"].append() = "Optional parameter 'precision' must be a string";
+      info["errors"].append() = "Optional parameter 'rover/precision' must be a string";
+      info["errors"].append() = "Optional parameter 'rover/precision' must be 'single' or 'double'";
       res = false;
     }
     else
     {
-      std::string precision = params["precision"].as_string();
+      std::string precision = params["rover/precision"].to_string();
```
JustinPrivitera:
```suggestion
      const std::string precision = params["rover/precision"].to_string();
```

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -195,30 +204,31 @@ RoverXRay::verify_params(const conduit::Node &params,
...
         res = false;
       }
     }
   }
 
-  if (params.has_child("blueprint"))
+  if (params.has_path("rover/blueprint"))
   {
-    std::string protocol = params["blueprint"].as_string();
+    std::string protocol = params["rover/blueprint"].as_string();
```
JustinPrivitera:
```suggestion
    const std::string protocol = params["rover/blueprint"].as_string();
```

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
JustinPrivitera: The clarification comment you removed is about the topology, not the absorption field. I'm confused.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -242,62 +252,33 @@ RoverXRay::execute()
   }
   std::shared_ptr<VTKHCollection> collection = data_object->as_vtkh_collection();
 
-  std::string absorption = params()["absorption"].as_string();
+  // Validate that the 'absorption' field exists in the dataset
+  std::string absorption = params()["rover/absorption"].as_string();
```
JustinPrivitera:
```suggestion
  const std::string absorption = params()["rover/absorption"].as_string();
```

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -242,62 +252,33 @@ RoverXRay::execute()
...
   Rover rover;
   rover.update_settings(params());
 
   int mpi_comm_id = -1;
 #ifdef ASCENT_MPI_ENABLED
   mpi_comm_id = flow::Workspace::default_mpi_comm();
   rover::Logger::get_instance()->set_mpi_comm_id(mpi_comm_id);
-  // these use different styles of naming functions ....
-  rover::DataLogger::GetInstance()->set_mpi_comm_id(mpi_comm_id);
+  rover::DataLogger::get_instance()->set_mpi_comm_id(mpi_comm_id);
```
JustinPrivitera: With this change we are still mixing camelcase and pot_hole_case. I don't know that it matters all that much, as long as we try to keep consistency with new things that we add.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -242,62 +252,33 @@ RoverXRay::execute()
...
-    render_settings.m_secondary_field = params()["emission"].as_string();
-  }
-
-  if(params().has_path("unit_scalar"))
-  {
-    render_settings.m_energy_settings.m_unit_scalar = params()["unit_scalar"].to_float64();
-  }
-
-  rover.set_render_settings(render_settings);
+  // Calling execute initializes everything that rover needs based on the input params
```
JustinPrivitera: I guess `rover.add_data_set()` now handles all of the stuff that got removed here? It updates and camera and sets up render settings? I ask because we don't want any setup stuff happening in `execute()`.

src/libs/rover/CMakeLists.txt
```diff
@@ -70,7 +69,6 @@ set(rover_headers
       ray_generators/ray_generator.hpp
       ray_generators/camera_generator.hpp
       ray_generators/visit_generator.hpp
-      rover_exports.h
```
JustinPrivitera: Did this file get deleted too? I don't see it in the github diff

siramok: Good catch, this was unintentional.

src/libs/rover/domain.cpp
```diff
@@ -49,9 +51,9 @@ Domain::set_render_settings(const RenderSettings &settings)
   {
 #endif
 
-  ROVER_INFO("Render mode = energy");
   auto engine = std::make_shared<EnergyEngine>();
-  engine->set_unit_scalar(settings.m_energy_settings.m_unit_scalar);
+  float64 unit_scalar = m_settings["rover/unit_scalar"].value();
+  engine->set_unit_scalar(unit_scalar);
```
JustinPrivitera:
```suggestion
  engine->set_unit_scalar(m_settings["rover/unit_scalar"].value());
```

JustinPrivitera: we can cut out the middleman unless we need `unit_scalar` for anything later.

JustinPrivitera: also, set_unit_scalar takes a float32: https://github.com/Alpine-DAV/ascent/blob/ecd571649d5b5e320d8270cb850680167ec6a7cd/src/libs/rover/energy_engine.hpp#L41
So we are potentially doing a bunch of unnecessary conversions.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
siramok: It's definitely not an important change; it was done purely to justify removing the comment. We can do as you wish with this.

src/libs/rover/domain.cpp
JustinPrivitera: I have another question - why does energy engine need to own a unit scalar attribute? Is it possible for it to have access to the settings node at the time when it actually needs to use unit scalar?

src/libs/rover/domain.cpp
```diff
@@ -106,16 +105,17 @@ Domain::set_data_set(vtkmDataSet &dataset)
 void
 Domain::set_engine_fields()
 {
-  ROVER_INFO("Primary field: " << m_render_settings.m_primary_field);
-  ROVER_INFO("Secondary field: " << m_render_settings.m_secondary_field);
+  std::string absorption = m_settings["rover/absorption"].as_string();
+  std::string emission = m_settings["rover/emission"].as_string();
+  std::string color_table_name = m_settings["rover/color_table"].as_string();
```
JustinPrivitera:
```suggestion
  const std::string absorption = m_settings["rover/absorption"].as_string();
  const std::string emission = m_settings["rover/emission"].as_string();
  const std::string color_table_name = m_settings["rover/color_table"].as_string();
```

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
siramok: Thanks for the explanation, I am convinced.

src/libs/rover/domain.hpp
```diff
@@ -40,7 +42,7 @@ class Domain
   vtkmDataSet             m_data_set;
   vtkm::Bounds            m_global_bounds;
   vtkm::Bounds            m_domain_bounds;
-  RenderSettings          m_render_settings;
+  Node                    m_settings;
```
JustinPrivitera: We should discuss domain.hpp. Does it actually represent per-domain info? I am wondering when we would ever have per-domain settings? Aren't these settings the same across all domains?

src/libs/rover/image.hpp
```diff
@@ -10,8 +10,7 @@
 #include <rover_config.h>
 #include <vector>
 #include <vtkm/cont/ArrayHandle.h>
-
-#include <rover_types.hpp>
+#include <vtkm_typedefs.hpp>
```
JustinPrivitera, line 13: what did we need this for?

src/libs/rover/rover.cpp
JustinPrivitera: I like it

src/libs/rover/rover.cpp
```diff
@@ -233,10 +252,46 @@ Rover::print_settings()
...
+  return -1;
+#endif
+}
+
+void
+Rover::add_data_set(vtkh::DataSet &dataset)
+{
+  for (int i = 0; i < dataset.GetNumberOfDomains(); i++)
+  {
+    m_internals->add_data_set(dataset.GetDomain(i));
```
JustinPrivitera: I am unsure about the future of m_internals

src/libs/rover/rover.cpp
```diff
@@ -319,46 +372,24 @@ Rover::update_camera_generator()
...
-Rover::set_mpi_comm_handle(int mpi_comm_id)
-{
-#ifdef ROVER_PARALLEL
-  this->m_internals->set_comm_handle(MPI_Comm_f2c(mpi_comm_id));
-#else
-  (void)mpi_comm_id;
-#endif
-}
+  m_ray_generator.set_width(width);
+  m_ray_generator.set_height(height);
```
JustinPrivitera, line 238: Do these need the width and height? Or can they get settings when they need them from the global m_settings?

src/libs/rover/rover.cpp
```diff
@@ -319,46 +372,24 @@ Rover::update_camera_generator()
...
-#endif
+  m_internals->set_ray_generator(&m_ray_generator);
 }
 
 void
-Rover::add_data_set(vtkh::DataSet &dataset)
+Rover::update_precision()
 {
-  for (int i = 0; i < dataset.GetNumberOfDomains(); i++)
+  std::string precision = m_settings["rover/precision"].as_string();
```
```suggestion
  const std::string precision = m_settings["rover/precision"].as_string();
```

src/libs/rover/rover.hpp
```diff
@@ -58,17 +57,15 @@ class ROVER_API Rover
...
-  void set_tracer_precision32();
-  void set_tracer_precision64();
   void get_result(Image<vtkm::Float32> &image);
   void get_result(Image<vtkm::Float64> &image);
-  void update_camera();
-  void update_camera_generator();
 private:
+  Node m_settings;
+  vtkmCamera m_camera;
+  CameraGenerator m_ray_generator;
```
JustinPrivitera, line 72: 😡

src/libs/rover/scheduler.cpp
```diff
@@ -272,7 +272,8 @@ Scheduler<FloatType>::composite()
   }
 #endif
 
-  if (m_render_settings.m_secondary_field != "")
+  std::string emission = m_settings["rover/emission"].as_string();
```
JustinPrivitera:
```suggestion
  const std::string emission = m_settings["rover/emission"].as_string();
```

src/libs/rover/scheduler.cpp
```diff
@@ -461,14 +462,16 @@ Scheduler<FloatType>::trace_rays()
     partial_image.m_height = height;
     partial_image.m_buffer =
       vtkm::rendering::raytracing::ChannelBuffer<FloatType>(num_channels, 0);
-    if(m_render_settings.m_secondary_field != "")
+
+    std::string emission = m_settings["rover/emission"].as_string();
```
JustinPrivitera:
```suggestion
    const std::string emission = m_settings["rover/emission"].as_string();
```

src/libs/rover/rover.cpp
siramok: Same.

src/libs/rover/domain.hpp
siramok: Not sure, another mystery for the pile.

src/libs/rover/scheduler_base.cpp
```diff
@@ -18,13 +18,10 @@ SchedulerBase::~SchedulerBase()
...
 void
-SchedulerBase::set_render_settings(const RenderSettings render_settings)
+SchedulerBase::set_settings(const Node &settings)
 {
-  //
-  //  In the serial schedular, the only setting that matter are
-  //  m_render_mode and m_scattering_mode
-  //
-  m_render_settings = render_settings;
+  // In the serial scheduler, the only setting that matters is m_scattering_mode
```
JustinPrivitera: this isn't true

src/libs/rover/scheduler_base.cpp
```diff
@@ -18,13 +18,10 @@ SchedulerBase::~SchedulerBase()
...
+  m_settings = settings;
```
JustinPrivitera, line 23: it's m_settings all the way down

src/libs/rover/utils/rover_logging.cpp
```diff
@@ -102,11 +102,11 @@ DataLogger::get_mpi_comm_id()
 }
 
 DataLogger*
-DataLogger::GetInstance()
+DataLogger::get_instance()
```
JustinPrivitera: return to camelcase

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -242,62 +252,35 @@ RoverXRay::execute()
...
   {
     ASCENT_ERROR("Absorption field name '" << absorption << "' is not in the dataset");
   }
 
+  // Fetch the dataset associated with the 'absorption' field
   std::string topo_name = collection->field_topology(absorption);
-
-  // Returns an empty dataset if topo_name doesn't exist in the collection
+  // TODO: Validate topo_name
   vtkh::DataSet &dataset = collection->dataset_by_topology(topo_name);
```
JustinPrivitera, line 265: Can we also add a note to validate the `dataset`? We should figure out what `dataset_by_topology` returns if it gets a bad topo name. I would recommend reading docs or finding the source code.

src/libs/rover/domain.cpp
JustinPrivitera: ^^^

src/libs/rover/rover.cpp
```diff
@@ -16,227 +16,140 @@
...
+}
 #endif
-    m_scheduler->save_bov(file_name);
-  }
 
-  void execute()
+void
+Rover::create_scheduler()
+{
+  std::string precision = m_settings["rover/precision"].as_string();
```
```suggestion
  const std::string precision = m_settings["rover/precision"].as_string();
```

src/libs/rover/rover.cpp
```diff
@@ -16,227 +16,140 @@
...
+}
 
-  void set_tracer_precision32()
+void
+Rover::add_data_set(vtkh::DataSet &dataset)
+{
+  ROVER_INFO("Executing Rover::add_data_set");
+  // The scheduler needs to be created before data can be added
+  // to it, else we segfault
+  if (m_scheduler == nullptr)
```
```suggestion
  if (nullptr == m_scheduler)
```

JustinPrivitera: or you could do `if (! m_scheduler)`

src/libs/rover/rover.cpp
```diff
@@ -297,14 +210,12 @@ Rover::update_camera()
...
 
 void
-Rover::update_camera_generator()
+Rover::update_ray_generator()
 {
-  m_camera_generator.set_camera(m_camera);
+  m_ray_generator.set_camera(m_camera);
 
   int64 width = 200;
   int64 height = 200;
```
JustinPrivitera: I don't love that these are hardcoded. We can guarantee that width and height are always in the settings, right? So we should just read directly from there.

src/libs/rover/rover.cpp
```diff
@@ -448,26 +298,38 @@ Rover::about()
...
 void Rover::to_blueprint(conduit::Node &dataset)
 {
-  m_internals->to_blueprint(dataset);
+#ifdef ROVER_PARALLEL
+    if(m_rank != 0)
+    {
+      return;
+    }
+#endif
+    m_scheduler->to_blueprint(dataset);
```
JustinPrivitera: Can we ensure that all of the data has been collected on the root rank? I can imagine a world where we would want to write in parallel.

src/libs/rover/rover.cpp
```diff
@@ -448,26 +298,38 @@ Rover::about()
...
 Rover::save_png(const std::string &file_name)
 {
-  m_internals->save_png(file_name);
+#ifdef ROVER_PARALLEL
+    if(m_rank != 0)
+    {
+      return;
+    }
+#endif
+  m_scheduler->save_result(file_name);
```
JustinPrivitera: same thought here

src/libs/rover/rover.cpp
```diff
@@ -476,38 +338,31 @@ Rover::save_png(const std::string &file_name,
...
                 const bool log_scale)
 {
-  m_internals->save_png(file_name, min_val, max_val, log_scale);
+#ifdef ROVER_PARALLEL
+  if(m_rank != 0)
+  {
+    return;
+  }
+#endif
+  m_scheduler->save_result(file_name, min_val, max_val, log_scale);
```
JustinPrivitera: and here

src/libs/rover/rover.cpp
```diff
@@ -16,227 +16,140 @@
...
-  }
-  void save_png(const std::string &file_name,
-                const float min_val,
-                const float max_val,
-                const bool log_scale)
+void
+Rover::set_mpi_comm_handle(MPI_Comm comm_handle)
+{
+  m_comm_handle = MPI_Comm_f2c(comm_handle);
+  MPI_Comm_rank(m_comm_handle, &m_rank);
```
JustinPrivitera: Why did this change and what is going on here? We pass in a comm_handle by value only to blow that value away? I am confused where this came from.

src/libs/rover/rover.hpp
```diff
@@ -31,25 +31,26 @@ namespace rover
...
-  
   Rover();
   ~Rover();
 
   void update_settings(Node &params);
   void print_settings();
 
-  void set_mpi_comm_handle(int mpi_comm_id);
+  #ifdef ROVER_PARALLEL
+  void set_mpi_comm_handle(MPI_Comm comm_handle);
```
JustinPrivitera: confused why the input arg changed

src/libs/rover/scheduler_base.cpp
```diff
@@ -52,12 +48,6 @@ SchedulerBase::set_background(const std::vector<vtkm::Float32> &background)
 
 }
 
-void
-SchedulerBase::clear_data_sets()
-{
-  m_domains.clear();
-}
-
```
JustinPrivitera, line 60: Why don't we need this anymore? Is there another place that frees the domains?

JustinPrivitera (changes requested):

src/libs/rover/domain.cpp
siramok: 1) I took your suggested float32 change. 2) I agree with you, so I added a TODO for getting rid of set_unit_scalar in favor of just querying m_settings directly once it's needed.

src/libs/rover/rover.cpp
siramok: This suggestion has been implemented.

src/libs/rover/scheduler.cpp
siramok: This suggestion has been implemented.

src/libs/rover/scheduler.cpp
siramok: This suggestion has been implemented.

src/libs/rover/scheduler_base.cpp
siramok: Hopefully not forever.

src/libs/rover/utils/rover_logging.cpp
siramok: This suggestion has been implemented.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
siramok: A TODO has now been added for this.

src/libs/rover/rover.cpp
siramok: This suggestion has been implemented.

src/libs/rover/rover.cpp
siramok: I went with `!m_scheduler`.

src/libs/rover/rover.cpp
siramok: Fixed.

src/libs/rover/rover.cpp
siramok: I added a TODO to look into support for parallel writes.

src/libs/rover/rover.cpp
siramok: I added a TODO to look into support for parallel writes.

src/libs/rover/rover.cpp
siramok: I added a TODO to look into support for parallel writes.

src/libs/rover/rover.cpp
JustinPrivitera: I'm just wondering why you added this logic in

JustinPrivitera: it looks like you added it from the diff

src/libs/rover/scheduler_base.cpp
siramok: I think this function is vestigial from rover's initial vision as a standalone library. My best guess is that the idea was to reuse a rover instance by clearing out data to add new data, but I removed it because it's never used anywhere. What do you think?

src/libs/rover/rover.cpp
siramok: It has been disappeared.

src/libs/rover/scheduler_base.cpp
siramok: Removed.

src/libs/rover/rover.hpp
siramok: This change was done to reflect what was happening within `InternalsType`, since I ended up removing `InternalsType`. So rover.hpp was previously taking an `int` here but was then converting it to an `MPI_Comm` before calling the `m_internals` analog. Technically we can keep this as an int, and do the conversion to `MPI_Comm` within `set_mpi_comm_handle`.

src/libs/rover/rover.cpp
siramok: I added a TODO to look into support for parallel writes.

src/libs/rover/scheduler_base.cpp
siramok: I have made a TODO for either making m_settings globally available within the rover namespace or changing the functions that use m_settings values to receive those via function arguments. I'm leaning towards making it globally available within the rover namespace.

src/libs/rover/scheduler_base.cpp
JustinPrivitera: is there anther place where we free domain data? Is there anything that needs to be freed or will things just go out of scope nicely?

src/libs/rover/rover.hpp
JustinPrivitera: We'll have to think about these things more later. This is fine for now.

JustinPrivitera (approved):
Thanks for all the iteration on this. Changes are solid and we are making good progress. 💪

src/libs/rover/rover.cpp
siramok: It came from m_internals, just pulling out the existing functionality so that we know nothing was lost in removing `InternalsType`.

src/libs/rover/scheduler_base.cpp
siramok: I'm not aware of another place where the domain data is explicitly freed. I took a quick peek at `vtk::DataSet` and there's an `std::vector<vtkm::cont::DataSet>`. My expectation would be that it gets cleaned up nicely upon going out of scope.

JustinPrivitera: Gotcha. I haven't stared at the code enough to know if it involved pointers and `new` somehow. Thanks for thinking about this.

src/libs/rover/rover.cpp
JustinPrivitera: Gotcha, the diff is confusing. Thanks for clarifying!

PR #1548, opened 2025-06-06
Change rover's default background intensity to 0.0f

siramok:
Addresses Bullets in #1513

- VisIt's XRay query has a default background intensity value of 0.0f, and now rover has parity.


src/tests/_baseline_images/tout_rover_xray_curv3d_camera_param48.png
JustinPrivitera, line 1: What happened with this one? Why is it chopped?

JustinPrivitera (changes requested):
All looks good except for the one strange baseline

JustinPrivitera (approved):

src/tests/_baseline_images/tout_rover_xray_curv3d_camera_param48.png
siramok: Thanks for being attentive to detail. It's caused by setting xpan/ypan, but I'm not sure why at this point. Not setting them produces the same baseline image, just unclipped.

PR #1549, opened 2025-06-06
Use a global settings node for rover

siramok:
Addresses Bullets in #1513

- Rather than propagate references to rover's settings to all of the subsystems, we move the settings node into rover's namespace so that the subsystems can access it directly.


src/libs/rover/rover.cpp
```diff
@@ -19,18 +19,23 @@
 namespace rover
 {
 
+// Namespaced settings instance
+Node settings;
```
siramok, line 24: I didn't intend to leave this in after creating settings.cpp, however removing it breaks everything. I'll have to take a peek at what's going on. Maybe I just get rid of settings.cpp and call it good?

src/libs/rover/settings.hpp
```diff
@@ -0,0 +1,19 @@
...
+#ifndef ROVER_SETTINGS_H
+#define ROVER_SETTINGS_H
+
+#include <conduit.hpp>
+
+using namespace conduit;
+
+namespace rover
+{
+  extern Node settings;
```
siramok, line 16: This could also live in rover.hpp, however then we'll have to `#include <rover.hpp>` everywhere and I think `#include <settings.hpp>` is potentially clearer.

src/libs/rover/rover.cpp
JustinPrivitera: I think it's fine to get rid of settings.cpp and have it live here. I'd be curious to see what the compiler is angry about. You might need CMake lists file changes since you are adding new source files, and that could be your issue.

JustinPrivitera (approved):
Let's blow away settings.cpp and also make sure the new header is added to the CMake files wherever it makes sense to do so.

JustinPrivitera (changes requested):
Meant to click request changes

src/libs/rover/settings.hpp
JustinPrivitera: I think it is good in the rover cpp file.

JustinPrivitera (approved):

src/libs/rover/rover.cpp
siramok: I removed settings.cpp. The compiler error was about a missing reference to rover::settings, which honestly may have been related to not including it in the CMakeLists.txt.

src/libs/rover/settings.hpp
siramok: Great, settings.hpp has the declaration while the instantiation happens in rover.cpp.

src/libs/rover/rover.cpp
JustinPrivitera: Thanks. Yes, I imagine that was the source of the error.

PR #1551, opened 2025-06-09
Rover width, height cleanup + adds settings used to the output

siramok:
Addresses bullets in https://github.com/Alpine-DAV/ascent/issues/1513

- Removes instances of passing/storing width and height, in favor of querying the global settings node wherever we need those values.
- Dumps all of the options used by Rover into the output node under "state/xray_query".

### DIFF
--- a/src/libs/rover/image.cpp
+++ b/src/libs/rover/image.cpp
@@ -46,7 +46,9 @@ Image<FloatType>::normalize_handle(vtkm::cont::ArrayHandle<FloatType> &handle,
   FloatType inv_delta;
   inv_delta = min_scalar == max_scalar ? 1.f : 1.f / (max_scalar - min_scalar);
   auto portal = handle.WritePortal();
-  const int size = m_width * m_height;
+  const int32 width = rover::settings["rover/width"].value();
+  const int32 height = rover::settings["rover/height"].value();
+  const int size = width * height;
 #ifdef ROVER_OPENMP_ENABLED
   #pragma omp parallel for
 #endif
@@ -79,7 +81,9 @@ Image<FloatType>::normalize_handle(vtkm::cont::ArrayHandle<FloatType> &handle, b
   FloatType inv_delta;
   inv_delta = min_scalar == max_scalar ? 1.f : 1.f / (max_scalar - min_scalar);
   auto portal = handle.WritePortal();
-  const int size = m_width * m_height;
+  const int32 width = rover::settings["rover/width"].value();
+  const int32 height = rover::settings["rover/height"].value();
+  const int size = width * height;
 #ifdef ROVER_OPENMP_ENABLED
   #pragma omp parallel for
 #endif
@@ -94,8 +98,6 @@ Image<FloatType>::normalize_handle(vtkm::cont::ArrayHandle<FloatType> &handle, b
 
 template<typename FloatType>
 Image<FloatType>::Image()
-  : m_width(0),
-    m_height(0)
 {
 
 }
@@ -134,8 +136,6 @@ void cast_array_handle(vtkm::cont::ArrayHandle<T> &cast_to,
 //
 template<typename T, typename O> void init_from_image(Image<T> &left, Image<O> &right)
 {
-  left.m_height = right.m_height;
-  left.m_width = right.m_width;
   left.m_valid_intensities = right.m_valid_intensities;
   left.m_valid_optical_depths = right.m_valid_optical_depths;
 
@@ -150,8 +150,6 @@ template<typename T, typename O> void init_from_image(Image<T> &left, Image<O> &
 template<> void init_from_image<vtkm::Float32, vtkm::Float32>(Image<vtkm::Float32> &left,
                                                               Image<vtkm::Float32> &right)
 {
-  left.m_height = right.m_height;;
-  left.m_width = right.m_width;
   left.m_intensities = right.m_intensities;
   left.m_optical_depths = right.m_optical_depths;
   left.m_valid_intensities = right.m_valid_intensities;
@@ -161,8 +159,6 @@ template<> void init_from_image<vtkm::Float32, vtkm::Float32>(Image<vtkm::Float3
 template<> void init_from_image<vtkm::Float64, vtkm::Float64>(Image<vtkm::Float64> &left,
                                                               Image<vtkm::Float64> &right)
 {
-  left.m_height = right.m_height;;
-  left.m_width = right.m_width;
   left.m_intensities = right.m_intensities;
   left.m_optical_depths = right.m_optical_depths;
   left.m_valid_intensities = right.m_valid_intensities;
@@ -227,18 +223,15 @@ Image<FloatType>::init_from_partial(PartialImage<FloatType> &partial)
   m_valid_intensities.clear();
   m_valid_optical_depths.clear();
 
-  m_height = partial.m_height;
-  m_width  = partial.m_width;
-
-  assert(m_width >= 0);
-  assert(m_height >= 0);
+  const int32 width = rover::settings["rover/width"].value();
+  const int32 height = rover::settings["rover/height"].value();
 
   const int num_channels = partial.m_buffer.GetNumChannels();
   for(int i = 0; i < num_channels; ++i)
   {
     vtkmRayTracing::ChannelBuffer<FloatType> channel = partial.m_buffer.GetChannel( i );
     const FloatType default_value = partial.m_source_sig.size() != 0 ? partial.m_source_sig[i] : 0.0f;
-    const int channel_size = m_height * m_width;
+    const int channel_size = height * width;
     vtkmRayTracing::ChannelBuffer<FloatType>  expand;
     expand = channel.ExpandBuffer(partial.m_pixel_ids,
                                   channel_size,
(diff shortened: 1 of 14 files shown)


src/libs/rover/scheduler.cpp
```diff
@@ -679,16 +635,21 @@ Scheduler<FloatType>::to_blueprint(Node &data)
...
   xray_view["ypan"] = xy_pan[1];
 
-  Node &n_coords = data["coordsets/" + coord_name];
-  n_coords["type"] = "rectilinear";
+  Node &xray_query = data["xray_query"];
+  xray_query.set(rover::settings["rover"]);
+
+  const int num_channels = m_result.get_num_channels();
+  const std::string coord_name = "image_coords";
+  Node &coordsets = data["coordsets"][coord_name];
```
JustinPrivitera: I don't think `coordsets` is a good name, because `data["coordsets"][coord_name];` is a single coordinate set, not multiple. `n_coords` was good and is stylistically what we use all over the place.

src/libs/rover/scheduler.cpp
```diff
@@ -648,22 +616,10 @@ template<typename FloatType>
...
 Scheduler<FloatType>::to_blueprint(Node &data)
 {
-  int width;
-  int height;
-  m_ray_generator->get_dims(width, height);
-
-  if (width <= 0 || height <= 0)
-  {
-    ROVER_ERROR("Error: to_blueprint failed due to non-positive output dimensions: " << width << "x" << height );
-  }
```
JustinPrivitera, line 658: Do we catch this somewhere else now?

src/libs/rover/scheduler.cpp
```diff
@@ -359,14 +354,8 @@ Scheduler<FloatType>::trace_rays()
...
-  int width;
-  int height;
-  m_ray_generator->get_dims(width, height);
-
-  if (width <= 0 || height <= 0)
-  {
-    ROVER_ERROR("Error: trace_rays failed due to non-positive output dimensions: " << width << "x" << height );
-  }
+  const int32 width = rover::settings["rover/width"].value();
+  const int32 height = rover::settings["rover/height"].value();
```
JustinPrivitera: what are these needed for?

src/libs/rover/scheduler.cpp
```diff
@@ -272,6 +269,8 @@ Scheduler<FloatType>::composite()
   }
 #endif
 
+  const int32 width = rover::settings["rover/width"].value();
+  const int32 height = rover::settings["rover/height"].value();
```
JustinPrivitera: What are these needed for?

src/libs/rover/ray_generators/visit_generator.cpp
```diff
@@ -103,7 +103,7 @@ VisitGenerator::gen_rays(vtkmRayTracing::Ray<T> &rays)
 
   view_height = m_params.m_parallel_scale;
   // I think this is flipped
-  view_width = view_height * (m_height / m_width);
+  view_width = view_height * (height / width);
```
JustinPrivitera: This is bugged - not your fault. These need to get casted to floats.

Otherwise if `height` is `20` and `width` is `17`, then `20 / 17` is `1`, which is wrong.

src/libs/rover/ray_generators/camera_generator.cpp
```diff
@@ -65,25 +66,35 @@ CameraGenerator::~CameraGenerator()
 void
 CameraGenerator::get_rays(vtkmRayTracing::Ray<vtkm::Float32> &rays)
 {
-  vtkm::rendering::CanvasRayTracer canvas(m_width, m_height);
-  vtkm::rendering::raytracing::Camera ray_gen;
-  ray_gen.SetParameters(m_camera, m_width, m_height);
+  const int32 width = rover::settings["rover/width"].value();
+  const int32 height = rover::settings["rover/height"].value();
+  // vtkm::rendering::CanvasRayTracer canvas(width, height);
```
JustinPrivitera: Why was this commented out? Is it unused?

src/libs/rover/image.cpp
```diff
@@ -46,7 +46,9 @@ Image<FloatType>::normalize_handle(vtkm::cont::ArrayHandle<FloatType> &handle,
   FloatType inv_delta;
   inv_delta = min_scalar == max_scalar ? 1.f : 1.f / (max_scalar - min_scalar);
   auto portal = handle.WritePortal();
-  const int size = m_width * m_height;
+  const int32 width = rover::settings["rover/width"].value();
+  const int32 height = rover::settings["rover/height"].value();
+  const int size = width * height;
```
JustinPrivitera, line 52: I don't know what the answer is because I haven't stared intensely at the code for how width and height are being used. We do want them out of places to reduce redundancy, but if there is a place that is using them repeatedly, it might make more sense to leave them in as member variables so that we're not extracting them from a conduit node a huge number of times. I trust your discretion on when we should rely on the settings node and when we should use member variables. I just want to make sure we were not overzealous.

JustinPrivitera (changes requested):

src/libs/rover/scheduler.cpp
siramok: Sounds good, reverted.

src/libs/rover/scheduler.cpp
siramok: I don't think either variable can ever be `<= 0` by this point, but the checks can come back if we're not convinced:

1. A default value of 200 is set in rover's constructor for both width and height settings.
2. This can be overwritten by user input, however, verify_params checks that both width and height are both `> 0` and fails if that's not true.

src/libs/rover/scheduler.cpp
siramok: Seem to not be needed, removed.

src/libs/rover/scheduler.cpp
siramok: Seem to not be needed, removed.

src/libs/rover/ray_generators/visit_generator.cpp
siramok: True, so would you fix this with:
```
view_width = view_height * static_cast<float>(height) / static_cast<float>(width);
```
?

src/libs/rover/ray_generators/camera_generator.cpp
siramok: Yes, I noticed that it doesn't actually get used, so I commented it out to confirm.

src/libs/rover/scheduler.cpp
JustinPrivitera: Ok, it sounds like verify params handles this case for us then.

src/libs/rover/image.cpp
siramok: I think you make a good point. Maybe a hybrid approach would be best: for cases where settings are needed more than once within any given class, query the settings node once in the constructor and then reuse the values.

src/libs/rover/ray_generators/visit_generator.cpp
JustinPrivitera: Yup - that's how we did it in VisIt: https://github.com/visit-dav/visit/blob/7b906cfed048d59ff9411daa1b9afb9de8f12408/src/avt/Filters/avtXRayFilter.C#L2323

src/libs/rover/ray_generators/camera_generator.cpp
JustinPrivitera: If everything works fine with it commented, then we can delete. I think it was commented out somewhere else in this file too, so the same is true for that case as well.

src/libs/rover/image.cpp
JustinPrivitera: I think we can be a little less conservative than that. If something is used in a loop over all the pixels/rays, or maybe even over all the energy groups, then we probably want to reuse it. Otherwise, if it's just used a couple times not in some loop, then it is fine to grab from settings.

siramok: I attempted to address this in a sensible way, I'm curious how you feel about the changes. Some places now query the settings inline (e.g. camera_generator.cpp), and other places that reuse the values several times now just query the settings once within a constructor and store them (e.g. image.cpp).

src/libs/rover/ray_generators/visit_generator.cpp
siramok: Fixed.

src/libs/rover/ray_generators/camera_generator.cpp
siramok: Deleted.

src/libs/rover/image.cpp
```diff
@@ -46,9 +47,7 @@ Image<FloatType>::normalize_handle(vtkm::cont::ArrayHandle<FloatType> &handle,
   FloatType inv_delta;
   inv_delta = min_scalar == max_scalar ? 1.f : 1.f / (max_scalar - min_scalar);
   auto portal = handle.WritePortal();
-  const int32 width = rover::settings["rover/width"].value();
-  const int32 height = rover::settings["rover/height"].value();
-  const int size = width * height;
+  const int size = m_width * m_height;
```
JustinPrivitera: let's discuss this afternoon what the best thing to do is

src/libs/rover/image.cpp
JustinPrivitera: let's go through them and discuss in a few minutes

siramok: For posterity, we settled on reverting the hybrid approach and querying the settings node in every instance that width and height are needed. With the understanding that our approach can be changed later on if we find a good reason to do so.

src/libs/rover/partial_image.hpp
```diff
@@ -7,7 +7,10 @@
 #ifndef rover_partial_image_h
 #define rover_partial_image_h
 
+#include <cstdint>
```
JustinPrivitera: why did we need this?

JustinPrivitera (commented):
One question

src/libs/rover/partial_image.hpp
siramok: Great question, I have no recollection of adding this myself, and the code builds without it. Removed.

JustinPrivitera (approved):

PR #1553, opened 2025-06-09
Simplify Rover's engine code

siramok:
Addresses bullets in #1513

- Since we no longer have the volume rendering engine, this PR simplifies the remaining engine code. Previously `Engine` was a base class that would be extended from, but for the time being we're only interested in supporting a single engine.


src/libs/rover/domain.cpp
```diff
@@ -136,16 +134,12 @@ Domain::init_rays(Ray64 &rays)
 PartialVector32
 Domain::partial_trace(Ray32 &rays)
 {
-  int32 num_samples = rover::settings["rover/num_samples"].value();
-  m_engine->set_samples(m_global_bounds, num_samples);
```
siramok, line 140: The reason I removed this is because, as far as I could tell, `->set_samples` didn't actually set anything.

src/libs/rover/domain.cpp
```diff
@@ -136,16 +134,12 @@ Domain::init_rays(Ray64 &rays)
...
   return m_engine->partial_trace(rays);
 }
 
 PartialVector64
 Domain::partial_trace(Ray64 &rays)
 {
-  int32 num_samples = rover::settings["rover/num_samples"].value();
-  m_engine->set_samples(m_global_bounds, num_samples);
```
siramok, line 148: The reason I removed this is because, as far as I could tell, `->set_samples` didn't actually set anything.

src/libs/rover/domain.cpp
```diff
@@ -49,8 +48,7 @@ Domain::init()
   {
 #endif
 
-  auto engine = std::make_shared<EnergyEngine>();
-  engine->set_unit_scalar(rover::settings["rover/unit_scalar"].value());
+  auto engine = std::make_shared<Engine>();
   m_engine = engine;
```
JustinPrivitera: why not just a one-liner to set the m_engine?

src/libs/rover/domain.cpp
```diff
@@ -110,9 +108,9 @@ Domain::set_engine_fields()
   ROVER_INFO("Primary (absorption) field: " << absorbtion);
   ROVER_INFO("Secondary (emission) field: " << emission);
 
-  m_engine->set_primary_field(absorption);
-  m_engine->set_secondary_field(emission);
-  m_engine->set_color_table(color_table);
+  m_engine->set_primary_field();
+  m_engine->set_secondary_field();
+  m_engine->set_color_map(color_table);
```
JustinPrivitera: what if we combine these methods into an `init` method that does them all?

src/libs/rover/domain.cpp
JustinPrivitera: I think it was an idea that was not fully fleshed out. Fine to remove.

JustinPrivitera (changes requested):

JustinPrivitera:
This is solid; just a few minor nitpicks.

src/libs/rover/domain.cpp
siramok: We already do this in Domain's constructor, so I wondered if these lines were even necessary. I removed them and verified that things didn't break, so that's what I pushed.

src/libs/rover/domain.cpp
siramok: Done.

JustinPrivitera (approved):

PR #1554, opened 2025-06-10
Simplify Rover's scheduler code

siramok:
Addresses bullets in #1513

- SchedulerBase has been renamed to Scheduler and is now purely a virtual class that exists for type erasure purposes.
- Scheduler has been renamed to SchedulerImpl (we can call it something else) and now owns all of the implementation.
- Removes unused functions.


src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -300,32 +300,33 @@ RoverXRay::execute()
...
 
-      if(time != -1)
+      if (time != -1)
       {
         data["state/time"] = time;
       }
     }
 
+    // TODO: Why is this hardcoded to -1?
     const int num_files = -1;
```
JustinPrivitera: I wonder where it gets used. This is useless for the Blueprint output unless we think of a way to decompose the outputs.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -339,30 +340,24 @@ RoverXRay::execute()
...
-  {
-    rover.save_png(filename);
-  }
+  rover.save_png(filename);
 
-  if (params().has_path("bov_filename"))
+  // TODO: Can't we just use rover/filename? Why would we want one
+  // filename for a .png and another filename for a .bov? We don't
+  // currently check if rover/bov_filename is valid in verify_params.
+  if (params().has_path("rover/bov_filename"))
```
JustinPrivitera: We might want both png outputs and bov outputs and we might want them to have different names.

src/libs/rover/CMakeLists.txt
```diff
@@ -61,7 +61,7 @@ set(rover_headers
       rover_exceptions.hpp
       rover.hpp
       scheduler.hpp
-      scheduler_base.hpp
+      scheduler_impl.hpp
```
JustinPrivitera: I think I like scheduler_typed or typed_scheduler. I'm thinking about the semantics - a rover owns a scheduler. A typed_scheduler is a kind of scheduler. A scheduler_impl doesn't sound like something that would also be itself a scheduler.

src/libs/rover/rover.hpp
```diff
@@ -53,23 +53,14 @@ class ROVER_API Rover
...
-                const bool log_scale);
-  void save_bov(const std::string &file_name);
-  void get_result(Image<vtkm::Float32> &image);
-  void get_result(Image<vtkm::Float64> &image);
+  void save_bov(const std::string &filename);
 private:
   vtkmCamera m_camera;
   CameraGenerator m_ray_generator;
-  SchedulerBase *m_scheduler;
+  Scheduler *m_scheduler;
```
JustinPrivitera, line 61: Was this always a pointer? Or was this something we changed along the way to refactoring scheduler stuff?

JustinPrivitera (commented):
Looks good so far!

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
siramok: I just took a peek, it's used by `mesh_blueprint_save` which then eventually passes it to `conduit::relay::io::blueprint::generate_root_filename`. Not that important for us.

src/libs/rover/CMakeLists.txt
siramok: I renamed it.

src/libs/rover/rover.hpp
siramok: I believe this was always a pointer, that's how we get away with type erasure. Holding a pointer to the non-templated base class works around the compiler needing to know what type we'll actually use at compile time. The base class declares all of the virtual methods that we want to call, and the templated class overrides them all with actual implementations. I guess as long as the inheriting class implements every virtual function from the base class, the compiler is happy.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
siramok: I adjusted the TODO to just say that it needs to be validated in verify_params.

src/libs/rover/rover.hpp
siramok: A note for future reference, if/when ascent moves to C++17 I think this type erasure could be handled in a better way using `std::variant`. Deciding that we don't need to support 2 different precisions also would simplify things.

siramok:
From the perspective of reorganizing the code that already existed, I think this PR is in a great place. However, since the scheduler does important things, I'm planning to take a pass over the full contents of `scheduler_typed.cpp` today to 1) make sure all of the nitty gritty details pass through my eyeballs, 2) further clean things up where possible, and 3) add/edit comments as needed.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -146,7 +146,7 @@ RoverXRay::verify_params(const conduit::Node &params,
     }
     else
     {
-      int width = params["rover/width"].to_int32();
+      const int32 width = params["rover/width"].value();
```
cyrush: note: its still better to use `to_` b/c if a user passes something with a different bitwidth, value will throw an error.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -280,52 +280,55 @@ RoverXRay::execute()
...
     conduit::Node &data = multi_domain.append();
     rover.to_blueprint(data);
 
     if (data.has_path("coordsets"))
     {
-      double time = -1;
-      if (metadata.has_path("time"))
+      float64 time = -1;
+      // TODO: Is this ever not true?
+      if (metadata.has_child("time"))
```
cyrush: the state entries like time and cycle are optional in Blueprint, rare that they would be skipped in a real case, but they don't have to be there

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
siramok: Interesting, I have a lot of instances of `.value()` throughout rover. Can you clarify when to use `.to_` and when not to? For example, is it a concern only in `verify_params`? (We can also discuss this offline)

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
cyrush: if you have a user facing numeric scalar, you should generally use `to_int64` and `to_float64`, since when folks enter as literal via yaml -- we may have an int when we wanted a float, and we have a unknown bitwidth.  It's best to think about what you want to consume and generally use wide types to support the broadest set of inputs.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
cyrush: if you have a user facing numeric scalar, you should generally use to_int64 and to_float64, since when folks enter as literal via yaml -- we may have an int when we wanted a float, and we have a unknown bitwidth. It's best to think about what you want to consume and generally use wide types to support the broadest set of inputs

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -280,49 +282,35 @@ RoverXRay::execute()
...
-
-      if(Metadata::n_metadata.has_path("cycle"))
-      {
-        cycle = Metadata::n_metadata["cycle"].to_int32();
-      }
-      if(Metadata::n_metadata.has_path("time"))
+      if(Metadata::n_metadata.has_child("cycle"))
       {
-        time = Metadata::n_metadata["time"].to_float64();
+        data["state/cycle"].set(Metadata::n_metadata["cycle"]);
```
siramok, line 297: Is this valid? (might have to look at the actual changed source, this diff is misleading)

JustinPrivitera: yes - but you should be able to verify

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
JustinPrivitera: Thanks for chiming in with the explanation Cyrus!

JustinPrivitera: It would be good to go through and audit our use of `.value()`. I apologize for not catching this previously.

src/libs/rover/image.cpp
```diff
@@ -71,7 +71,7 @@ template<typename FloatType>
 void
 Image<FloatType>::normalize_handle(vtkm::cont::ArrayHandle<FloatType> &handle, bool invert)
 {
-
+  // TODO: Surely we can do better than "name meaningless"
   vtkm::cont::Field as_field("name meaningless",
```
JustinPrivitera, line 76: We should ask ourselves what this is used for. Is the name going to ever be used anywhere?

src/libs/rover/scheduler_typed.cpp
```diff
@@ -203,20 +197,24 @@ Scheduler<FloatType>::set_global_bounds()
...
-  ROVER_DATA_ADD("set_global_bounds", time);
+  ROVER_DATA_ADD("set_global_range_and_bounds", time);
 }
 
 template<typename FloatType>
 void
-Scheduler<FloatType>::add_partial(vtkmRayTracing::PartialComposite<FloatType> &partial)
+SchedulerTyped<FloatType>::add_partial(vtkmRayTracing::PartialComposite<FloatType> &partial)
 {
+  // TODO: We might be able to simplify this by using emplace_back
```
JustinPrivitera: 👀 Let's discuss 👀

src/libs/rover/scheduler_typed.cpp
```diff
@@ -327,96 +326,89 @@ Scheduler<FloatType>::composite()
...
     std::stringstream domain_s;
-    domain_s<<"trace_domain_"<<i;
+    domain_s << "trace_domain_" << i;
     ROVER_DATA_OPEN(domain_s.str());
 
     vtkmLogger::GetInstance()->Clear();
-    if(dynamic_cast<CameraGenerator*>(m_ray_generator) != NULL)
+
+    // TODO: Don't love that we need dynamic_cast
+    if (!dynamic_cast<CameraGenerator*>(m_ray_generator))
```
JustinPrivitera: we need to take a pass at simplifying the "camera generator" stuff

src/libs/rover/scheduler_typed.cpp
```diff
@@ -483,46 +473,56 @@ Scheduler<FloatType>::trace_rays()
...
+  
+  if (rover::settings.has_child("image_params"))
+  {
+    // Rover's verify_params ensures that if any of these
+    // are set, all of these are set
+    has_image_params = true;
+    min_value = rover::settings["image_params/min_value"].value();
+    max_value = rover::settings["image_params/max_value"].value();
+    log_scale = rover::settings["image_params/log_scale"].as_string() == "true";
+  }
```
JustinPrivitera: can we ensure that if we have "image_params" then we are guaranteed to have "min_value", "max_value", and "log_scale"?

src/libs/rover/scheduler_typed.cpp
```diff
@@ -638,6 +603,9 @@ Scheduler<FloatType>::to_blueprint(Node &data)
   Node &n_coords = data["coordsets"][coord_name];
   n_coords["type"] = "rectilinear";
   
+  const int32 width = rover::settings["rover/width"].value();
+  const int32 height = rover::settings["rover/height"].value();
```
JustinPrivitera: to_int32()

JustinPrivitera (changes requested):
Minor feedback

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
JustinPrivitera: It makes sense that it would get passed to blueprint save, and `mesh_blueprint_save` actually does important things with `num_files` after passing it to `conduit::relay::io::blueprint::generate_root_filename`. I know because I have written much of that code.

But num files is only really useful if we have more than one domain, and currently our output is single domain. We don't partition the output at all from VisIt. I think we should leave num_files in just in case we decide to do domain decomposition later and want control over the number of files written.

src/libs/rover/scheduler_typed.cpp
siramok: Big agree

src/libs/rover/image.cpp
siramok: I suspect it probably is actually meaningless, but I think documenting it properly would be nice.

src/libs/rover/scheduler_typed.cpp
siramok: Currently yes (maybe later we wouldn't want this to be all or nothing). Within rover's verify_params, if `params.has_child("image_params")` is true then all of these parameters have to be set, else verification fails.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
siramok: I've made a pass at fixing this.

src/libs/rover/scheduler_typed.cpp
siramok: I believe this has been resolved.

siramok:
Whoops, I thought I already marked this ready for review. Barring any further requested changes, I feel that this PR is good to go.

src/libs/rover/image.cpp
```diff
@@ -352,8 +354,8 @@ template<typename FloatType>
 int
 Image<FloatType>::get_size()
 {
-  const int32 width = rover::settings["rover/width"].value();
-  const int32 height = rover::settings["rover/height"].value();
+  const int64 width = rover::settings["rover/width"].to_int64();
+  const int64 height = rover::settings["rover/height"].to_int64();
   return width * height;
```
JustinPrivitera: I am wondering if this method should even live in `Image` now that width and height do not belong to it.

src/libs/rover/partial_image.hpp
```diff
@@ -289,9 +289,9 @@ struct PartialImage
     const int size = m_pixel_ids.GetNumberOfValues();
     const int num_channels = m_buffer.GetNumChannels();
     const bool has_emission = m_intensities.Buffer.GetNumberOfValues() != 0;
-    const int32 width = rover::settings["rover/width"].value();
-    const int32 height = rover::settings["rover/height"].value();
-    const int debug = width * (height - y) + x;
+    const int64 width = rover::settings["rover/width"].to_int64();
+    const int64 height = rover::settings["rover/height"].to_int64();
+    const long debug = width * (height - y) + x;
```
JustinPrivitera: use int64 instead

src/libs/rover/partial_image.hpp
```diff
@@ -317,9 +317,9 @@ struct PartialImage
   {
     const int size = m_pixel_ids.GetNumberOfValues();
     const int num_channels = m_buffer.GetNumChannels();
-    const int32 width = rover::settings["rover/width"].value();
-    const int32 height = rover::settings["rover/height"].value();
-    const int debug = width * (height - y) + x;
+    const int64 width = rover::settings["rover/width"].to_int64();
+    const int64 height = rover::settings["rover/height"].to_int64();
+    const long debug = width * (height - y) + x;
```
JustinPrivitera: use int64 instead

src/libs/rover/ray_generators/visit_generator.cpp
```diff
@@ -72,9 +72,9 @@ VisitGenerator::gen_rays(vtkmRayTracing::Ray<T> &rays)
   vtkmTimer timer;
   double time = 0;
   ROVER_DATA_OPEN("visit_ray_gen");
-  const int32 width = rover::settings["rover/width"].value();
-  const int32 height = rover::settings["rover/height"].value();
-  const int size = width * height;
+  const int64 width = rover::settings["rover/width"].to_int64();
+  const int64 height = rover::settings["rover/height"].to_int64();
+  const long size = width * height;
```
JustinPrivitera: use int64 instead

src/libs/rover/typed_scheduler.cpp
```diff
@@ -545,74 +545,39 @@ void Scheduler<FloatType>::save_result(std::string file_name)
...
-    FloatType * buffer
-      = get_vtkm_ptr(m_result.get_intensity(i));
-
-    encoder.EncodeChannel(buffer, width, height);
-    encoder.Save(sstream.str());
-  }
-}
+  const int64 width = rover::settings["rover/width"].to_int64();
+  const int64 height = rover::settings["rover/height"].to_int64();
+  const long size = height * width;
```
JustinPrivitera: int64

JustinPrivitera (approved):

src/libs/rover/image.cpp
siramok: Turns out that it wasn't used anywhere anyways, removed.

PR #1556, opened 2025-06-13
Add spatial topology blueprint mesh to Rover's output

siramok:
Addresses bullets in #1513

- Using VisIt as a guide, rover now generates a spatial topology blueprint mesh.
- Metadata is now plumbed through rover, specifically time and cycle (better output parity with VisIt).
- xray_data information has been added to the output state.
- Rover tests have been refactored and now render every field as part of each test.
- The existing tests now have "rotated" variants.
- In all cases, spatial fields have rotated baseline images so that their depth can be observed.
- There are now 2 conduit blueprint diff tests.


siramok:
VisIt uses separate functions to fill the output state, topologies, coordsets, and fields. For organization's sake I think it would be best to copy that approach before adding additional blueprint meshes, but I haven't currently done so.

src/libs/rover/rover.cpp
```diff
@@ -26,6 +26,12 @@ Rover::Rover()
 {
   // Ensure that we always start from a default state
   rover::settings.reset();
+
+  // Metadata
+  rover::settings["state/time"] = -1.0f;
+  rover::settings["state/cycle"] = -1;
```
JustinPrivitera: I'm not sure how I feel about these default values. It probably makes more sense to not include these by default, and only add them if the dataset includes them.

You should craft a test case that does not have cycle or time and see what happens.

JustinPrivitera:
> VisIt uses separate functions to fill the output state, topologies, coordsets, and fields. For organization's sake I think it would be best to copy that approach before adding additional blueprint meshes, but I haven't currently done so.

I agree that that will be the best approach once we get everything into the output. Otherwise it will be too cluttered.

JustinPrivitera:
I will review after you add multiple tests that create renders of the spatial extents mesh. Ideally the camera is rotated so that the mesh is not viewed from head on but at an angle so that we can see that it is different from the image topo mesh.

JustinPrivitera:
We can also rotate the x ray detector so that we create more interesting renders.

JustinPrivitera:
We should also add tests that diff the state node outputs with an inlined conduit node with the correct answers.

JustinPrivitera:
We should have something like this: https://github.com/visit-dav/visit/blob/355dd795dcbde0b692d8284cfe632c8775f63555/src/test/tests/queries/xrayimage.py#L620

JustinPrivitera:
Let's have renders for optical_depth fields as well. We should attempt to have tests that capture every single thing included in the output and the inputs we provide in tests should cover as many of the edge cases as possible.

siramok:
> Let's have renders for optical_depth fields as well. We should attempt to have tests that capture every single thing included in the output and the inputs we provide in tests should cover as many of the edge cases as possible.

We discovered offline that the optical_depth fields are currently broken and have decided to not include baselines for them until they're fixed.

src/libs/rover/rover.cpp
```diff
@@ -176,63 +176,83 @@ Rover::update_camera()
...
-  if (rover::settings.has_path("camera/look_at"))
+  if (n_camera.has_child("elevation"))
   {
-    float64_accessor vec3 = rover::settings["camera/look_at"].value();
-    vtkmVec3f look_at(vec3[0], vec3[1], vec3[2]);
+    const float64 elevation = n_camera["elevation"].to_float64();
+    m_camera.Elevation(elevation);
+  }
+
+  if (n_camera.has_child("zoom"))
```
JustinPrivitera: m_camera and n_camera are confusing... maybe we rename n_camera to be more descriptive.

src/tests/_baseline_images/tout_rover_xray_blueprint_braid_rotated_intensities_000100.png
JustinPrivitera, line 1: This rotation is cool, but it might also be neat to rotate the actual mesh output by a few degrees so we can see that it is a 3d box. TLDR both kinds of rotation are good.

src/tests/_baseline_images/tout_rover_xray_blueprint_multi_curv3d_rotated_intensities_spatial_000048.png
JustinPrivitera, line 1: is there an option in rover for parallel projection versus perspective projection? If not, we should add one. If there is, we should add tests for it.

siramok: There currently is not, but it's on the list of things to add.

src/tests/_baseline_images/tout_rover_xray_blueprint_braid_rotated_intensities_000100.png
siramok: Ah I see, I can work on that.

src/libs/rover/rover.cpp
siramok: Agree with this.

siramok: This has been addressed.

src/tests/_baseline_images/tout_rover_xray_blueprint_braid_rotated_intensities_000100.png
siramok: This has been addressed.

siramok:
Whoops, didn't mean to commit all the diff images. Fixing...

src/libs/rover/rover.cpp
siramok: I gave this a try by taking an example braid mesh and calling `.remove("state");` on it before performing an xray extract. Setting default values for time and cycle worked but not setting them at all also worked. Given that, I've removed these default values, and now only append time/cycle to the blueprint output if it's provided by metadata.

siramok:
Seems that CI doesn't like the conduit blueprint diff tests.

siramok:
Decided to comment them out for now while we discuss what to do next.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -267,6 +267,7 @@ RoverXRay::execute()
 
   // Initialize rover and configure its behavior with the input params
   Rover rover;
+  rover.update_metadata(Metadata::n_metadata);
```
JustinPrivitera: let's use a more descriptive name for this method. Metadata can refer to everything we put under state, while this method only handles cycle and time.

src/tests/_baseline_images/tout_rover_xray_blueprint_curv3d_camera_params_intensities_000048.png
JustinPrivitera, line 1: why is this one slanted again? Is the camera slanted like that on purpose?

src/tests/_baseline_images/tout_rover_xray_params_000100.png
JustinPrivitera, line 1: what is this file here for?

src/tests/data/rover_xray_blueprint_braid_baseline.root
JustinPrivitera, line 1: what is this file for?

src/tests/data/rover_xray_blueprint_curv3d_baseline.root
JustinPrivitera, line 1: what is this file for?

src/tests/_baseline_images/tout_rover_xray_blueprint_curv3d_camera_params_intensities_000048.png
siramok: Yes, it's intentional. I rotated the camera for the spatial field baseline images to address your feedback about having both types of rotation. Since the spatial fields are always 3D, I figured why not tilt the camera so that their depth can be observed.

src/tests/_baseline_images/tout_rover_xray_params_000100.png
siramok: I think this may have been old, as far as I can tell it's not used for anything anymore.

src/tests/data/rover_xray_blueprint_braid_baseline.root
siramok: It's rover's output from the example conduit braid mesh, it's used as a baseline for the braid blueprint diff test.

src/tests/data/rover_xray_blueprint_curv3d_baseline.root
siramok: It's rover's output from the curv3d data, it's used as a baseline for the curv3d blueprint diff test.

siramok:
Paths failing the diff tests:

braid:
- domain_000000/state/xray_data/optical_depth_max
- domain_000000/fields/intensities/values
- domain_000000/fields/optical_depth/values
- domain_000000/fields/intensities_spatial/values
- domain_000000/fields/optical_depth_spatial/values

curv3d:
- domain_000000/fields/intensities/values
- domain_000000/fields/optical_depth/values
- domain_000000/fields/intensities_spatial/values
- domain_000000/fields/optical_depth_spatial/values

src/tests/_baseline_images/tout_rover_xray_params_000100.png
JustinPrivitera: doesn't need to be in this PR then. Let's remove.

src/tests/data/rover_xray_blueprint_braid_baseline.root
JustinPrivitera: like we discussed, likely better to just do inline node as a string and then parse and diff.

src/tests/data/rover_xray_blueprint_curv3d_baseline.root
JustinPrivitera: see above.

JustinPrivitera:
> Paths failing the diff tests:
> 
> braid:
> 
>     * domain_000000/state/xray_data/optical_depth_max
> 
>     * domain_000000/fields/intensities/values
> 
>     * domain_000000/fields/optical_depth/values
> 
>     * domain_000000/fields/intensities_spatial/values
> 
>     * domain_000000/fields/optical_depth_spatial/values
> 
> 
> curv3d:
> 
>     * domain_000000/fields/intensities/values
> 
>     * domain_000000/fields/optical_depth/values
> 
>     * domain_000000/fields/intensities_spatial/values
> 
>     * domain_000000/fields/optical_depth_spatial/values

Let's follow the visit test more closely and inline a conduit node with just the state info. If we want to diff fields we can try that, but they are likely to be different on different platforms as we have seen.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
siramok: This has been addressed.

src/tests/_baseline_images/tout_rover_xray_params_000100.png
siramok: I brought it back.

src/tests/data/rover_xray_blueprint_braid_baseline.root
siramok: This has been addressed.

src/tests/data/rover_xray_blueprint_curv3d_baseline.root
siramok: This has been addressed.

src/tests/_baseline_images/tout_rover_xray_params_000100.png
JustinPrivitera: I'm confused; the github diff suggests that you are deleting this file with this PR. But you said "I brought it back." Where did it come from? Why are we deleting it?

src/libs/rover/typed_scheduler.cpp
```diff
@@ -652,8 +652,9 @@ TypedScheduler<FloatType>::to_blueprint(Node &data)
   xray_data["detector_height"] = detector_height;
   xray_data["intensity_max"].set(DataType::float64(1));
   xray_data["intensity_min"].set(DataType::float64(1));
-  xray_data["optical_depth_max"].set(DataType::float64(1));
-  xray_data["optical_depth_min"].set(DataType::float64(1));
+  // TODO: Uncomment this when optical_depth is fixed
+  // xray_data["optical_depth_max"].set(DataType::float64(1));
+  // xray_data["optical_depth_min"].set(DataType::float64(1));
```
JustinPrivitera: FYI: I don't think you need to set the type of the data to float64 if you are later fetching a float64 and using .set to put it in the node.

siramok: Good to know. Since this change didn't fix CI like I had hoped anyways, I'll just revert it to how it was.

siramok: Actually, your comment is applicable either way, will address.

siramok: Double actually, doing it this way forces the blueprint output to be ordered the same way as visit's xray query output (not functionally different, but nice), so I think it's best to leave alone. Thanks for making me consider this decision more carefully, and sorry for the whiplash.

src/tests/_baseline_images/tout_rover_xray_params_000100.png
siramok: Yes, let me clarify. This seems to be one of the original baseline images for rover, originally committed in 2020 with 2-3 renames since then. Since this PR added a bunch of new baseline images with field names appended, I noticed that this image wasn't used for anything anymore and removed it. Then I interpreted your comment "doesn't need to be in this PR then" to mean "let's leave this file alone and not make this change right now" and therefore brought it back.

src/libs/rover/typed_scheduler.cpp
JustinPrivitera: The order doesn't necessarily matter. If you were concerned about order, you could just do 
```c++
xray_data["optical_depth_max"];
xray_data["optical_depth_min"];
```
to add those children to the node.

siramok: Nice, the best of every world, I'll do this.

JustinPrivitera (approved):

PR #1561, opened 2025-06-18
add vtkm patch for z extents ray culling bug

nicolemarsaglia:
add vtkm patch for z extents ray culling bug


cyrush (approved):
thanks for hunting this down and getting this patch in!

PR #1568, opened 2025-06-26
Task/6 26 25 1536 conduit error when rendering a rectilinear mesh with integer coordinate values

### LINKED ISSUE #1536: Conduit error when rendering a rectilinear mesh with integer coordinate values
A Conduit error is thrown when attempting to render a pseudocolor plot of a rectilinear mesh with integer (both int32 and int64) (x, y, z) coordinates:
```
Node::as_float64_ptr() const -- DataType int64 at path domain_000000/coordsets/image_coords/values/x does not equal expected DataType float64
```

This error does not occur when using float for the (x, y, z) coordinates (I can provide an example if that would be helpful).

The mesh I used: [tout_rover_xray_query100.cycle_000100.zip](https://github.com/user-attachments/files/20378263/tout_rover_xray_query100.cycle_000100.zip)

The Ascent actions I used:
```
...
conduit::Node scenes;
scenes["s1/plots/p1/type"] = "pseudocolor";
scenes["s1/plots/p1/field"] = "intensities";

conduit::Node actions;
conduit::Node &add_plots = actions.append();
add_plots["action"] = "add_scenes";
add_plots["scenes"] = scenes;

Ascent ascent;
conduit::Node ascent_opts;
ascent.open(ascent_opts);
ascent.publish(data);
ascent.execute(actions);
...
```

emily-howell:


src/tests/data/tout_rover_xray_query100.cycle_000100.root
JustinPrivitera, line 1: What is this file and what changed about it?
@emily-howell @siramok

emily-howell: New file, not changed

```
domain_000000: 
  state: 
    cycle: 100
    time: 3.1414999961853
    domain_id: 0
  coordsets: 
    image_coords: 
      type: "rectilinear"
      values: 
        x: [0, 1]
        y: [0, 1, 2, ..., 1023, 1024]
        z: [0, 1, 2, ..., 1023, 1024]
      labels: 
        x: "energy_group"
        y: "width"
        z: "height"
      units: 
        x: "bins"
        y: "pixels"
        z: "pixels"
  topologies: 
    image_topo: 
      coordset: "image_coords"
      type: "rectilinear"
  fields: 
    intensities: 
      topology: "image_topo"
      association: "element"
      units: "intensity units"
      values: [1.0, 1.0, 1.0, ..., 1.0, 1.0]
      strides: [1, 1, 1024]
    optical_depth: 
      topology: "image_topo"
      association: "element"
      units: "path length metadata"
      values: [1.0, 1.0, 1.0, ..., 1.0, 1.0]
      strides: [1, 1, 1024]
```

emily-howell: It is the file that @siramok added to his issue

JustinPrivitera: Gotcha. It might be good to give it a different name, since the naming convention it has now is that of a test output, while it is now being used as a test input.

emily-howell: Name updated, tests still passing 😄

JustinPrivitera: Thanks!

src/libs/ascent/runtimes/ascent_vtkh_data_adapter.cpp
```diff
@@ -1336,16 +1336,46 @@ VTKHDataAdapter::RectilinearBlueprintToVTKmDataSet
...
 
-    // todo assumes float64
-    const float64 *x_coords_ptr = n_coords["values/x"].as_float64_ptr();
-    const float64 *y_coords_ptr = n_coords["values/y"].as_float64_ptr();
+    const float64 *x_coords_ptr;
+    if (!n_coords["values/x"].dtype().is_float64())
+    {
+        Node temp_x;
+        n_coords["values/x"].to_float64_array(temp_x);
+        x_coords_ptr = temp_x.value();
```
cyrush: temp_x goes out of scope here, so this pointer will not be valid.
temp_x, etc have to exist one level up

cyrush:
We also can't zero-copy if we are using these temp arrays, so we need to take that into account.

src/libs/ascent/runtimes/ascent_vtkh_data_adapter.cpp
```diff
@@ -1336,45 +1336,58 @@ VTKHDataAdapter::RectilinearBlueprintToVTKmDataSet
 
     int32 ndims = 2;
 
+    if (zero_copy && 
+        (!n_coords["values/x"].dtype().is_float64() || 
+         !n_coords["values/y"].dtype().is_float64()))
+    {
+        ASCENT_ERROR("Zero-copy requested, but x or y coordinate data is not float64.\n" <<
+                    "x type: " << n_coords["values/x"].dtype().name() << 
+                    ", y type: " << n_coords["values/y"].dtype().name());
```
cyrush: we can actually set `zero_copy = false;`, there is precedent for other cases where we can't make the zero copy wish come true.

cyrush (approved):
Looks good, thanks for the updates!

Issue #1449, opened 2025-01-30
sampling interface additions


cyrush:
For uniform grid: 
- Allow spatial extents to be specified as an alternative to `origin` and `spacing`. This from will be more natural for many use cases
- Allow `fields` in addition to `field`, and support sampling multiple fields.

cyrush:
Expand sampling cases beyond uniform grids, with a generic "sample" filter.

We should provide an interface that mirrors clip (perhaps expanded to also include the seed patterns in the streamline filter)

- box (this basically is our uniform grid filter)
- line (not in clip, but perhaps should be) 
- plane
- sphere
- cylinder
- mesh: allow users to provide a blueprint mesh to sample onto

Like current sampling - start with results only on rank 0 (this is the most pressing use case)

cyrush:
@nicolemarsaglia I updated this ticket with ideas about expanding how we leverage sampling to provide users more options.

cyrush:
Box Cases:

```
type: "sample"
params: 
          field: "den"
          # or fields: ["den", "temp"]
          box:
            dims:
              i: 1500
              j: 500
            origin:
              x: 0.0
              y: 0.0
            spacing:
              dx: .01
              dy: .01 
```
```
type: "sample"
params: 
          field: "den"
          # or fields: ["den", "temp"]
          box:
            dims:
              i: 1500
              j: 500
            min:
               x: 0.0
               y: 0.0
            max:
               x: 100
               y: 100
```

Line case:
```
type: "sample"
params: 
          field: "den"
          # or fields: ["den", "temp"]
          line:
            samples: 100
            start:
               x: 0.0
               y: 0.0
            end:
               x: 100
               y: 100
```

Points case:
```
type: "sample"
params: 
          field: "den"
          # or fields: ["den", "temp"]
          points:
            x: [ x0, x1, x2, ... ]
            y: [ x0, x1, x2, ... ]
```

cyrush:
`plane`, `cylinder`, `sphere` -- require us to create a target mesh.  It's more complex than setting streamline source points or specifying a clip.

The target mesh for `box`, `line`, and `points` is straight forward.  We should support those first.

PR #1571, opened 2025-06-27
add a sample filter that samples at points (list or specd by line)

cyrush:
working towards #1449


cyrush:
This work builds off @nicolemarsaglia's uniform grid sampling work to add sampling at a list of points, or via a line (start, end, and number of samples).

I creating `vtkh::Sampler` from `vtkh::UniformGrid`. It leverages that reduction logic.

I found a few cases where types were confused due to `using` statements across header files so I renamed all of the using types to avoid future confusion.  There are also a few cases we can use newer vtk-m helpers to init arrays.

Ultimately, the `vtkh::Sampler` should subsume the uniform grid case so we have a single filter. 

Added tests for 2d/3d line sampling as well as 2d/3d list of points.

It does not seem the `radial` file is sampled correctly for the list of points cases. Radial isn't sampled in the line cases. 

Current tests create extracts, need to add render tests as well.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_vtkh_filters.cpp
```diff
@@ -3929,6 +3930,244 @@ VTKHUniformGrid::execute()
...
+}
+
+//-----------------------------------------------------------------------------
+void
+VTKHSample::execute()
+{
+
+    if(!input(0).check_type<DataObject>())
+    {
+        ASCENT_ERROR("vtkh_uniform_grid input must be a data object");
```
nicolemarsaglia: leftover copy pasta: vtkh_uniform_grid -> vtkh_sample

cyrush:
radial is fine, i had a pathological case due to bad params:
Screenshot 2025-07-01 at 10 35 47 AM

src/libs/vtkh/vtkm_filters/vtkmProbe.cpp
```diff
@@ -5,54 +5,138 @@
...
+      for(int i = 0; i < num_points; ++i)
+      {
+          conn_portal.Set(i, i);
+      }
+      cellset.Fill(num_points, shape_id, indices_per, connectivity);
+      ds_probe.SetCellSet(cellset);
   }
+  else // error
+  {
+      // TODO: ERROR!
```
nicolemarsaglia: something like? ASCENT_ERROR("vtkmProbe does not recognize sampled mode: " << m_mode);

nicolemarsaglia (approved):
Looks great! Thanks for getting this done!

src/libs/vtkh/vtkm_filters/vtkmProbe.cpp
cyrush: vtk-h does not use ascent error, but has its own error class -- adding a throw of an instance of that error class.

nicolemarsaglia: oh yeah, duh!

Issue #1559, opened 2025-06-16
The largest bounding box of a multi-topology dataset is always used when rendering


siramok:
The xray extract's output defines several meshes. Some of the meshes are defined in image-space and some are defined in world-space, but both types of meshes reuse the same field values. Here's an example dataset that exhibits the problem: [multiple-topologies.zip.](https://github.com/user-attachments/files/20762500/multiple-topologies.zip) The image-space extents in this example are (0, 200, 0, 200, 0, 1) while the world-space extents are (0, 80, 0, 80, 0, 1).

Here's what ascent outputs if I ask it to render a basic psuedocolor plot of one of the image-space fields ("intensities"):
![Image](https://github.com/user-attachments/assets/49193f2d-d4f9-4729-be20-b933d450273e)

And here's what ascent outputs if I ask it to render a basic psuedocolor plot of one of the world-space fields ("intensities_spatial"):
![Image](https://github.com/user-attachments/assets/013c9eef-9aaa-4fad-9198-afc97246af39)

I'm able to produce the expected render of the "intensities_spatial" field by applying a box clip:
![Image](https://github.com/user-attachments/assets/e7d9d049-2fbc-4f72-82b2-c745527b71d0)

While troubleshooting, I ruled out a few things:
- The order in which topologies, coordsets, and fields appear in the output doesn't make a difference.
- The alphabetical order of the topology/coordset names doesn't make a difference (e.g. `aspatial_topo` vs `spatial_topo`).
- Forcing all fields to use the `spatial_topo` topology didn't make a difference as long as `image_topo` was defined and referenced `image_coords`.

My best guess as to what's happening here: it seems like ascent iterates over every topology and computes a bounding box for each coordset, and then always uses the largest bounding box it finds. If that's true, then I think a solution would be to only compute bounding boxes on coordsets that are actually used in the current scene, and to then pick the largest bounding box of those (and maybe do some merging if necessary). That way we always use the smallest possible bounding box that still encompasses the current scene.

nicolemarsaglia:
Rover eventually uses vtkh? 

We union bounds here: https://github.com/Alpine-DAV/ascent/blob/77fd3b9ccdf9a08d581f801c3548366b64adcf91/src/libs/ascent/runtimes/flow_filters/ascent_runtime_rendering_filters.cpp#L1504

That's likely to be the culprit

siramok:
[...]

PR #1573, opened 2025-07-08
Resolve many Rover TODOs

siramok:
Addresses bullets in #1513

- Improve rover's verify_params (much more thorough)
- Improve how rover's output types are enabled
- Add dataset validation before calling rover execute
- Refactor rover's global settings to avoid including "rover/" in every path
- Remove unused settings
- Improved intensity and optical depth buffer names
- Improved ray generator class naming scheme
- Add a parameter to specify a background intensity (test added)
- Merge the diff tests with existing tests
- Improvements throughout typed_scheduler.cpp and image.cpp
- Implement divide_emis_by_absorb (test added)
- Adds imaging plane and ray meshes to rover's output (no test added)

(we can split this into smaller PRs if desired)


src/libs/rover/domain.cpp
```diff
@@ -18,6 +18,12 @@ Domain::Domain()
   m_engine = std::make_shared<Engine>();
 }
 
+Domain::Domain(vtkmDataSet &dataset)
```
siramok, line 21: Adding a new constructor for `Domain` that directly takes a dataset allowed for simplification in `typed_scheduler.cpp`.

src/libs/rover/image.cpp
```diff
@@ -173,104 +173,98 @@ template<typename FloatType>
...
+  const int64 height = rover::settings["height"].to_int64();
   const int64 channel_size = width * height;
-  const int num_channels = partial.m_buffer.GetNumChannels();
-
-  // TODO: This could be simplified with a helper function
-  for (int i = 0; i < num_channels; i++)
+  const int num_channels = partial.m_transmission.GetNumChannels();
+
+  // Helper lambda to expand a channel and push its buffer to the output vector
+  auto expand_and_push = [&](int channel_index,
```
siramok, line 206: Don't think this is much different in terms of efficiency, debatable whether it's more readable, but extracting out the common functionality did simplify the loop body by quite a bit.

src/libs/rover/partial_image.hpp
```diff
@@ -27,22 +27,22 @@ struct PartialImage
   // TODO: Improve naming to reflect what these things actually represent,
   // will require changes elsewhere (absorptionpartial and emissionpartial)
   IdHandle                                 m_pixel_ids;
-  vtkmRayTracing::ChannelBuffer<FloatType> m_buffer;          // holds the absorption
-  vtkmRayTracing::ChannelBuffer<FloatType> m_intensities;     // holds the intensity emerging from each ray
-  vtkmRayTracing::ChannelBuffer<FloatType> m_optical_depths;
+  vtkmRayTracing::ChannelBuffer<FloatType> m_transmission;          // holds the absorption
```
siramok: I'm not particularly attached to `m_transmission`, but I asked Livchat for help coming up with a better name for this buffer with respect to what it's actually filled with. This was the reply (which I found convincing enough):

> This value accumulates the total transmission (i.e., the fraction of incoming intensity that survives absorption) along the ray up to this segment. If you start with intensity 1, after passing through all segments, this bin holds the fraction of that intensity remaining due to absorption.

src/libs/rover/ray_generators/ray_generator.hpp
```diff
@@ -48,11 +48,14 @@
 namespace rover
 {
 
+// TODO: Since the visit ray generator ultimately modifies a vtkm camera anyways, better
```
siramok, line 51: I don't think this is a top priority, but I do think this would be a good thing to do down the road.

src/libs/rover/rover.cpp
```diff
@@ -20,23 +20,23 @@ namespace rover
 {
 
 // Namespaced settings instance
+Node metadata;
```
siramok, line 23: We didn't discuss this change previously, but I decided that the codebase would be slightly nicer without having `rover/` in all of the settings queries. I'm happy with the result of that change, but simultaneously leaving `state/time` and `state/cycle` in the settings node was complicating things ever so slightly when it came time to create the blueprint output. I figured having a second global `Node` for metadata was a simple, unintrusive solution for that.

src/libs/rover/rover.cpp
```diff
@@ -20,23 +20,23 @@ namespace rover
...
 
 Rover::Rover()
 {
-  // Ensure that we always start from a default state
+  // Always start from a default state
+  rover::metadata.reset();
   rover::settings.reset();
 
   // Settings
-  // TODO: Figure out if color_table needs to be set by us here
```
siramok, line 31: We weren't really using this option. Setting a colormap was relevant for volume rendering, but I couldn't find any indication that it mattered for XRay stuff. So I ripped it all out, along with the other options that were never being used for anything. I figure that if they represent important features, that they can be added back in once there's code to actually hook up with these options.

src/libs/rover/rover.cpp
```diff
@@ -61,37 +61,38 @@ Rover::~Rover()
...
   ROVER_INFO("Executing Rover::update_settings");
 
   if (params.has_child("rover"))
   {
-    // TODO: use .update
-    for (const auto &param_name : params["rover"].child_names())
-    {
-      rover::settings["rover"][param_name].set(params["rover"][param_name]);
-    }
+    rover::settings.update(params["rover"]);
```
siramok, line 84: Much nicer, thanks to Cyrus.

src/libs/rover/rover.cpp
```diff
@@ -61,37 +61,38 @@ Rover::~Rover()
...
   }
 
   if (params.has_child("camera"))
   {
     rover::settings["camera"].set(params["camera"]);
   }
+
+  if (params.has_child("image_params"))
```
siramok, line 92: These options are only relevant for the .png output (which is broken (?) right now anyways), but it wasn't actually hooked up before this change.

src/libs/rover/rover.cpp
```diff
@@ -155,12 +156,14 @@ Rover::add_dataset(vtkh::DataSet &dataset)
...
   }
 
-  for (int i = 0; i < dataset.GetNumberOfDomains(); i++)
-  {
-    m_scheduler->add_dataset(dataset.GetDomain(i));
-  }
-
+  m_scheduler->add_dataset(dataset);
   m_camera.ResetToBounds(dataset.GetGlobalBounds());
+  // TODO: Do we want to further tighten the bounds? vtkm sets
```
siramok: I think this is worth thinking about further. It would still be relevant even if #1559 is fixed.

src/libs/rover/rover.cpp
```diff
@@ -172,6 +175,7 @@ Rover::update_camera()
     return;
   }
 
+  // The order in which these parameters are applied matters
```
siramok, line 175: Per Nicole, just making an explicit note within the code.

src/libs/rover/rover.cpp
```diff
@@ -348,6 +352,13 @@ Rover::save_png(const std::string &filename)
 void
 Rover::save_bov(const std::string &filename)
 {
+#ifdef ROVER_PARALLEL
+  // TODO: Support writing in parallel
```
siramok, line 353: The other outputs had this ifdef, so I just brought it over to `save_bov`.

src/libs/vtkh/rendering/ConnectivityTracer.cpp
```diff
@@ -789,7 +784,23 @@ class IntegrateEmission : public vtkm::worklet::WorkletMapField
...
       absorptionBins.Set(rayOffsetI, absorbIntensity * tmp);
-      emissionBins.Set(rayOffsetI, emissionIntensity * tmp + emission * (1.0f - tmp));
+
+      if (DivideEmisByAbsorb)
+      {
+        if (absorb > 0)
+        {
+          emissionBins.Set(rayOffsetI, emissionIntensity * tmp + (emission / absorb) * (1.0f - tmp));
+        }
+        else // (absorb <= 0)
```
siramok: I'm not sure if this ever happens in practice, but I was thinking about how 0.0 absorption could technically happen, hence the check for avoiding division by 0.

src/tests/_baseline_images/tout_rover_xray_blueprint_braid_intensities_spatial_000100.png
siramok, line 1: Even though the field looks identical, math changes (necessary for imaging planes + rays meshes) changed the spatial extents by a factor of 1/2. If the previous extents were the correct ones, we can do `* 2` to compensate and get the old extents back.

src/tests/_baseline_images/tout_rover_xray_blueprint_braid_optical_depth_000100.png
siramok, line 1: Looks pretty much identical, but I noticed that the baselines were now "different" somehow and decided to update them all.

src/tests/_baseline_images/tout_rover_xray_blueprint_braid_rotated_intensities_000100.png
siramok, line 1: Intentionally different, since this test now includes `background_intensity = 100.0f` to further differentiate it from the unrotated test.

src/tests/_baseline_images/tout_rover_xray_blueprint_multi_curv3d_intensities_000048.png
siramok, line 1: Intentionally different, since this test enables `divide_emis_by_absorb = true`.

src/tests/ascent/t_ascent_rover.cpp
```diff
@@ -54,7 +54,8 @@ render_blueprint(const string &field_name,
                  const double detector_width)
 {
     // Define Ascent actions
-    // TODO: Remove this now
+
+    // TODO: Remove this once issue #1559 is fixed
```
siramok, line 58: I initially removed box clipping per Cyrus' wishes, but adding the imaging planes and rays meshes exacerbated the issue caused by #1559, even when they're disabled via `enable_imaging_planes` due to changes in the setup math to make imaging planes and rays meshes work correctly.

src/tests/ascent/t_ascent_rover.cpp
```diff
@@ -205,7 +205,9 @@ TEST(ascent_rover, test_xray_blueprint_braid)
     extracts["e1/params/rover/absorption"] = "radial";
     extracts["e1/params/rover/emission"] = "radial";
     extracts["e1/params/rover/filename"] = query_path;
-    extracts["e1/params/rover/blueprint"] = "yaml";
+    extracts["e1/params/rover/blueprint_output"] = "true";
+    extracts["e1/params/rover/blueprint_protocol"] = "yaml";
+    extracts["e1/params/rover/enable_imaging_planes"] = "false"; // TODO: Remove this once issue #1559 is fixed
```
siramok, line 206: Disabled on all of the existing tests. I want to create a separate test where `enable_imaging_planes"] = true`, however my first attempt at getting ascent to render the imaging planes together didn't immediately work, so I decided to just commit my progress and work on it next.

src/tests/ascent/t_ascent_rover.cpp
```diff
@@ -288,115 +348,6 @@ TEST(ascent_rover, test_xray_blueprint_braid_rotated)
     ASCENT_ACTIONS_DUMP(actions, image_path, msg);
 }
 
-//-----------------------------------------------------------------------------
-TEST(ascent_rover, test_xray_blueprint_braid_diff)
```
siramok, line 292: Per your feedback about how diff testing could just be part of the normal tests, I merged them.

src/tests/ascent/t_ascent_rover.cpp
```diff
@@ -435,7 +386,8 @@ TEST(ascent_rover, test_xray_blueprint_braid_lowres)
     extracts["e1/params/rover/absorption"] = "radial";
     extracts["e1/params/rover/emission"] = "radial";
     extracts["e1/params/rover/filename"] = query_output_file;
-    extracts["e1/params/rover/blueprint"] = "json";
+    extracts["e1/params/rover/blueprint_output"] = "true";
+    extracts["e1/params/rover/blueprint_protocol"] = "yaml";
```
siramok: We output yaml for every other test, so may as well be consistent. Except this test is disabled atm anyhow.

src/tests/_baseline_images/tout_rover_xray_blueprint_braid_intensities_spatial_000100.png
JustinPrivitera: I guess I'm confused why the spatial extents would have to change based on imaging planes and rays meshes? That seems odd to me.

src/tests/_baseline_images/tout_rover_xray_blueprint_braid_optical_depth_000100.png
JustinPrivitera: Maybe some very slight pixel diffs.

src/tests/_baseline_images/tout_rover_xray_blueprint_braid_rotated_intensities_000100.png
JustinPrivitera: Interesting. Have you verified that VisIt produces a comparable result when you provide the same background intensity? We should use the same color table and produce an image to know for sure.

src/tests/_baseline_images/tout_rover_xray_blueprint_curv3d_rotated_intensities_spatial_000048.png
JustinPrivitera, line 1: I like this render and the next one (optical depth). Showcases what's happening well.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -107,130 +107,394 @@ RoverXRay::verify_params(const conduit::Node &params,
...
       {
-        info["errors"].append() = "Optional parameter 'rover/width' must be greater than 0";
+        info["errors"].append() = "Optional numeric parameter 'rover/unit_scalar' must be positive";
         res = false;
       }
     }
+  }
 
-    if (!params["rover/height"].dtype().is_integer())
+  const bool has_blueprint_output = n_rover.has_child("blueprint_output");
```
siramok: We didn't really discuss this change, but it bothered me that .png output was always taking place despite being "broken" (I've never looked into it). My thought for how to fix this was to make all of the output types off by default, requiring the user to explicitly toggle on the output types that they want rover to produce. I reasoned that each of these outputs have different file extensions, so I just use `filename` for all of them. This makes `execute` really nice and clean, and no longer do we waste cycles outputting a .png unless that's explicitly what we want to have happen.

src/tests/_baseline_images/tout_rover_xray_blueprint_multi_curv3d_intensities_000048.png
JustinPrivitera: Have you verified a similar result in VisIt? I recommend using the same color table for both to know for sure.

src/tests/_baseline_images/tout_rover_xray_blueprint_multi_curv3d_optical_depth_000048.png
JustinPrivitera, line 1: We still need to track down the multi-domain compositing issue, but that is a challenge for another day.

src/tests/ascent/t_ascent_rover.cpp
JustinPrivitera: sounds good.

src/tests/ascent/t_ascent_rover.cpp
JustinPrivitera: ok

src/tests/ascent/t_ascent_rover.cpp
```diff
@@ -218,6 +220,61 @@ TEST(ascent_rover, test_xray_blueprint_braid)
     // Load and verify output mesh
     Node xray_blueprint_output, verify_info;
     load_and_verify_local_data(xray_blueprint_output, output_data_path);
+    Node &state_output = xray_blueprint_output["domain_000000/state"];
+    state_output.remove("xray_query/filename");
```
JustinPrivitera: I don't love modifying the data we read in from the output. It might be better to add the filename to the baseline node we create with a note that we had to modify for the diff to pass.

src/tests/ascent/t_ascent_rover.cpp
JustinPrivitera: cool

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -107,130 +107,394 @@ RoverXRay::verify_params(const conduit::Node &params,
...
-      info["errors"].append() = "Missing required image parameter 'min_value' must be a number";
+      info["errors"].append() = "Optional bool string parameter 'rover/png_output' must be 'true' or 'false'";
       res = false;
     }
+  }
 
-    if (!params.has_path("image_params/max_value") ||
-        !params["image_params/max_value"].dtype().is_number())
+  // This won't be necessary once rover becomes a filter
+  if (!has_blueprint_output && !has_bov_output && !has_png_output)
```
siramok: Building on the prior comment about making output types toggleable, if no output types are specified then we essentially have no work to do. Whether we consider this an error (as I have gone ahead and done here) or whether we check for it in `execute` and early return, I don't think makes a big difference. Long term this would go away once rover is converted to a filter rather than an extract, but for the time being, I think trying to run rover without asking for output is something worth erroring and telling the user about.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -107,130 +107,394 @@ RoverXRay::verify_params(const conduit::Node &params,
...
+  if (n_rover.has_child("unit_scalar"))
   {
-    const std::string protocol = params["rover/blueprint"].as_string();
-    if (protocol != "hdf5" && protocol != "yaml" && protocol != "json")
+    if (!n_rover["unit_scalar"].dtype().is_number())
+    {
+      info["errors"].append() = "Optional numeric parameter 'rover/unit_scalar' is not numeric";
+      res = false;
+    }
+    else // (n_rover["unit_scalar"].dtype().is_number()
```
siramok, line 338: Whoops, missed a closing `)`.

src/tests/ascent/t_ascent_rover.cpp
JustinPrivitera: It would be good to output a variety of output types (yaml, json, hdf5). Just to put everything through its paces. You could have every test do all three. I leave the choices up to you. But it is best to test every option, since that is likely how the png output being broken was never noticed for rover.

src/tests/ascent/t_ascent_rover.cpp
```diff
@@ -794,6 +697,8 @@ TEST(ascent_rover, test_xray_blueprint_curv3d_camera_params)
     ASCENT_ACTIONS_DUMP(actions, image_path, msg);
 }
 
+// TODO: Add a test for imaging planes
```
JustinPrivitera, line 677: Are imaging planes in this PR? Is it time to add the test for them?

src/libs/vtkh/rendering/PartialComposite.hpp
```diff
@@ -27,9 +27,9 @@ struct VTKH_API PartialComposite
 {
   vtkm::cont::ArrayHandle<vtkm::Id> PixelIds;   // pixel that owns composite
   vtkm::cont::ArrayHandle<FloatType> Distances; // distance of composite end
-  vtkm::rendering::raytracing::ChannelBuffer<FloatType> Buffer;              // holds either color or absorption
-  vtkm::rendering::raytracing::ChannelBuffer<FloatType> Intensities;           // holds the intensity emerging from each ray
-  vtkm::rendering::raytracing::ChannelBuffer<FloatType> OpticalDepths;
+  vtkm::rendering::raytracing::ChannelBuffer<FloatType> Transmission;              // holds either color or absorption
+  vtkm::rendering::raytracing::ChannelBuffer<FloatType> Intensity;           // holds the intensity emerging from each ray
+  vtkm::rendering::raytracing::ChannelBuffer<FloatType> OpticalDepth;
```
JustinPrivitera, line 32: I like this

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -107,130 +107,394 @@ RoverXRay::verify_params(const conduit::Node &params,
...
+      info["errors"].append() = "Expected numeric parameter 'image_params/max_value' is not numeric";
+      res = false;
+    }
+  }
+
+  //
+  // Surprise check
+  //
+  
+  const std::vector<std::string> valid_paths = {
```
siramok, line 405: This is certainly above and beyond, but my bias is to make well-defined and explicit user-facing interfaces that will complain if anything is out of place. I find it nice for things like finding mispelled params and also reduces the surface area for potential bugs.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -107,130 +107,394 @@ RoverXRay::verify_params(const conduit::Node &params,
...
+    "rover/filename",
+    "rover/height",
+    "rover/png_output",
+    "rover/precision",
+    "rover/width",
+    "rover/unit_scalar"
+  };
+
+  const std::string surprises = surprise_check(valid_paths, params);
+  if ("" != surprises)
```
siramok: This should become `surprises.empty()`.

src/libs/vtkh/rendering/ConnectivityTracer.cpp
JustinPrivitera: I think we need to allow users to make the most horrible mistakes in this case. If they want to divide emission by absorption, then they can suffer the consequences, whatever their absorption field might be. The issue with your error check is that it's opaque to the user. They have no way of knowing if the division happened for sure or if their absorption was non-positive. I think we decided in VisIt to let users make this mistake if they are asking for it. So I think we should remove the `if (absorb > 0)` guard.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -240,6 +504,19 @@ RoverXRay::verify_params(const conduit::Node &params,
 void
 RoverXRay::execute()
 {
+  // MPI
+  int mpi_comm_id = -1;
+  int mpi_rank = 0;
+#ifdef ASCENT_MPI_ENABLED
```
siramok, line 449: The purpose for moving the MPI init stuff to the top of `execute` was because it's relevant to dataset validation. The dataset will report as being empty for all ranks except rank 0, hence why we need to know our MPI rank beforehand.

src/libs/rover/settings.hpp
```diff
@@ -13,6 +13,7 @@ using namespace conduit;
 
 namespace rover
 {
+  extern Node metadata;
   extern Node settings;
```
JustinPrivitera, line 17: what is metadata for versus settings?

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -240,6 +504,19 @@ RoverXRay::verify_params(const conduit::Node &params,
...
+  mpi_comm_id = flow::Workspace::default_mpi_comm();
+  rover::Logger::get_instance()->set_mpi_comm_id(mpi_comm_id);
+  rover::DataLogger::GetInstance()->set_mpi_comm_id(mpi_comm_id);
+  MPI_Comm_rank(MPI_Comm_f2c(mpi_comm_id), &mpi_rank);
+#endif
+
+  const conduit::Node &n_params = params();
```
siramok, line 456: Organizational change, no real purpose besides I felt it was a little nicer to cache `params()` as a variable rather than calling it twice directly. Plus separating out `n_rover` makes it simpler to remove `rover/` from all of the settings paths later on.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -250,75 +527,104 @@ RoverXRay::execute()
...
   {
-    ASCENT_ERROR("Absorption field name '" << absorption << "' is not in the dataset");
+    ASCENT_ERROR("The dataset does not have a field called '" << absorption << "'");
   }
 
-  // Fetch the dataset associated with the 'absorption' field
-  std::string topo_name = collection->field_topology(absorption);
-  // TODO: Validate topo_name
+  // Validate that the 'absorption' field has a topology
+  const std::string topo_name = collection->field_topology(absorption);
```
siramok, line 480: I need to double check that this isn't also something that always returns true for ranks other than 0.

src/libs/rover/rover.cpp
JustinPrivitera: I want to understand better how it works and what info is in metadata and what info is in settings. Perhaps I have more reviewing to do, but we can also discuss this afternoon.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -250,75 +527,104 @@ RoverXRay::execute()
...
-  if (params().has_path("rover/bov_filename"))
-  {
-    std::string bov_filename = params()["rover/bov_filename"].as_string();
-    bov_filename = output_dir(bov_filename);
-    rover.save_bov(expand_path_special_variables(bov_filename, ".bov", mpi_comm_id));
+  //
+  // Outputs
+  //
+
+  const std::string filename = n_rover["filename"].as_string();
```
siramok, line 521: As mentioned, I figure it's safe to reuse the same filename for all outputs since the file extensions are all unique anyhow.

src/libs/rover/rover.cpp
JustinPrivitera: great

src/libs/rover/rover.cpp
JustinPrivitera: awesome

src/libs/rover/rover.cpp
JustinPrivitera: What are the image_params? They're only relevant for png?

src/libs/rover/rover.cpp
JustinPrivitera: What do you mean by automatically tighten?

src/libs/rover/rover.cpp
JustinPrivitera: good

src/libs/rover/rover.cpp
JustinPrivitera: cool

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -250,75 +527,104 @@ RoverXRay::execute()
...
+    {
+      const std::string bov_filename = output_dir(expand_path_special_variables(
+                                                    filename,
+                                                    ".bov",
+                                                    mpi_comm_id));
+      rover.save_bov(bov_filename);
+    }
+  }
+
+  if (n_rover.has_child("png_output"))
```
siramok: Technically supposed to be "faster" but I just want to note that I've tried (pretty successfully) to avoid `.has_path` everywhere in favor of `.has_child`. Even if performance was exactly the same, I find use one-word paths to be nice because it forces intentionality.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -250,75 +527,104 @@ RoverXRay::execute()
...
+  {
+    const std::string png_output = n_rover["png_output"].as_string();
+    if ("true" == png_output)
+    {
+      ASCENT_WARN("Rover's png output is currently broken\n");
```
siramok: Here's that warning we discussed.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
siramok: Seems to not matter, so I'll leave it be.

src/libs/rover/partial_image.hpp
JustinPrivitera: let's ask @cyrush about the name to make sure.

src/libs/rover/partial_image.hpp
```diff
@@ -310,38 +310,37 @@ struct PartialImage
...
         std::cout<<"\n";
       }
     }
+  }
 
-  }// print
-
-  // TODO: Investigate if we need this
+  // For debugging
   void make_red_pixel(const int x, const int y)
```
JustinPrivitera: We need make_red_pixel for debugging?

src/libs/rover/ray_generators/ray_generator.hpp
JustinPrivitera: That's a good idea, assuming all works correctly in practice.

src/libs/rover/rover.cpp
siramok: It was there before me (and may also be a good candidate for removal once rover becomes a filter rather than an extract), but its only purpose is [to control aspects of the .png output](https://github.com/Alpine-DAV/ascent/blob/d8e3f33b3a183648a657c794b139b841e52dd159/src/libs/rover/typed_scheduler.cpp#L454).

src/libs/rover/rover.cpp
siramok: Good question. If you take a look at [the vtkm camera source](https://github.com/Kitware/VTK-m/blob/2cd5a5e168124787192f4f857e7e7df58793a67a/vtkm/rendering/Camera.cxx#L305) for `ResetToBounds`, the near and far planes are automatically set like so: `SetClippingRange(0.1f * diagonalLength, diagonalLength * 10.0f);`.

In the case of curv3d, this sets a far plane of ~330 resulting in a very large imaging plane placed way behind the mesh. Instead of using what vtkm resets us to by default, we could do some math to figure out "tighter" near and far plane values (e.g. ~50 works great for curv3d) that still encompass the mesh and result in better sized imaging planes. This would be done 1) after the call to `ResetToBounds` so that all of the other params still get set, and 2) using the bounding box to guarantee that we encompass the mesh rather than using different magic numbers (e.g. 0.5 and 5).

src/libs/vtkh/rendering/ConnectivityTracer.cpp
siramok: Sounds reasonable.

src/tests/_baseline_images/tout_rover_xray_blueprint_braid_intensities_spatial_000100.png
siramok: The formula for computing near, view, and far widths and heights changed, which indirectly affected how the spatial coordset gets computed. As to whether the new extents are correct, I think that's worth spending extra time to validate.

src/tests/_baseline_images/tout_rover_xray_blueprint_braid_rotated_intensities_000100.png
siramok: I have not but can add it to the list of things to validate.

src/tests/_baseline_images/tout_rover_xray_blueprint_multi_curv3d_intensities_000048.png
siramok: I have not but can add it to the list of things to validate.

src/tests/ascent/t_ascent_rover.cpp
siramok: Good point actually, I agree with you. I think it would be nice to use each protocol for at least 2 datasets each.

src/tests/_baseline_images/tout_rover_xray_blueprint_multi_curv3d_optical_depth_000048.png
siramok: After this PR is finished, my plan is to focus all my efforts onto solving multi-domain next.

src/tests/ascent/t_ascent_rover.cpp
siramok: I see your point. I originally did this because any difference in where ascent gets built is enough to cause the test to fail (e.g. different CI runs using different environments), but I see the value in fixing this at the other end (modifying the baseline node) so that we do still test for filename. Thanks for pointing that out.

src/tests/ascent/t_ascent_rover.cpp
siramok: Yes, they are in this PR. I would say that yes, it's time to add a test for them, however my naive attempt at rendering all of the imaging plane fields resulted in an empty scene. I have confidence that the meshes are defined correctly (don't want to be too definitive without tests) because I'm able to open the output in VisIt and see the imaging planes + original mesh, which looks good. Ascent did still produce an image, although it's just a black void with all of the individual field legends. I need to spend time digging into what's happening there and didn't want that to hold up review on the rest of the implementation.

src/libs/rover/settings.hpp
siramok: In reality, nothing. In practice, it makes life ever so slightly nicer after making the other changes to remove `rover/` from the settings queries. Time and cycle aren't really "settings" anyways, which are how I further justified separating them out. It only gets used for one thing: plumbing metadata into the blueprint output.

src/libs/rover/partial_image.hpp
siramok: I don't personally need it for debugging, but it seems like that's what Matt wrote it for. I'm not opposed to just ripping it out entirely.

siramok:
Meta-note: the CI failures seem to be unrelated to this PR.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
JustinPrivitera: They don't all have different file extensions. I think each of the Blueprint output types (json, yaml, hdf5) will all use `.root` files.

Let's turn off png entirely. We should have `hdf5` be the default. If users set a different output type we can use that instead.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -107,130 +107,394 @@ RoverXRay::verify_params(const conduit::Node &params,
...
+  if (has_blueprint_output && !has_blueprint_protocol)
+  {
+    info["errors"].append() = "Optional bool string parameter 'rover/blueprint_output' requires 'rover/blueprint_protocol' to also be set";
+    res = false;
+  }
+  else if (!has_blueprint_output && has_blueprint_protocol)
+  {
+    info["errors"].append() = "Optional string parameter 'rover/blueprint_protocol' requires 'rover/blueprint_output' to also be set";
+    res = false;
+  }
```
JustinPrivitera: we could condense these perhaps:

`if (! has_blueprint_output || ! has_blueprint_protocol)`

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
JustinPrivitera: I agree that we should throw an error in the no output case. But if we have hdf5 be the default then we never hit this.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
JustinPrivitera: 🫨

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
JustinPrivitera: cool

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
JustinPrivitera: what does `params()` return? Does it return a reference?
```c++
const conduit::Node& params();
```
versus
```c++
const conduit::Node params();
```

If it is the former then we get a performance benefit. If it is the latter then there is no difference between
```c++
const conduit::Node &n_params = params();
```
and
```c++
const conduit::Node n_params = params();
```
so we should do the 2nd version for clarity.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
JustinPrivitera: Just so it's here too, I don't think the file extensions are unique.

Here is the proof: https://github.com/visit-dav/visit/blob/2fa90c9dde96dafbb2030d2800a5c3bf54d562dd/src/avt/Queries/Queries/avtXRayImageQuery.C#L82-L86

JustinPrivitera: Let's have there be one output type that is chosen - not blueprint and/or png and/or bov, but instead, hdf5 blueprint, json bluepring, yaml blueprint, png, or bov, and not more than one.

JustinPrivitera: I think I understand where you're at now - you won't have blueprint (hdf5, yaml, json) extensions compete with bov or png. That is true. But I still think it would be better to have a single output type that you choose. That will simplify this a bit.

JustinPrivitera: hmmm

JustinPrivitera: we can always turn on multiple outputs later if we really want to

JustinPrivitera: but the way to do that would be to let users pick from

- hdf5
- yaml
- json
- png
- bov

not how it works now, which is

 - blueprint (pick between hdf5, yaml, json)
 - png
 - bov

Hopefully that makes sense.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
JustinPrivitera: sounds good

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
JustinPrivitera: great

src/libs/rover/image.cpp
JustinPrivitera: I like lambdas

src/libs/rover/image.cpp
```diff
@@ -173,104 +173,98 @@ template<typename FloatType>
...
+                                       auto& channel_group,
+                                       auto default_value,
+                                       auto& output_vector) {
```
JustinPrivitera: Why use auto here?

src/libs/rover/typed_scheduler.cpp
```diff
@@ -34,12 +35,15 @@ TypedScheduler<FloatType>::set_comm_handle(MPI_Comm comm_handle)
...
-  ROVER_INFO("TypedScheduler::add_dataset: adding domain " << m_domains.size());
-  Domain domain;
-  domain.set_dataset(dataset);
-  m_domains.push_back(domain);
+  const int num_domains = dataset.GetNumberOfDomains();
+  m_domains.reserve(num_domains);
+  for (int i = 0; i < num_domains; i++)
+  {
+    ROVER_INFO("TypedScheduler::add_dataset: adding domain " << m_domains.size());
+    m_domains.emplace_back(dataset.GetDomain(i));
```
JustinPrivitera, line 45: is there any reason to use emplace_back here? We are not creating the domain, just fetching it. Does this actually have a performance benefit? Though it probably doesn't have any drawback.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
siramok: Gotcha. A solution could be to append the protocol type to the filename for blueprint outputs? E.g. "my_filename" + "_yaml" + ".root".

src/libs/rover/typed_scheduler.cpp
```diff
@@ -51,43 +55,21 @@ TypedScheduler<FloatType>::set_ray_generator(RayGenerator *ray_generator)
...
-{
-  // TODO: See if this is done better with std::transform
-  const size_t size = background.size();
-  m_background.resize(size);
-
-  for (size_t i = 0; i < size; ++i)
-  {
-    m_background[i] = static_cast<vtkm::Float64>(background[i]);
-  }
-}
```
JustinPrivitera, line 73: we don't need this method?

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
siramok: Not sure, but worth looking into. I would hope that it returns a reference, because it gets used all over ascent filters.

src/libs/rover/image.cpp
siramok: No particular reason. I can see what it's like with the real types.

src/libs/rover/typed_scheduler.cpp
```diff
@@ -216,29 +198,69 @@ TypedScheduler<FloatType>::set_global_range_and_bounds()
...
-  // probably be improved
+  const std::string emission = rover::settings["emission"].as_string();
+  if (!emission.empty())
+  {
+    typed_composite<vtkh::EmissionPartial<FloatType>>();
+  }
+  else // (emission.empty())
+  {
+    typed_composite<vtkh::AbsorptionPartial<FloatType>>();
+  }
```
JustinPrivitera, line 226: I still don't love these names... feels like it should be EmissionPartial -> AbsorptionAndEmissionPartial and AbsorptionPartial -> OnlyAbsorptionPartial.

But I don't know why we need to have these two cases totally separate anyway. I'm sure there are reasons.

src/libs/rover/typed_scheduler.cpp
siramok: I can't comment on performance because I haven't measured it, but since we know how many domains are in our dataset ahead of time, I think I added a `m_domains.reserve(num_domains)` somewhere. Then `emplace_back` should just construct `Domain` instances in-place using the reserved memory?

src/libs/rover/typed_scheduler.cpp
```diff
@@ -580,39 +535,321 @@ void TypedScheduler<FloatType>::save_bov(std::string file_name)
...
-  const double detector_width = 2.0 * near_width;
-  const double detector_height = 2.0 * near_height;
-  const double far_detector_width = 2.0 * far_width;
-  const double far_detector_height = 2.0 * far_height;
-  const double near_dx = detector_width / width;
-  const double near_dy = detector_height / height;
+  // These calculations are based on VisIt's perspective calculations, but diverge
+  // due to differences in how near and far planes are represented. VisIt represents
+  // them as offsets from the view plane (e.g. -0.5, 0.5), while vtkm represents them
+  // as positive distances away from the camera position (e.g. 3, 300).
```
JustinPrivitera, line 819: Interesting. There is a lot to think about.

src/libs/rover/typed_scheduler.cpp
siramok: Not currently. Right now we construct and use exactly one background during the lifetime of each `rover.execute()` call. I suspect that this type of setter would have been useful in the context of a standalone library.

src/libs/rover/typed_scheduler.cpp
```diff
@@ -580,39 +535,321 @@ void TypedScheduler<FloatType>::save_bov(std::string file_name)
...
+
+  vtkmVec3f llc_near;
+  vtkmVec3f lrc_near;
+  vtkmVec3f ulc_near;
+  vtkmVec3f urc_near;
+
+  vtkmVec3f llc_view;
+  vtkmVec3f lrc_view;
+  vtkmVec3f ulc_view;
+  vtkmVec3f urc_view;
```
JustinPrivitera, line 849: so are we creating a mesh for the near plane?

src/libs/rover/typed_scheduler.cpp
siramok: Yeah, to be fair I didn't make any attempt to simplify `EmissionPartial` vs `AbsorptionPartial` in this PR. They almost certainly can be combined IMO.

src/libs/rover/typed_scheduler.cpp
```diff
@@ -625,34 +862,35 @@ TypedScheduler<FloatType>::to_blueprint(Node &data)
...
 
   Node &xray_query = state["xray_query"];
-  xray_query.set(rover::settings["rover"]);
+  xray_query.set(rover::settings);
 
   Node &xray_data = state["xray_data"];
-  xray_data["detector_width"] = detector_width;
-  xray_data["detector_height"] = detector_height;
+  xray_data["detector_width"] = detector_width * 10.0f; // TODO: Needs validation against VisIt
+  xray_data["detector_height"] = detector_height * 10.0f; // TODO: Needs validation against VisIt
```
JustinPrivitera: times 10???

src/libs/rover/typed_scheduler.cpp
siramok: Yes, a separate mesh for the near, image, and far planes.

src/libs/rover/typed_scheduler.cpp
```diff
@@ -798,11 +1034,89 @@ TypedScheduler<FloatType>::to_blueprint(Node &data)
...
+                                  view_width,
+                                  view_height,
+                                  center_view,
+                                  left,
+                                  up,
+                                  llc_view,
+                                  lrc_view,
+                                  ulc_view,
+                                  urc_view);
+
```
JustinPrivitera, line 1078: What is this? Does view work the same as it does in VisIt? We should make tests that render the planes to understand.

src/libs/rover/typed_scheduler.cpp
siramok: Yeah, I'm not sure what's correct here. To have parity with the old detector width and height values, this should actually be `* 20`. Hence the note about validating against VisIt.

JustinPrivitera (changes requested):
Lots of great stuff. We should discuss.

src/libs/rover/typed_scheduler.cpp
siramok: Yep, more or less identical to VisIt, just decided to change the name from `_image` to `_view` for consistency.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
siramok: Will look into it.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
siramok: Let's discuss IRL.

siramok: I see what you're saying. I'm fine with that. We could also move to a single param, but support a comma separated list of output types...

src/libs/rover/rover.cpp
siramok: Emily commented that this might be called something else to not be confused with ascent's metadata node.

src/libs/rover/rover.cpp
siramok: We decided this is a good idea.

src/libs/vtkh/rendering/ConnectivityTracer.cpp
siramok: We'll make this change and also add an ascent warning about abnormal absorb value. Maybe the check can be pulled out of the loop also for performance.

src/tests/_baseline_images/tout_rover_xray_blueprint_braid_intensities_spatial_000100.png
siramok: We decided this needs to be investigated and compared with visit.

siramok: Spatial extent mesh should match the dimensions of the near plane.

src/tests/ascent/t_ascent_rover.cpp
siramok: Another way to do this could be to have a loop that does all 3 (but that might be slow).

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
siramok: We decided to go with a single variable, not a comma separated list.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
siramok: I agree. Will make this change.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
siramok: This is correct.

src/libs/rover/partial_image.hpp
siramok: We decided to rip it out.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
siramok: This becomes irrelevant after the other options change.

src/libs/rover/image.cpp
siramok: We decided to put the real types in.

src/libs/rover/typed_scheduler.cpp
siramok: These should match the near plane dimensions.

siramok: Don't scale anything here.

src/libs/vtkh/rendering/ConnectivityTracer.cpp
siramok: Fixed. We no longer guard against `absorb == 0`, and the check for dividing emission by absorption now only happens once.

src/tests/_baseline_images/tout_rover_xray_blueprint_braid_intensities_spatial_000100.png
siramok: Fixed (still needs validation with visit). Spatial extent meshes match the near plane dimensions again.

src/tests/ascent/t_ascent_rover.cpp
siramok: Fixed. All 3 blueprint types are used in the tests now, although using a loop to do each output type for each test can still be on the table for later.

src/tests/ascent/t_ascent_rover.cpp
siramok: Fixed. We now add filename to the baseline node, rather than removing it from the output node.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
siramok: Fixed. Now there's just one param that controls the output type, and it uses hdf5 by default.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
siramok: Fixed. Now there's just one param that controls the output type, and it uses hdf5 by default.

src/libs/rover/image.cpp
siramok: Fixed. Real types were added here instead of auto.

src/libs/rover/partial_image.hpp
siramok: Fixed. It's ripped out.

src/libs/rover/typed_scheduler.cpp
siramok: Added a TODO to combine these later.

src/libs/rover/typed_scheduler.cpp
siramok: Fixed. Uses the near plane dimensions and no longer has scaling.

src/tests/_baseline_images/tout_rover_xray_blueprint_braid_rotated_intensities_000100.png
siramok: Comparing to visit, rover's output for the same background intensity looks correct to me.

src/tests/_baseline_images/tout_rover_xray_blueprint_multi_curv3d_intensities_000048.png
siramok: I compared to visit, it looks right to me.

JustinPrivitera: the min and max intensity values looked the same?

src/libs/rover/partial_image.hpp
siramok: We decided that transmission is fine.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -572,59 +511,51 @@ RoverXRay::execute()
...
+                                                      mpi_comm_id));
+    const int num_files = -1;
+    conduit::Node extra_opts;
+    std::string result_path;
+    mesh_blueprint_save(multi_domain,
+                        blueprint_filename,
+                        output_type,
+                        num_files,
+                        extra_opts,
+                        result_path);
```
JustinPrivitera: I am asking @cyrush about this; there may be a way to generate the output filename that is more robust.

src/libs/rover/image.cpp
```diff
@@ -9,6 +9,7 @@
 #include <rover_exceptions.hpp>
 #include <utils/rover_logging.hpp>
 
+#include <vector>
```
JustinPrivitera: did we need to include vector?

src/libs/rover/partial_image.hpp
```diff
@@ -292,58 +291,6 @@ struct PartialImage
...
-    }
-  }
-
-  // For debugging
-  void make_red_pixel(const int x, const int y)
-  {
-    const int size = m_pixel_ids.GetNumberOfValues();
-    const int num_channels = m_transmission.GetNumChannels();
-    const int64 width = rover::settings["width"].to_int64();
-    const int64 height = rover::settings["height"].to_int64();
```
JustinPrivitera: was there also a `make_green_pixel` and `make_blue_pixel`?

siramok: There were not.

src/libs/rover/image.cpp
siramok: I just checked, we do not need to include vector.

src/libs/vtkh/rendering/ConnectivityTracer.cpp
```diff
@@ -737,75 +737,91 @@ class IntegrateEmission : public vtkm::worklet::WorkletMapField
...
+    // that makes it out of this mesh segment. If this is really run on a single node,
+    // we can get the final energy value by multiplying the background intensity by
+    // the total absorption of the mesh segment and add in the amount of emitted
+    // energy that escapes.
+    //
+
+    // NumBins can potentially be a very large number, so the loops are duplicated
+    // like this to avoid checking if DivideEmisByAbsorb == true each iteration.
+
+    if (DivideEmisByAbsorb)
```
JustinPrivitera, line 765: glad you pulled this out to the top level - there may be a way to refactor such that we don't have so much duplicated code in the two cases.

src/tests/ascent/t_ascent_rover.cpp
JustinPrivitera: we can add the tests later

src/libs/rover/image.cpp
siramok: Fixed. Just removed it.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
siramok: Any update on this?

JustinPrivitera: This is good the way it is. Thanks for your patience.

JustinPrivitera (approved):

PR #1574, opened 2025-07-08
Move verbiage in LICENSE around and update it to conform with GitHub format

markcmiller86:
These changes will cause GitHub to recognize Ascent as a BSD 3-Clause project.


cyrush (approved):
thanks @markcmiller86 !

PR #1578, opened 2025-07-11
Add compositing for Rover's optical depth output

siramok:
Addresses bullets in #1513

- Optical depth now gets composited along with intensity, fixing the multi-domain case.
- Some cleanup performed along the way.


src/libs/vtkh/compositing/EmissionPartial.hpp
```diff
@@ -103,41 +103,42 @@ struct EmissionPartial
...
     }
   }
 
+  inline void blend_optical_depth(const EmissionPartial<FloatType> &other)
+  {
+    const int num_bins = static_cast<int>(m_optical_depth_bins.size());
+    assert(num_bins == (int)other.m_optical_depth_bins.size());
+    for (int i = 0; i < num_bins; i++)
+    {
+      m_optical_depth_bins[i] += other.m_optical_depth_bins[i];
```
JustinPrivitera, line 128: how do we know += is the right answer? I see up above that emission bins uses *= and add emission below uses +=.

src/libs/vtkh/compositing/PartialCompositor.cpp
```diff
@@ -102,6 +103,9 @@ BlendEmission(const int &total_segments,
               std::vector<EmissionPartial<T>> &output_partials,
               const int output_offset)
 {
+  // TODO: Compare how optical depth is computed here vs how
+  // it's computed in connectivity tracer
+
```
JustinPrivitera, line 108: we can add a new field output temporarily to the blueprint output called "bonus" or something and then render both side by side

src/tests/_baseline_images/tout_rover_xray_blueprint_curv3d_rotated_optical_depth_000048.png
JustinPrivitera, line 1: This one looks weird. I'd like to see what VisIt does from the same camera angle.

src/tests/_baseline_images/tout_rover_xray_blueprint_multi_curv3d_optical_depth_000048.png
JustinPrivitera, line 1: fantastic

JustinPrivitera (commented):
Maybe we could follow up on the bonus optical depths and check the one camera angle with VisIt, then this should be good to go.

src/tests/_baseline_images/tout_rover_xray_blueprint_curv3d_rotated_optical_depth_000048.png
JustinPrivitera: We should have all the tools in our toolbox to set up an identical query with VisIt with all camera parameters matching.

JustinPrivitera (approved):
Awesome

src/libs/vtkh/compositing/EmissionPartial.hpp
siramok: We discussed this on teams. The current output with the latest fix applied looks reasonably close to visit's output. As far as why we do += to blend rather than *=, my intuition is that we want to add each ray segment's length per-pixel per-partial image to end up with the total distance that the ray spent travelling through the mesh.

src/libs/vtkh/compositing/PartialCompositor.cpp
siramok: We discussed this on teams. The output looked "like garbage" and changing the way it's blended to use += rather than *= produced something slightly better, but still completely off. We've decided that the current approach of computing intensity and transmission separately should just be combined within vtkm so that we compute the full intensity values.

PR #1579, opened 2025-07-11
per scene default bounds using only relevent topos

cyrush:
update scene bounds calc to use union of topologies that are used in scene plots


cyrush:
Screenshot 2025-07-14 at 9 47 40 AM

now tracking per topo bounds -- adding final scene as input to create_renders, now we have complete info at the time we setup the renders (and camera params)

cyrush:
before:
Screenshot 2025-07-14 at 9 55 00 AM

after:
Screenshot 2025-07-14 at 9 54 02 AM

cyrush:
@siramok  @JustinPrivitera close to a solution for #1559

JustinPrivitera:
Excellent; thanks so much for figuring this out!

siramok:
Nice, thanks for working on this.

src/libs/ascent/runtimes/ascent_main_runtime.cpp
```diff
@@ -1638,7 +1638,7 @@ AscentRuntime::CreateScenes(const conduit::Node &scenes)
 
     std::string renders_name = names[i] + "_renders";
 
-    if(!flow::Workspace::supports_filter_type("default_render"))
+    if(!flow::Workspace::supports_filter_type("create_renders"))
```
cyrush, line 1641: `default_render` wasn't a great name, `create_renders` conveys the purpose better

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rendering_filters.cpp
```diff
@@ -1031,30 +1080,31 @@ std::map<std::string, CinemaManager> CinemaDatabases::m_databases;
 //-----------------------------------------------------------------------------
 
 //-----------------------------------------------------------------------------
-DefaultRender::DefaultRender()
+CreateRenders::CreateRenders()
```
cyrush, line 1083: used better name

src/tests/_baseline_images/tout_render_2d_poly_and_nopolypolytess_000000.png
cyrush, line 1: the extends change is correct. 

Before, it was also looking at a dataset that also had a braid topo, which has slightly larger extents

src/tests/ascent/t_ascent_render_2d.cpp
```diff
@@ -406,6 +406,121 @@ TEST(ascent_render_2d, test_render_2d_uniform_render_serial_backend)
...
+
+    scenes["s2/plots/p2/type"]  = "pseudocolor";
+    scenes["s2/plots/p2/field"] = "field_big";
+    scenes["s2/image_prefix"] = output_fbig;
+
+
+    scenes["s2/plots/p3/type"]  = "pseudocolor";
+    scenes["s2/plots/p3/field"] = "field_big";
+    scenes["s2/image_prefix"] = output_fbig;
+
```
nicolemarsaglia, line 506: multiple plots of field_big on the same scene? either copy/pasta or I'm missing the point.

cyrush: This is strange, but I did this to test the data flow graph cases in play.

Scene with a single plot does not use the union bounds filter
Scene with multiple plots tests the union bounds filter

nicolemarsaglia (approved):
I'm sure this fix will be much appreciated!

siramok:
I checked out this branch, removed the box clipping workaround we currently do for the rover tests, and then ran the tests. All of the tests still pass, so this does nicely fix the original issue.

cyrush:
Nice, thanks for that additional vetting of these changes

-Cyrus
________________________________

PR #1581, opened 2025-07-15
Remove Rover tests box clip workaround

siramok:
Addresses bullets in #1513
- Removes the box clipping workaround for rover's tests now that #1559 is fixed.
- The "enable_imaging_planes" setting was no longer needed and has been reworked into "enable_rays_mesh".
- Updates Windows CI to upload test output as an artifact, since we don't have an easy way to re-run Windows CI locally.

Note: none of the baseline images needed to change, they are still 1-1 identical.

### DIFF
--- a/src/libs/rover/typed_scheduler.cpp
+++ b/src/libs/rover/typed_scheduler.cpp
@@ -1040,72 +1040,73 @@ TypedScheduler<FloatType>::to_blueprint(Node &data)
   optical_depth_spatial.set(optical_depth);
   optical_depth_spatial["topology"] = "spatial_topo";
 
-  // TODO: Remove this once issue #1559 is fixed
-  const bool enable_imaging_planes = rover::settings["enable_imaging_planes"].as_string() == "true";
-  if (enable_imaging_planes)
+  //
+  // Near plane mesh
+  //
+
+  write_blueprint_imaging_plane(data,
+                                "near_plane",
+                                near_width,
+                                near_height,
+                                center_near,
+                                left,
+                                up,
+                                llc_near,
+                                lrc_near,
+                                ulc_near,
+                                urc_near);
+
+  //
+  // View plane mesh
+  //
+
+  write_blueprint_imaging_plane(data,
+                                "view_plane",
+                                view_width,
+                                view_height,
+                                center_view,
+                                left,
+                                up,
+                                llc_view,
+                                lrc_view,
+                                ulc_view,
+                                urc_view);
+
+  //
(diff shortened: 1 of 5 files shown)


siramok:
My first instinct was to run this locally using docker, but it looks like azemu.py doesn't provide a launch Windows script. I suspect this PR is revealing a Windows-specific edge case with #1579. Perhaps @cyrush can take a look when you have some time?

There's no rush to get this merged. :)

cyrush:
That's correct. We don't have windows containers or a good way to emulate the windows case. I'll take a look.

cyrush:
@siramok  my windows build ran fine :-(

We need to publish the test results from github actions so we can take a look.

They should exist in:
```
build/tests/_output/
```

Can you look into how to publish artifacts with github actions?

src/libs/rover/typed_scheduler.cpp
```diff
@@ -1040,72 +1040,73 @@ TypedScheduler<FloatType>::to_blueprint(Node &data)
...
+                                   lrc_far,
+                                   urc_near,
+                                   urc_far,
+                                   ulc_near,
+                                   ulc_far);
+
+  // This mesh can be very large depending on the image width and height, and it won't
+  // always be useful. We only include it in the output if the user explicitly asks for it.
+  const bool enable_rays_mesh = rover::settings["enable_rays_mesh"].as_string() == "true";
+  if (enable_rays_mesh)
```
JustinPrivitera, line 1108: 👍

JustinPrivitera (commented):
Now we can add tests for the imaging planes and rays meshes since they won't be affected by the bounding box issue. That can be in a follow up. I am not approving since we want to figure out CI first.

siramok:
@cyrush Thanks, I'll look into how to publish artifacts with GitHub actions.

siramok:
@cyrush As soon as I added artifact uploading, the test decided to pass... I don't know what to make of that. Would you like me to undo that change before we merge?

cyrush:
> @cyrush As soon as I added artifact uploading, the test decided to pass... I don't know what to make of that. Would you like me to undo that change before we merge?

will still be good to have them for future mysteries, lets make sure they look sensible and we can keep the upload

siramok:
Sweet. I only added it for Windows, is that fine or do you want it added to all of CI?

siramok:
I downloaded the artifact and the output images LGTM.

cyrush:
Windows is fine for now

JustinPrivitera (approved):

PR #1588, opened 2025-07-23
Fix MPI support in Rover

siramok:
Addresses bullets in https://github.com/Alpine-DAV/ascent/issues/1513
- Fixes segfaults that appear when executing rover with MPI.
- Adds MPI tests for rover.


src/tests/ascent/t_ascent_mpi_rover.cpp
```diff
@@ -0,0 +1,223 @@
...
+    // TODO can we ask Ascent for the name of the file it wrote?
+    // std::cout << ascent.info().to_yaml() << std::endl;
+    ascent.close();
+}
+
+void
+load_and_verify_local_data(Node &data,
+                           const std::string &data_path)
+{
+    conduit::relay::io::blueprint::load_mesh(data_path, data);
```
JustinPrivitera:
```suggestion
    conduit::relay::mpi::io::blueprint::load_mesh(data_path, data);
```

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -447,8 +447,8 @@ RoverXRay::execute()
   int mpi_rank = 0;
 #ifdef ASCENT_MPI_ENABLED
   mpi_comm_id = flow::Workspace::default_mpi_comm();
-  rover::Logger::get_instance()->set_mpi_comm_id(mpi_comm_id);
-  rover::DataLogger::GetInstance()->set_mpi_comm_id(mpi_comm_id);
+  // rover::Logger::get_instance()->set_mpi_comm_id(mpi_comm_id);
```
siramok, line 450: I didn't investigate why rover's logger was segfaulting in the MPI case, but perhaps it's a good time to revisit the conversation about converting to use ascent's logging?

src/tests/_baseline_images/tout_rover_xray_blueprint_curv3d_rotated_intensities_spatial_000048.png
siramok, line 1: Must be a 1-pixel difference.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
JustinPrivitera: we should make a note in the big issue ticket

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -408,6 +408,7 @@ RoverXRay::verify_params(const conduit::Node &params,
     "camera/fov",
     "camera/look_at",
     "camera/near_plane",
+    "camera/position",
```
JustinPrivitera: what is camera position for?

siramok: It's a vtkm camera parameter, like if the user wants to specify an exact x-ray detector position. It used to be settable, but I guess I removed it at some point without meaning to. However, it doesn't need to be fixed in this PR, so I removed this and added a TODO instead.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
siramok: It's already listed there actually.

src/libs/rover/typed_scheduler.cpp
```diff
@@ -37,11 +38,12 @@ template<typename FloatType>
 void
 TypedScheduler<FloatType>::add_dataset(vtkh::DataSet &dataset)
 {
-  const int num_domains = dataset.GetNumberOfDomains();
-  m_domains.reserve(num_domains);
-  for (int i = 0; i < num_domains; i++)
+  // TODO: In cases where ascent is not given data with 1 domain per MPI rank, every rank gets every domain
+  m_num_local_domains = dataset.GetNumberOfDomains();
+  m_domains.reserve(m_num_local_domains);
```
JustinPrivitera: Every rank gets every domain? I don't think so. This isn't right.

JustinPrivitera: - single domain case: one rank gets the domain and the others have nothing and do nothing
- num domains == num ranks: each rank has one domain to work on
- num domains < num ranks: some ranks get domains and work on them, the rest have nothing and do nothing
- num domains > num ranks: some ranks get multiple domains and work on each of them in a loop

JustinPrivitera: whoever calls rover will control which ranks have which domains, but there should never be a case where a domain is shared between ranks (and certainly never a case where all domains are on all ranks.)

siramok: True, leftover from before we found the answer. I'll remove it.

JustinPrivitera: I guess I'm confused by the comment. Is the code already in line with what I've said above?

src/libs/rover/typed_scheduler.cpp
```diff
@@ -253,15 +254,15 @@ TypedScheduler<FloatType>::typed_composite()
 
   std::vector<PartialType> result;
   compositor.composite(partials, result);
```
JustinPrivitera, line 255: I am confused by this. Shouldn't we be doing the compositing only on rank 0? We should be pulling the partial images to rank 0 and then compositing? Maybe that is happening in the implementation of composite?

src/libs/rover/typed_scheduler.cpp
```diff
@@ -333,7 +333,7 @@ TypedScheduler<FloatType>::trace_rays()
   }
   vtkmRayTracing::Ray<FloatType> rays;
 
-  for (int i = 0; i < num_domains; ++i)
+  for (int i = 0; i < m_num_local_domains; i++)
```
JustinPrivitera, line 335: I guess I am confused by this. Why not just rename `num_domains` to `local_num_domains`? If we want to make it a member that's ok, but the value shouldn't be computed any differently than before. Is it computed the same way?

src/libs/vtkh/compositing/vtkh_diy_partial_blocks.hpp
```diff
@@ -172,6 +172,7 @@ struct Serialization<vtkh::EmissionPartial<double>>
   {
     vtkhdiy::save(bb, partial.m_bins);
     vtkhdiy::save(bb, partial.m_emission_bins);
+    vtkhdiy::save(bb, partial.m_optical_depth_bins);
```
JustinPrivitera, line 175: Just curious, but this change seems like something that would have had to happen when we got optical depths working. What was this change needed for in this PR?

src/tests/ascent/t_ascent_rover.cpp
```diff
@@ -132,6 +133,35 @@ get_valid_test_data(Node &data)
...
+        {
+            rank_vals[j] = static_cast<float64>(i);
+        }
+
+        data.append().set(domain);
+    }
+
+    Node verify_info;
+    EXPECT_TRUE(conduit::blueprint::mesh::verify(data, verify_info));
+}
```
JustinPrivitera, line 163: I wonder if we can refactor any of this stuff out to a shared header file that both the serial tests and mpi tests can draw from. The body of the for loop here looks like it is identical. I imagine there are other helpers for which we can refactor as well.

src/tests/ascent/t_ascent_rover.cpp
```diff
@@ -319,6 +349,237 @@ TEST(ascent_rover, test_xray_blueprint_braid_rotated)
...
+            detector_height: 5.65685440236809
+            intensity_max: 213.735064037837
+            intensity_min: 0.0
+            optical_depth_max: 3683.56120526528
+            optical_depth_min: 0.0
+            image_topo_order_of_domain_variables: "xyz"
+        domain_id: 0
+    )yaml";
+
+    Node baseline_data;
```
JustinPrivitera, line 440: I wish we could do the yaml differencing in a helper. A bunch of these have the exact same yaml so it shouldn't be hard to refactor out, but we could go a step further and generate the correct yaml for each test like we do in the visit tests:

https://github.com/visit-dav/visit/blob/84169e780964ebabbddc976a23ba40054c7d5dc9/src/test/tests/queries/xrayimage.py#L564-L573

We are passing in the correct expected values to the strings that then are parsed as conduit nodes for the diff. It should be possible to do the same pattern in C++.

src/tests/ascent/t_ascent_mpi_rover.cpp
```diff
@@ -0,0 +1,418 @@
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+// Copyright (c) Lawrence Livermore National Security, LLC and other Ascent
+// Project developers. See top-level LICENSE AND COPYRIGHT files for dates and
+// other details. No copyright assignment is required to contribute to Ascent.
+//~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+
+//-----------------------------------------------------------------------------
+///
+/// file: t_ascent_render_3d.cpp
```
JustinPrivitera: this should be changed

src/tests/ascent/t_ascent_mpi_rover.cpp
```diff
@@ -0,0 +1,418 @@
...
+const MPI_Comm COMM = MPI_COMM_WORLD;
+int par_rank = 0;
+int par_size = 1;
+
+//
+// Utilities
+//
+
+void
+execute_ascent(const Node &data,
```
JustinPrivitera, line 42: this could probably live in a shared header file and then use #ifdefs to handle the parallel stuff

src/tests/ascent/t_ascent_mpi_rover.cpp
```diff
@@ -0,0 +1,418 @@
...
+    ascent.open(ascent_opts);
+    ascent.publish(data);
+    ascent.execute(actions);
+    // TODO can we ask Ascent for the name of the file it wrote?
+    // std::cout << ascent.info().to_yaml() << std::endl;
+    ascent.close();
+}
+
+void
+render_blueprint(const string &field_name,
```
JustinPrivitera, line 59: same here

src/tests/ascent/t_ascent_mpi_rover.cpp
```diff
@@ -0,0 +1,418 @@
...
+    Node &add_plots = actions.append();
+    add_plots["action"] = "add_scenes";
+    add_plots["scenes"] = scenes;
+
+    // Execute Ascent actions
+    execute_ascent(data, actions);
+}
+
+void
+render_all_fields(const Node &data,
```
JustinPrivitera, line 87: and here

src/tests/ascent/t_ascent_mpi_rover.cpp
```diff
@@ -0,0 +1,418 @@
...
+                           const std::string &data_path)
+{
+    conduit::relay::mpi::io::blueprint::load_mesh(data_path, data, COMM);
+
+    Node verify_info;
+    EXPECT_TRUE(conduit::blueprint::mpi::mesh::verify(data, verify_info, COMM));
+}
+
+const bool
+is_vtkm_disabled()
```
JustinPrivitera, line 128: and this guy

JustinPrivitera (changes requested):
I left a number of comments and some clarifying questions.

src/libs/rover/typed_scheduler.cpp
siramok: I'm sorry, I didn't refresh the page before commenting and initially missed your 2nd and 3rd messages.

I added this TODO before we figured out that I was using `conduit::blueprint::mesh::examples::braid` instead of `blueprint::mpi::mesh::examples::braid_uniform_multi_domain`. Using the non-MPI API was causing each rank to publish an identical but separate 1-domain braid mesh, which seems to be a bad/unsupported case (from ascent's perspective, rover was still tracing and compositing the partial images just fine). Ascent expects each rank to publish its own domain data, and critically, **for each domain to belong to the same mesh**. The second piece is where I had messed up.

Basically, my input was wrong, but my understanding at the time was that I was only publishing 1 domain and 1 mesh in total. In reality, I was publishing 2 domains total from 2 separate meshes, one per rank, hence why I thought there was a bug to solve here. Since it was a misunderstanding on my part, I think this TODO can safely go away.

To answer your other questions, rover is designed for each MPI rank to iterate over each of its local domains, individually ray trace them, and then globally composite all of the partial images together. Rover doesn't manage which MPI ranks get which domains, but rather that comes in predetermined by ascent, and so rover just assumes that the partitioning is correct and what the user wants. The `num_domains == num_ranks` case is what the tests cover. If the other scenarios are valid from ascent's POV, then they should work in rover too.

The only potential bug I see is that the code currently creates a blank partial image in the event that an MPI rank had 0 domains to trace. My concern in that scenario would be that the blank partial images still get composited with the non-blank partial images. Best case, we're doing extra compositing work for no reason, but compositing with a blank image doesn't change the output. Worst case, we're doing extra compositing work for no reason, but compositing with a blank image does change the output. Although we don't currently have tests for scenarios other than `num_domains == num_ranks`, so I'm not certain about what happens there. I would think that we just shouldn't ever create blank partial images.

src/libs/rover/typed_scheduler.cpp
siramok: Truthfully, I don't know how composite handles this, but I do know that all ranks are intended to call `composite()` and that rank 0 ends up with the final composited image.

src/libs/rover/typed_scheduler.cpp
siramok: It's computed the same way, but now it's cached as a member variable as soon as we add new data: `m_num_local_domains = dataset.GetNumberOfDomains();` The number of domains per rank should never change during the same `rover.execute()` call, so it should be safe to cache. The purpose for adding `local` to the name is to hint that we're talking about the domains that the current rank has, because technically they could each have different numbers of domains to work on which I didn't find obvious before.

src/libs/vtkh/compositing/vtkh_diy_partial_blocks.hpp
siramok: Good question. It turns out that this code path only gets executed in the MPI case, so I didn't catch it before now. Not having these changes causes the optical depth values to not make it into each partial image result, but only when using MPI I guess.

src/tests/ascent/t_ascent_rover.cpp
siramok: I had the same thought actually. I think refactoring both tests to be uniform and share code where possible is a nice idea.

src/tests/ascent/t_ascent_rover.cpp
siramok: I could take a stab at it as part of the test refactoring mentioned above?

src/tests/ascent/t_ascent_mpi_rover.cpp
siramok: Yes, I can fix that. Funny enough that's not even the test file I copied from, and so I did a grep:
image

src/tests/ascent/t_ascent_mpi_rover.cpp
siramok: It could, but other MPI tests in the same situation are kept as separate files. I just followed the existing convention, so maybe we ask Cyrus for his thoughts first?

siramok: I deleted the first message as I misunderstood what you were saying here. I think this could work, maybe as part of the other refactoring we've mentioned? I'm assuming that would be its own PR.

src/libs/rover/typed_scheduler.cpp
JustinPrivitera: Thanks for the clarification. Any chance we could add a test where one rank gets a domain and another rank does not?

src/libs/rover/typed_scheduler.cpp
JustinPrivitera: We should investigate the composite implementation to understand what is going on.

src/libs/rover/typed_scheduler.cpp
JustinPrivitera: gotcha, sounds good.

src/libs/vtkh/compositing/vtkh_diy_partial_blocks.hpp
JustinPrivitera: Interesting. I wonder what vtkhdiy namespace is for any why it only comes up when running in parallel.

src/tests/ascent/t_ascent_rover.cpp
JustinPrivitera: want to do it in this PR or in a follow-on?

src/tests/ascent/t_ascent_rover.cpp
JustinPrivitera: yes, I like that idea

src/tests/ascent/t_ascent_mpi_rover.cpp
JustinPrivitera: sometimes we get loosey-goosey. Not if I'm reviewing 🙃

src/tests/ascent/t_ascent_mpi_rover.cpp
JustinPrivitera: yes, and up to you

src/libs/rover/typed_scheduler.cpp
siramok: Yes, should be doable.

src/libs/rover/typed_scheduler.cpp
siramok: I traced this out to see how it works:
1. [An allreduce happens](https://github.com/Alpine-DAV/ascent/blob/fc65b983d1e80bf97d50781ba064188c85cd42de/src/libs/vtkh/compositing/PartialCompositor.cpp#L504) so that each rank knows how many total partials images there are.
2. [The separate partial image vectors are combined into a single vector](https://github.com/Alpine-DAV/ascent/blob/fc65b983d1e80bf97d50781ba064188c85cd42de/src/libs/vtkh/compositing/PartialCompositor.cpp#L249) and 2 allreduce calls happen so that every rank knows the global min and max pixel ID.
3. All of the partial data is [distributed around in an intelligent way](https://github.com/Alpine-DAV/ascent/blob/fc65b983d1e80bf97d50781ba064188c85cd42de/src/libs/vtkh/compositing/vtkh_diy_partial_redistribute.hpp#L125) such that all ranks "own" various pixel ID ranges and receive all of the partial data that belong to those ranges.
4. Compositing happens as it normally does in the serial case, but each MPI rank has its own pixel ranges to work on (good to know that there aren't ranks waiting around doing nothing).
5. All of the composited partials are [collected back to rank 0](https://github.com/Alpine-DAV/ascent/blob/fc65b983d1e80bf97d50781ba064188c85cd42de/src/libs/vtkh/compositing/vtkh_diy_partial_collect.hpp#L113).

src/libs/vtkh/compositing/vtkh_diy_partial_blocks.hpp
siramok: Having traced how compositing works, I can see that it's vtkm/vtkh APIs for doing data-parallel things. Blocks seem like a description of what the passed-around data actually looks like, used when redistributing the partials around to ranks and when collecting them back to rank 0.

src/tests/ascent/t_ascent_rover.cpp
siramok: Knowing myself, I suspect it'll be a lot of changes, so let's do it in a follow-on.

src/libs/rover/typed_scheduler.cpp
JustinPrivitera: Awesome. Thanks for investigating.

src/libs/vtkh/compositing/vtkh_diy_partial_blocks.hpp
JustinPrivitera: gotcha

src/tests/ascent/t_ascent_rover.cpp
JustinPrivitera: sounds good.

JustinPrivitera: let's make a note somewhere to do it :)

src/libs/rover/typed_scheduler.cpp
siramok: Done.

src/tests/ascent/t_ascent_rover.cpp
siramok: A note has been added to the meta-issue.

JustinPrivitera (approved):
Looks good; thanks for the robust tests.

PR #1589, opened 2025-07-28
Fix Rover's absorption-only case

siramok:
Resolves bullets in #1513
- Fixes the crash that was occurring in the absorption-only case (when only an absorption field is passed).
- Optical depth is now computed in the absorption-only case.


src/libs/vtkh/compositing/VolumePartial.hpp
```diff
@@ -94,6 +94,11 @@ struct VolumePartial
     m_alpha = m_alpha > 1.f ? 1.f : m_alpha;
   }
 
+  inline void blend_optical_depth(const VolumePartial<FloatType> &other)
```
siramok, line 97: This is only here because `blendPartials()` is templated, and I can't call `blend_optical_depth()` for `AbsorptionPartial`s if it doesn't also exist for `VolumePartial`. This is more volume rendering stuff that didn't live within rover, and so wasn't removed earlier in the summer.

src/libs/vtkh/rendering/ConnectivityProxy.cpp
```diff
@@ -176,16 +184,17 @@ struct ConnectivityProxy::InternalsType
   void PartialTrace(vtkm::rendering::raytracing::Ray<vtkm::Float64> &rays,
                     PartialVector64 &partials)
   {
-    if (this->Mode == RenderMode::Volume)
+    if (EmissionFieldName.empty())
     {
-      Tracer.SetVolumeData(this->Dataset.GetField(this->FieldName),
-                           this->ScalarRange,
+      // Absorption-only case
+      Tracer.SetEnergyData(this->Dataset.GetField(this->FieldName),
```
siramok, line 194: This is why we were crashing, because `SetEnergyData` was always doing this: `Dataset.GetField(this->EmissionFieldName)`, attempting to access a field that doesn't exist in the dataset. The fix is that I just overloaded this function to not require an emission to be passed.

src/tests/_baseline_images/tout_rover_xray_blueprint_braid_absorption_only_intensities_000100.png
siramok, line 1: We accumulate `1/exp(absorption * segment_length)` to compute this field, but it seems like `absorption * segment_length` is always large enough to make this term become 0. This is true despite a unit scalar of 0.000001f (which gets multiplied with absorption in an attempt to make the term not be 0). I think we need a proper test case in which we know what we want to see the output be.

src/tests/_baseline_images/tout_rover_xray_blueprint_braid_absorption_only_optical_depth_000100.png
siramok, line 1: Optical depth output is identical to the absorption+emission case. The output of the absorption-only case is equivalent to just taking transmission/transmittance directly from the absorption+emission case. An argument could be made for merging both cases into one code path, since the absorption+emission case could also be wired up to have total absorption in the blueprint output. The main benefit would be simpler code, easier to maintain.

The argument for keeping them as separate code paths is that there's a (unmeasured) performance or memory benefit to be had in the case that someone only cares about absorption, but I also don't have context for how important this feature is to users to know whether it's a worthwhile tradeoff.

However, if the powers that be want them to stay separate, then there may be a further (unmeasured) performance and memory improvement to be had by only computing and compositing the optical depth (accumulated `absorption * segment_length`) and then deriving absorption from it at the very end by just taking `exp(-optical_depth)`.

src/tests/ascent/t_ascent_rover.cpp
```diff
@@ -349,6 +349,118 @@ TEST(ascent_rover, test_xray_blueprint_braid_rotated)
     ASCENT_ACTIONS_DUMP(actions, query_path, msg);
 }
 
+//-----------------------------------------------------------------------------
+TEST(ascent_rover, test_xray_blueprint_braid_absorption_only)
```
siramok, line 373: As mentioned, a proper test case is needed that yields something other than a field of 0s, so this can be treated as just a placeholder.

src/libs/vtkh/compositing/PartialCompositor.cpp
```diff
@@ -349,71 +343,71 @@ PartialCompositor<PartialType>::composite_partials(std::vector<PartialType> &par
...
 
 #ifdef VTKH_OPENMP_ENABLED
   #pragma omp parallel for
 #endif
-  for(int i = 1; i < n_minus_one; ++i)
+  for(int i = 1; i < n_minus_one; i++)
   {
-    unsigned char work_flag = 0;
-    unsigned char unique_flag = 0;
     bool is_begining = false;
```
JustinPrivitera:
```suggestion
    bool is_beginning = false;
```

JustinPrivitera: will need to change elsewhere too

src/libs/vtkh/rendering/ConnectivityProxy.cpp
JustinPrivitera: Excellent, thanks for figuring this out.

src/tests/_baseline_images/tout_rover_xray_blueprint_braid_absorption_only_intensities_000100.png
JustinPrivitera: This is unfortunate. We will need to think more about the math for this case.

src/tests/_baseline_images/tout_rover_xray_blueprint_braid_absorption_only_optical_depth_000100.png
JustinPrivitera: Let's discuss high-bandwidth. Not sure I quite understand.

JustinPrivitera (commented):
Hey this all looks great. Just some minor things and then we have a few discussion items to better understand later.

src/tests/_baseline_images/tout_rover_xray_blueprint_braid_absorption_only_optical_depth_000100.png
siramok: Happy to explain.

src/libs/vtkh/compositing/PartialCompositor.cpp
siramok: Nice, I didn't catch that.

siramok: Fixed.

src/tests/_baseline_images/tout_rover_xray_blueprint_braid_absorption_only_intensities_000100.png
siramok: We talked to Steve and sorted this out.

src/tests/_baseline_images/tout_rover_xray_blueprint_braid_absorption_only_optical_depth_000100.png
siramok: There's almost a 2x performance difference with absorption-only, so I think keeping them as separate code paths is best. (In total test execution time, not a rigorous timing of rover itself.)

src/tests/ascent/t_ascent_rover.cpp
siramok: Sorted this out, intensity doesn't mean anything in the absorption-only case. Optical depth is what we're after.

JustinPrivitera (approved):
Looks great; thanks for these changes!

siramok:
The failing CI appears to be unrelated to this PR, but  regardless I'll wait for an explicit go-ahead to merge. We can also just try re-running CI tomorrow.

PR #1590, opened 2025-07-30
Refactor Rover's ray generators

### LINKED ISSUE #1464: m_has_rays in libs/rover/ray_generators/{ray_generator,camera_generator} issues
I noticed there is a member `m_has_rays` in the `ray_generator` class. It is always set to `false` in `camera_generator`, which seems incorrect. I assume it should only be set to false if there are actually no rays generated. The only place `m_has_rays` is used is in `RayGenerator::get_has_rays`, which is never called in the entire code base.

Assuming this is an issue, either the function and variable can be removed or the code in `camera_generator` can be fixed. It should probably also be set in `visit_generator`.

I'm happy to fix if it is an issue and if we decide which of the two paths to take.

siramok:
Resolves #1464 and resolves bullets in #1513.
- Refactor ray generators -- we no longer need to dynamic cast.
- We no longer have m_has_rays, and instead return a bool to indicate whether ray creation succeeded.


src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
```diff
@@ -402,6 +402,7 @@ RoverXRay::verify_params(const conduit::Node &params,
     "camera/fov",
     "camera/look_at",
     "camera/near_plane",
+    "camera/position",
```
JustinPrivitera, line 405: Again, I am confused what this is for. Do we allow users to set this via rover settings?

JustinPrivitera:
Could you describe in a couple sentences how the ray generator class hierarchy design changed?

src/libs/rover/typed_scheduler.cpp
```diff
@@ -316,20 +316,11 @@ TypedScheduler<FloatType>::trace_rays()
...
 
-  m_ray_generator->reset();
-
   set_global_range_and_bounds();
 
   vtkmTimer trace_timer;
   trace_timer.Start();
 
-  // TODO: Don't love that we need dynamic_cast
-  // TODO: Actually support both cases, vtkm and visit. Add tests
```
JustinPrivitera, line 327: we never added tests for the visit camera generator case. Are we supporting that case in this PR? If we are, we need to add tests. If not, we should leave a note somewhere to add tests.

src/libs/rover/typed_scheduler.cpp
```diff
@@ -343,12 +334,17 @@ TypedScheduler<FloatType>::trace_rays()
...
 
     timer.Start();
 
-    cast_ray_generator->get_rays(rays);
+    // TODO: I'm curious about which conditions can cause rays to fail to be created
+    if (!m_ray_generator->get_rays(rays))
+    {
+      ROVER_ERROR("Failed to create new rays");
+    }
+
```
JustinPrivitera, line 347: Why did you add in this conditional? Did you encounter a case where it failed?

JustinPrivitera (commented):
Looks good. Just some design questions.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
siramok: Yes. Imagine I want the x-ray detector to be positioned at [0, 1, 2] (just something specific), this is necessary for allowing that. All of the other vtkm camera parameters are supported, and I think setting position is one that got removed (by me) by accident.

src/libs/rover/typed_scheduler.cpp
siramok: I didn't encounter a case in which it failed, but this serves the purpose that I would have expected `m_has_rays` to serve. Once you create rays, see if you actually have them, and do something in the case that you don't.

src/libs/rover/typed_scheduler.cpp
siramok: Good point, that TODO specifically should not have been removed. I will add it back.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rover_filters.cpp
JustinPrivitera: sounds good.

siramok:
Regarding what this PR does: The base `RayGenerator` class wasn't being used for anything directly, sort of like what `Scheduler` is to `TypedScheduler`. However, it was a bit too basic to be useful, and so the `TypedScheduler` code was previously doing a dynamic cast to ensure that the ray generator was always a `VtkmRayGenerator`. This was wonky. Since the `VtkmRayGenerator` is essentially the default ray generator, this PR turns it into the base `RayGenerator` and the `VisitRayGenerator` now inherits from it. The `VisitRayGenerator` should ultimately use the default `RayGenerator` methods for creating rays but just apply visit camera params to the vtkm camera before doing so. However, it's not currently possible for a user to provide visit camera params, but this would make it a bit easier to support that if there's interest.

JustinPrivitera:
> Regarding what this PR does: The base `RayGenerator` class wasn't being used for anything directly, sort of like what `Scheduler` is to `TypedScheduler`. However, it was a bit too basic to be useful, and so the `TypedScheduler` code was previously doing a dynamic cast to ensure that the ray generator was always a `VtkmRayGenerator`. This was wonky. Since the `VtkmRayGenerator` is essentially the default ray generator, this PR turns it into the base `RayGenerator` and the `VisitRayGenerator` now inherits from it. The `VisitRayGenerator` should ultimately use the default `RayGenerator` methods for creating rays but just apply visit camera params to the vtkm camera before doing so. However, it's not currently possible for a user to provide visit camera params, but this would make it a bit easier to support that if there's interest.

Thanks for this. I remember discussing this with you now. This is useful info to capture somewhere in the PR conversation so that future developers (or us) can look back and remember.

src/libs/rover/typed_scheduler.cpp
JustinPrivitera: what does get_rays return?

siramok: Previously, it was returning void. I modified it to return `return rays.NumRays > 0`. That way if 0 rays are created, the conditional fails (essentially checking if `m_has_rays`) and then we can't proceed with the ray trace.

siramok: Which is the piece that resolves Eric's original issue, where I said that `m_has_rays` looked to be dead code.

JustinPrivitera: I see, this makes sense now.

JustinPrivitera (approved):

PR #1593, opened 2025-08-05
Refactor Rover tests

### LINKED ISSUE #1601: xray emission parsing yaml issue
If not used emission should not be present. 

While in c++ we can create an empty string, we cannot do so with yaml.

Here is an example:

```
-
  action: "add_extracts"
  extracts:
    xray:
      type: "xray"
      params:
        rover:
          absorption: briad
          emission: ""
          filename: tout_xray
          output_type: hdf5
          enable_rays_mesh: false
        camera:
            far_plane:  5
            near_plane: 0.1
            fov: 60
```

Emission in this case will parse to Conduit's `None` Type.

```
  emission: 

>>> n["emission"].dtype()
<DataType object at 0x1115ae370>
>>> print(n["emission"].dtype())

{
  "dtype":"empty"
}
```

Rover is unhappy with this:

```
line: 138
message: 
Cannot create filter xray(type: xray) because verify_params failed.
Details:

errors: 
  - "Optional string parameter 'rover/emission' is not a string"
 
```

siramok:
Fixes #1601 and resolves bullets in #1513.
- Creates a separate `rover_test_utils.hpp` header file that holds the utility functions that the serial and MPI rover tests both use or share. It uses `ifdef`s to include/exclude MPI-specific code.
- New utilities have been created to further reduce boilerplate across both test files.
- A pass was taken at refactoring the way we do diff tests. The new approach tries to fill in as much of the baseline node as possible using what we already know about the query (i.e. the user params). Although some things still have to be manually set, mainly the final camera information and data min/maxes.
- Adds diff checks to the tests which didn't previously have them (all tests should have a diff check going forward).
- Adds an additional serial absorption-only test to ensure we exercise both the case in which emission is omitted and also the case in which emission is explicitly set to `""`.
- Adds an MPI absorption-only test since one didn't already exist.
- MPI tests can now be executed with any number of ranks and still pass.
- Remove support for setting emission to an empty string `""`. Doing so will now cause `verify_params` to fail, the only way to ask for the absorption-only case is to omit the emission field entirely.


src/libs/rover/rover.cpp
```diff
@@ -32,6 +32,7 @@ Rover::Rover()
   // Settings
   rover::settings["background_intensity"] = 0.0f;
   rover::settings["divide_emis_by_absorb"] = "false";
+  rover::settings["emission"] = "";
```
siramok: Unfortunately, not having this would cause Conduit errors later when attempting to execute `rover::settings["emission"].as_string()`. Should have been caught in the absorption-only PR, but oh well. Making it an empty string by default preserves the behavior we want (since we always check for `.empty()` anyways) without having to check `.has_child("emission")` in several new places.

src/libs/rover/typed_scheduler.cpp
```diff
@@ -991,6 +991,11 @@ TypedScheduler<FloatType>::to_blueprint(Node &data)
     intensities_spatial.set(intensities);
     intensities_spatial["topology"] = "spatial_topo";
   }
+  else // (emission.empty())
+  {
+    xray_data["intensity_max"] = 0.0;
```
siramok: This is mainly to make the new diff checking approach happy, else it will find that the absorption-only output contains empty while the baseline has an actual type. We could also make it `-1.0` or something if we're worried that `0.0` could be confused for an actual result.

src/tests/ascent/CMakeLists.txt
```diff
@@ -141,7 +141,8 @@ if(GENTEN_FOUND)
     list(APPEND BASIC_TESTS t_ascent_genten_cokurt)
 endif()
 
-
+# Make t_rover_test_utils.hpp visible to the rover tests
+include_directories(${CMAKE_CURRENT_SOURCE_DIR})
```
siramok: Saw that this is how apcomp did it.

src/tests/ascent/CMakeLists.txt
```diff
@@ -200,6 +201,8 @@ if(MPI_FOUND AND ENABLE_MPI)
       add_cpp_mpi_test(TEST t_ascent_babelflow_comp_mpi NUM_MPI_TASKS 8 DEPENDS_ON ascent_mpi)
     endif()
   endif()
+
+  target_compile_definitions(t_ascent_mpi_rover PRIVATE ASCENT_MPI_ENABLED)
```
siramok: I guess none of the other tests have needed this before, which makes sense, since they don't share any utilities like rover does.

siramok:
Assuming CI passes, I've done all that I can think to do. No rush on a review. :)

siramok:
I lied, I thought of one more thing. I investigated the TODO to see if performing all of the renders as part of one ascent action was any faster. The answer was yes, but by ~1-2% as a 10 run average of the entire test on my home PC. Since I could measure the improvement, I went ahead and committed it.

```
----1 ASCENT ACTION PER RENDER----
hyperfine './build/ascent-checkout/tests/ascent/t_ascent_rover'
Benchmark 1: ./build/ascent-checkout/tests/ascent/t_ascent_rover
  Time (mean ± σ):      5.357 s ±  0.080 s    [User: 17.096 s, System: 0.516 s]
  Range (min … max):    5.248 s …  5.522 s    10 runs
 
hyperfine 'mpiexec -n 2 ./build/ascent-checkout/tests/ascent/t_ascent_mpi_rover'
Benchmark 1: mpiexec -n 2 ./build/ascent-checkout/tests/ascent/t_ascent_mpi_rover
  Time (mean ± σ):      3.439 s ±  0.032 s    [User: 5.690 s, System: 0.369 s]
  Range (min … max):    3.404 s …  3.506 s    10 runs
 
 ----1 ASCENT ACTION TOTAL----
hyperfine './build/ascent-checkout/tests/ascent/t_ascent_rover'
Benchmark 1: ./build/ascent-checkout/tests/ascent/t_ascent_rover
  Time (mean ± σ):      5.270 s ±  0.049 s    [User: 17.032 s, System: 0.656 s]
  Range (min … max):    5.173 s …  5.336 s    10 runs
 
hyperfine 'mpiexec -n 2 ./build/ascent-checkout/tests/ascent/t_ascent_mpi_rover'
Benchmark 1: mpiexec -n 2 ./build/ascent-checkout/tests/ascent/t_ascent_mpi_rover
  Time (mean ± σ):      3.439 s ±  0.054 s    [User: 5.585 s, System: 0.418 s]
  Range (min … max):    3.374 s …  3.536 s    10 runs
  ```

File: src/libs/rover/typed_scheduler.cpp (hunk 2 above)
Comment (JustinPrivitera) [reply src/libs/rover/typed_scheduler.cpp]: I think we should handle these kinds of differences in the test files instead of in the actual code. We should be able to manually add or remove children from the nodes we want to diff as needed. We can add notes that we are changing things so that the diff will pass.

File: src/libs/rover/engine.cpp [hunk 5]
```diff
@@ -64,7 +64,7 @@ Engine::init_emission(vtkmRayTracing::Ray<Precision> &rays,
   {
     ROVER_INFO("Engine::init_emission: emission not specified");
   }
-  else
+  else // (!emission.empty())
```
JustinPrivitera: why remove the comment?

src/libs/rover/rover.cpp
JustinPrivitera: Couldn't we ask for `rover::settings.has_child("emission");` before calling `rover::settings["emission"].as_string()`? I guess how many places are there where we would need to ask if it exists first? I don't think it is all that expensive.

src/tests/ascent/t_rover_test_utils.hpp
JustinPrivitera, line 1: why don't we just call this file `rover_test_utils.hpp`. Since it is not a test itself it should not have the `t_` prefix.

src/tests/ascent/CMakeLists.txt
JustinPrivitera: Do we even need any changes to cmake? What happens if we do not modify this file at all?

src/tests/ascent/t_rover_test_utils.hpp
```diff
@@ -0,0 +1,328 @@
...
+        ASCENT_INFO("Found differences in the blueprint diff:\n");
+        std::cout << diff_info.to_yaml() << std::endl;
+    }
+
+    EXPECT_FALSE(has_differences);
+}
+
+//-----------------------------------------------------------------------------
+inline void
+get_valid_test_data(Node &data)
```
JustinPrivitera: can we rename this to `get_braid_test_data`

src/tests/ascent/t_rover_test_utils.hpp
```diff
@@ -0,0 +1,328 @@
...
+                                              EXAMPLE_MESH_SIDE_DIM,
+                                              EXAMPLE_MESH_SIDE_DIM,
+                                              EXAMPLE_MESH_SIDE_DIM,
+                                              data);
+    EXPECT_TRUE(conduit::blueprint::mesh::verify(data, verify_info));
+}
+
+//-----------------------------------------------------------------------------
+inline void
+get_valid_multi_domain_test_data(Node &data, const int num_domains)
```
JustinPrivitera: `get_braid_multi_domain_test_data`

src/tests/ascent/CMakeLists.txt
cyrush: Do the rover test utils have mpi and non mpi conditional helpers?

src/libs/rover/rover.cpp
cyrush: In my tests using ascent on sim data, I still always had to set emission to something, or else I would get a conduit fetch error. 

Do we have a test that exercises that simple case? 

As justin suggests, I think its better to check for the existence not an `""` if possible.

cyrush: Here is an example of the error I see when I omit emission:

```
-
  action: "add_extracts"
  extracts:
    xray:
      type: "xray"
      params:
        rover:
          absorption: den
          #emission: den
          filename: tout_xray
```


```
[Error] Ascent::execute 
file: /usr/WS1/cyrush/2025.08.ascent/ascent/scripts/build_ascent/source/conduit-v0.9.4/src/libs/conduit/conduit_node.cpp
line: 15767
message: 
Node::as_string() const -- DataType empty at path emission does not equal expected DataType char8_str
```

Just to make sure I am not out of sync, here is the commit I built:
```
commit 4a594be45d9b060ea68bd5a4630e3763cdca5027 (HEAD -> develop, origin/task/8_1_2025-1278-allow-reuse-of-past-actions, origin/develop, origin/HEAD)
Author: Andres Sewell <30913496+siramok@users.noreply.github.com>
Date:   Thu Jul 31 04:19:42 2025 -0700

    Refactor Rover's ray generators (#1590)
```

src/tests/ascent/t_rover_test_utils.hpp
```diff
@@ -0,0 +1,328 @@
...
+        xray_view: 
+            position: [0.0, 0.0, 0.0]
+            zoom: 1.0
+            look_at: [0.0, 0.0, 0.0]
+            up: [0.0, 1.0, 0.0]
+            fov: 60.0
+            xpan: 0.0
+            ypan: 0.0
+            near_plane: 0.0
+            far_plane: 0.0
```
JustinPrivitera: It occurs to me that we are writing these values out from Rover different than VisIt: https://visit-sphinx-github-user-manual.readthedocs.io/en/develop/using_visit/Quantitative/XRayImageQuery.html#metadata

It should be 
```
position:
   x: 0
   y: 0
   z: 0
```
instead of 
```
position: [0, 0, 0]
```
Same goes for look_at and up, which should probably be named `view_up`. Instead of `xpan` and `ypan` we should have
```
image_pan:
   x:
   y:
```

JustinPrivitera: I added a ticket for this: #1598 and added a link to it in the larger Rover enhancements ticket: #1513

JustinPrivitera (changes requested):
See feedback above.

src/libs/rover/rover.cpp
siramok: I'll reply to the other feedback when I have time, but this one seems to be a bit more pressing so I'll reply now @cyrush:

You're not out of sync, the develop branch does currently have that bug. Silly mistake on my end, but thankfully it's an easy fix.

We were thinking that it should be possible to either 1) omit emission from the actions entirely or 2) set emission to `""` to get the absorption-only code-path. For whatever reason, I only tested with setting emission to `""` and didn't notice that an error was being thrown when emission was omitted. Whoops. This PR will fix it. Adding existence checks isn't a problem, I can do that instead of setting a default emission field of `""`. With respect to making sure there are tests that exercise both cases, I can also do that.

But that prompts a more fundamental question for me: would you prefer rover to consider both as valid ways of asking for absorption-only, or would you prefer to make things stricter and disallow  `""`? We could check for `""` in `verify_params` and just have verification fail if that's the case.

JustinPrivitera: > But that prompts a more fundamental question for me: would you prefer rover to consider both as valid ways of asking for absorption-only, or would you prefer to make things stricter and disallow `""`? We could check for `""` in `verify_params` and just have verification fail if that's the case.

We should be robust to both cases. Allow both I think. Not too hard to handle.

src/libs/rover/typed_scheduler.cpp
siramok: Your point makes sense. I'll remove this and will fix it within the tests instead.

src/tests/ascent/CMakeLists.txt
siramok: With respect to this line specifically, I commented it out and sure enough, it wasn't necessary. The tests still build and run, so I'll remove this. With respect to the other change in this file (adding a preprocessor macro to know whether we're building the utilities with MPI), I think the answer is yes if we want both the serial and MPI tests to call the same utilities from the same header file.

src/tests/ascent/CMakeLists.txt
siramok: Yes they do! For context, the motivation here was that both the serial and MPI test files had similar utilities defined in them that were ~95% the same code.

1. Is this approach okay with you? There may be other ways of achieving this that I could investigate if you're not a fan. 
2. If yes to 1., would you prefer `ASCENT_MPI_ENABLED` to only be defined for `t_ascent_mpi_rover` (what the above currently does) or do you think all of the MPI tests should have it defined for them?
3. If yes to 1. but no to 2., would you prefer it be called something else more rover or test specific, such as `ROVER_TEST_MPI_ENABLED`? I picked the current name for consistency with what's used elsewhere in the source.

src/libs/rover/engine.cpp
siramok: I went and double checked, this is adding the comment, not removing it. I noticed it was missing after investigating the Conduit `as_string()` error.

src/tests/ascent/t_rover_test_utils.hpp
siramok: I'm fine with your suggestion. The current name was also a convention borrowed from apcomp: `t_apcomp_test_utils.h`.

src/tests/ascent/t_rover_test_utils.hpp
siramok: I like that.

src/tests/ascent/t_rover_test_utils.hpp
siramok: Rolls off the tongue (I'll change it).

src/tests/ascent/t_rover_test_utils.hpp
siramok: Ah, good catch, should be simple to change. Since you made a dedicated issue for it, I can do that in a follow-up PR.

src/libs/rover/rover.cpp
siramok: Fixed. No longer do we set a default emission in rover's settings, and instead we do existence checking.

src/libs/rover/typed_scheduler.cpp
siramok: Fixed. Handled in the tests now.

src/tests/ascent/t_rover_test_utils.hpp
siramok: Fixed. Renamed it to your suggestion.

src/tests/ascent/t_rover_test_utils.hpp
siramok: Fixed. Renamed it to your suggestion.

src/tests/ascent/t_rover_test_utils.hpp
siramok: Fixed. Renamed it to your suggestion.

siramok:
In addition to addressing the quick points of feedback, I also added a rotated variant of the serial absorption-only test to ensure we exercise both omitting emission and setting emission to `""`. Furthermore, I noticed that there wasn't an absorption-only MPI test, so I added one because that seems important to have. Finally, I thought of a few more minor improvements that weren't asked for, like setting default query params with a utility function to further reduce boilerplate (since all the tests have to define the same required rover params).

src/tests/ascent/CMakeLists.txt
cyrush: 1) Yes, this sounds good.

I think it would be better to use a unique define, such as `ROVER_TEST_MPI_ENABLED`. (3)

The `ASCENT_MPI_ENABLED` allows us to compile the right API when building the lib, so its best not to use it when testing the lib. We have tricked ourselves with this before and it makes you question reality :-).

src/tests/ascent/t_rover_test_utils.hpp
cyrush: Ideally we can feed this camera right back into Ascent. So let's consider that as well.  I don't think that was ever considered when writing the VisIt data.

src/libs/rover/rover.cpp
JustinPrivitera: thanks

src/libs/rover/typed_scheduler.cpp
JustinPrivitera: great

src/tests/ascent/CMakeLists.txt
JustinPrivitera: sounds good

src/libs/rover/engine.cpp
JustinPrivitera: Sorry, my bad. Did not parse the diff correctly :(

src/tests/ascent/t_rover_test_utils.hpp
JustinPrivitera: I didn't realize that. I looked to Conduit for inspiration where we have done this other approach. I was hoping for consistency but perhaps it is too late. I still maintain that the `t_` prefix only being used for actual tests is clearer. 🙃

src/tests/ascent/t_rover_test_utils.hpp
JustinPrivitera: thanks

src/tests/ascent/t_rover_test_utils.hpp
JustinPrivitera: thanks

src/tests/ascent/t_rover_test_utils.hpp
JustinPrivitera: I added your comment to the ticket I opened so we can remember and discuss another time what the best solution is.

JustinPrivitera:
> In addition to addressing the quick points of feedback, I also added a rotated variant of the serial absorption-only test to ensure we exercise both omitting emission and setting emission to `""`. Furthermore, I noticed that there wasn't an absorption-only MPI test, so I added one because that seems important to have. Finally, I thought of a few more minor improvements that weren't asked for, like setting default query params with a utility function to further reduce boilerplate (since all the tests have to define the same required rover params).

Thanks for these changes!

src/tests/ascent/CMakeLists.txt
```diff
@@ -142,7 +142,7 @@ if(GENTEN_FOUND)
 endif()
 
 # Make t_rover_test_utils.hpp visible to the rover tests
-include_directories(${CMAKE_CURRENT_SOURCE_DIR})
+# include_directories(${CMAKE_CURRENT_SOURCE_DIR})
```
JustinPrivitera: Comment it out? Or just delete it entirely?

src/tests/ascent/rover_test_utils.hpp
```diff
@@ -304,25 +346,60 @@ remove_rover_test_data(const std::string &path,
...
+    return true;
+}
+
+//-----------------------------------------------------------------------------
+#ifdef ASCENT_MPI_ENABLED
+inline const bool
+has_two_mpi_ranks(int par_rank, int par_size)
+{
+    // The rover + MPI tests will fail if there are not exactly 2 ranks
+    // due to the baselines being generated with 2 ranks
```
JustinPrivitera: Why will the tests fail? The same images should be generated no matter how many ranks there are. Are the tests only set up to make sense for two ranks?

JustinPrivitera (approved):
Thanks @siramok!

src/tests/ascent/CMakeLists.txt
siramok: Whoops. Thanks for actually looking at my changes, I did mean to properly delete it.

src/tests/ascent/CMakeLists.txt
siramok: This has been fixed.

src/tests/ascent/rover_test_utils.hpp
siramok: The reason was because our input data was always being created as 1 domain per rank while our baseline images were generated using exactly 2 domains to pass CI. Running the tests with anything other than 2 ranks would change the number of domains in the input mesh with respect to what was used to generate the baseline images, and subsequently cause the image checks to fail.

Just for fun, I investigated and came up with this list of other tests which also fail with >2 MPI ranks:
- `t_ascent_mpi_add_ranks`
- `t_ascent_mpi_expressions`
- `t_ascent_mpi_relay_extract`
- `t_ascent_mpi_render_2d` (actually segfaults, not just fails)
- `t_ascent_mpi_render_3d`
- `t_ascent_mpi_slice`
- `t_ascent_mpi_uniform_grid`

In fairness, I don't view this as a huge problem since CI only ever runs with 2 ranks, implying that someone would have to be crazy enough (myself) to go out of their way and manually run each test with >2 ranks.

However, I thought you had a good point and it seemed like fun problem to solve, so I took a stab at it. The rover MPI tests will now pass for 1, 2, and >2 MPI ranks because I explicitly handle each case to make sure that exactly 2 domains are created regardless of the number of MPI ranks.

src/tests/ascent/rover_test_utils.hpp
```diff
@@ -0,0 +1,464 @@
...
+    // There are 3 cases to support with respect to emission:
+    // 1. Emission is not set in the params: the absorption-only case
+    // 2. Emission is set to an empty string: the absorption-only case
+    // 3. Emission is set to a non-empty string: the absorption + emission case
+    if (xray_query.has_child("emission"))
+    {
+        const std::string emission = xray_query["emission"].as_string();
+        if (emission.empty())
+        {
+            // TODO: Investigate if file format differences like this
```
siramok: In my refactoring adventures, I discovered what seems to be a difference in how ascent's `mesh_blueprint_save` writes JSON vs YAML (didn't have a chance to look into it further).

This is a problem I found with the absorption-only case (where emission is either omitted or set to `""`). The TLDR is that if we ask rover for YAML output and then load that output into memory, emission is parsed as just empty. But if we ask rover for JSON output and then load  that output into memory, emission is parsed as an empty string `""`.

src/tests/ascent/rover_test_utils.hpp
```diff
@@ -0,0 +1,464 @@
...
+            // If our output type is JSON, the emission field is present in the output as "",
+            // whereas YAML treats it as empty. We only have to do the following in the YAML case 
+            if ("yaml" == baseline_data["xray_query/output_type"].as_string())
+            {
+                // We don't want the emission field to report as a string type,
+                // so we remove it and then add it back as an empty field for the
+                // purpose of passing the diff check
+                baseline_data["xray_query"].remove_child("emission");
+                baseline_data["xray_query/emission"];
+            }
```
siramok: In my refactoring adventures, I discovered what seems to be a difference in how ascent's `mesh_blueprint_save` writes JSON vs YAML (didn't have a chance to look into it further).

This is a problem I found with the absorption-only case (where emission is either omitted or set to `""`). The TLDR is that if we ask rover for YAML output and then load that output into memory, emission is parsed as just empty. But if we ask rover for JSON output and then load  that output into memory, emission is parsed as an empty string `""`.

(reposted because I figured out that I could have been commenting on more than one line all along)

cyrush: @siramok yes, see #1601

Sorry I though I linked this issue yesterday with a discussion but I don't see it. 

I think we should remove the "" case for emission.
- If it is there, it should be a valid field name
- if it is not, then we are in the absorption only case.  

This matches the general philosophy of how we aim to provide parameters and data via conduit.

src/tests/ascent/rover_test_utils.hpp
cyrush: The tests are coded for a specific number of ranks. 
Most of them are set to use the default for our testing harness (2), https://github.com/Alpine-DAV/ascent/blob/0b7251d284b2678843bad1e4b6f2ed6929641658/src/cmake/SetupTests.cmake#L141 
but you can launch with more tasks if needed.  A few tests do do this , for example:

https://github.com/Alpine-DAV/ascent/blob/0b7251d284b2678843bad1e4b6f2ed6929641658/src/tests/ascent/CMakeLists.txt#L192

JustinPrivitera: I see @siramok. That makes sense. Thanks for being so thorough! It doesn't hurt any for the tests to pass with more than 2 ranks.

src/tests/ascent/rover_test_utils.hpp
JustinPrivitera: Sorry for the conflicting instructions! I think I told you to do the opposite. Cyrus is right and I didn't catch this.

siramok: No problem at all, I can fix that.

siramok: I took a pass at removing support for setting empty-string emission fields:
1. If emission is set to an empty string, `verify_params` will fail with an info message explaining why. So empty strings are simply no longer allowed.
2. Given that, once we've made it to `rover.execute()`, we can safely assume that if emission is set then it must not be an empty string. This allowed for a bit of simplification.
3. I was able to clean up the above test workaround, since the YAML vs JSON edge case is no longer triggerable.

siramok:
Once the most recent commit is reviewed, CI passes, and if there's not additional feedback to address, I think this PR is good to go!

JustinPrivitera (approved):
Do we have tests covering the absorption-only case? I think we do just want to check. Thanks for all the iteration on this!

siramok:
We sure do, 2 for serial and 1 for MPI.

cyrush (approved):
Looks good, thanks for these updates!

PR #1594, opened 2025-08-07
fix foreground and background bug when rendering a mesh

nicolemarsaglia:
Need to apply the foreground and background colors to the canvas


cyrush (approved):
Thanks for this fix!

PR #1595, opened 2025-08-07
Adding a caching functionality for the actions

### LINKED ISSUE #1278: allow reuse of past actions
if user knows actions won't change, provide a path to avoid checking for ascent_actions.{yaml,json}, and use last set of actions.

emily-howell:
This adds a runtime option to "cache_actions" which, when set to true, will only load actions if none have been loaded before. With this behavior, it doesn't matter what is passed to the execute function after the first time. The test demonstrates this by passing a blank actions node to the execute function for executions 2 & 3 of ascent showing that scenes are still rendered each time ascent is run.


src/tests/ascent/t_ascent_runtime_options.cpp
```diff
@@ -1231,3 +1231,89 @@ TEST(ascent_runtime_options, test_field_filtering_new)
...
+    }
+
+    //
+    // Create the actions.
+    //
+
+    conduit::Node scenes;
+    scenes["s1/plots/p1/type"] = "pseudocolor";
+    scenes["s1/plots/p1/field"] = "braid";
+    scenes["s1/image_prefix"] = output_file_base + "{family}";
```
cyrush, line 1286: family makes it easy to test this one :-)

cyrush (approved):
Looks good,thanks!

PR #1596, opened 2025-08-07
Task/7 15 25 1515 visit camera mode helper

### LINKED ISSUE #1515: visit camera mode helper
We should explore a unified helper to accept visit style camera and convert it to a ascent camera.

emily-howell:
Strange behavior with clipping planes persists.

Unimplimented visit parameters currently include:
- [ ] perspective
- [ ] eyeAngle
- [ ] centerOfRotationSet
- [ ] centerOfRotation
- [ ] axis3DScaleFlag
- [ ] axis3DScale
- [ ] shear
- [ ] windowValid

Not sure if there is a 1:1 conversion for those from visit parameters to vtkm ones yet


cyrush:
-  perspective # deals with switching between orthogonal and perspective modes. In the future we want orthogonal, but thats not possible now
-  eyeAngle # deals with stereo rendering, we don't need this now
-  centerOfRotationSet   # you can skip, tells visit if special center of rotation is set
-  centerOfRotation # you can skip, special center of rotation for visit
-  axis3DScaleFlag # you can skip, tells visit if it should apply extra scal
-  axis3DScale # you can skip, visit has an extra scale function for its 3d view 
-  shear # you can skip, visit has an extra shear function for this 3d view
-  windowValid # you can skip

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rendering_filters.cpp
```diff
@@ -200,6 +189,44 @@ check_renders_surprises(const conduit::Node &renders_node)
...
+  c_ascent_valid_paths.push_back("azimuth");
+  c_ascent_valid_paths.push_back("elevation");
+
+  // Valid Visit input camera format
+  std::vector<std::string> c_visit_valid_paths;
+  c_visit_valid_paths.push_back("viewNormal");
+  c_visit_valid_paths.push_back("focus");
+  c_visit_valid_paths.push_back("viewUp");
+  c_visit_valid_paths.push_back("viewAngle");
+  c_visit_valid_paths.push_back("parallelScale");
```
cyrush: I would still like to keep these snake_case.

emily-howell: Very valid. I was matching the naming in visit but can go and change those

src/libs/ascent/runtimes/flow_filters/ascent_runtime_conduit_to_vtkm_parsing.cpp
```diff
@@ -191,6 +196,124 @@ parse_camera(const conduit::Node camera_node, vtkm::rendering::Camera &camera)
...
+      double distance = (parallel_scale) / std::tan(view_angle);
+
+      // Normalize viewNormal vector
+      conduit::float64 norm = std::sqrt(v[0]*v[0] + v[1]*v[1] + v[2]*v[2]);
+      conduit::float64 v_norm[3] = { v[0]/norm, v[1]/norm, v[2]/norm };
+
+      conduit::float64 pos[3];
+      for(int i = 0; i < 3; ++i)
+      {
+        pos[i] = fc[i] + v_norm[i] * distance;
```
JustinPrivitera: I am confused. I think `position` equals `focus` minus `view_normal * distance`. Is that right?

emily-howell: I think it is plus. That is the math I came up with and the visit session to camera conversion script seems to agree: https://github.com/Alpine-DAV/ascent/blob/258e06ed902d251c1f8eb4d987043d049b77cade/src/utilities/visit_session_converters/session_to_camera.py#L54

JustinPrivitera: @cyrush Do we know that that script is correct?

image

If view normal is the direction the camera is looking, then 

`position` + `view_normal * distance` = `focus`, so
`position` = `focus` - `view_normal * distance`

emily-howell: In Visit: 
image

This is what those same parameters look like in ascent when using an addition:
image

This is what those same parameters look like in ascent when using subtraction:
image

I am open to making changes if the math is wrong but that's what I am currently seeing,

JustinPrivitera: I stand corrected. Sorry about this. In VisIt the view normal is indeed pointing towards the camera from the focus. I should have checked in VisIt first.

emily-howell: Thanks for looking into that. Always good to double check and it made me take a second pass at my math which was good :)

src/docs/sphinx/Actions/Scenes.rst
```diff
@@ -402,6 +402,63 @@ Now we add a second render to the same example using every available parameter:
...
+- ``elevation`` : Rotates the camera around the right axis.
+
+VisIt Style Camera
+^^^^^^^^^^^^^^^^^^
+Alternatively, the camera can be defined using the VisIt camera parameters which follows a view centric approach. 
+Instead of positioning the camera explicitly, it defines the view using vectors like ``view_normal`` and ``view_up``, which describe the direction and orientation of the camera relative to a focal point.
+The following parameters are supported:
+
+- ``view_normal`` : A vector [x, y, z] indicating the direction from the camera toward the scene.
+- ``focus`` : A 3D point [x, y, z] indicating the point of focus in the scene.
```
JustinPrivitera: Do you support the VisIt camera in 2D? Or just in 3D?

src/docs/sphinx/Actions/Scenes.rst
```diff
@@ -402,6 +402,63 @@ Now we add a second render to the same example using every available parameter:
...
+- ``view_up`` : A vector indicating which direction is up in the rendered view.
+- ``view_angle`` : Field of view angle in degrees.
```
JustinPrivitera:
```suggestion
- ``view_angle`` : Field of view angle in degrees for perspective projection.
```

JustinPrivitera: I believe the VisIt `view_angle` parameter is half the VTKm `fov` parameter, so maybe the description here could be different.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_conduit_to_vtkm_parsing.cpp
```diff
@@ -191,6 +196,124 @@ parse_camera(const conduit::Node camera_node, vtkm::rendering::Camera &camera)
...
+      conduit::Node n;
+      camera_node["focus"].to_float64_array(n);
+      const float64 *coords = n.as_float64_ptr();
+      vtkmVec3f look_at(coords[0], coords[1], coords[2]);
+      camera.SetLookAt(look_at);
+  }
+
+  if(camera_node.has_child("view_normal") && 
+     camera_node.has_child("focus") && 
+     camera_node.has_child("view_angle") &&
```
JustinPrivitera, line 218: Why not set the VTKm camera `fov` since we have `view_angle` available?

src/docs/sphinx/Actions/Scenes.rst
```diff
@@ -402,6 +402,63 @@ Now we add a second render to the same example using every available parameter:
...
+- ``azimuth`` : Rotates the camera around the up axis.
+- ``elevation`` : Rotates the camera around the right axis.
+
+VisIt Style Camera
+^^^^^^^^^^^^^^^^^^
+Alternatively, the camera can be defined using the VisIt camera parameters which follows a view centric approach. 
+Instead of positioning the camera explicitly, it defines the view using vectors like ``view_normal`` and ``view_up``, which describe the direction and orientation of the camera relative to a focal point.
+The following parameters are supported:
+
+- ``view_normal`` : A vector [x, y, z] indicating the direction from the camera toward the scene.
```
JustinPrivitera: Like [we discussed before](https://github.com/Alpine-DAV/ascent/pull/1596/files#r2292040412), in VisIt the view normal is pointing towards the camera from the focus, not from the camera toward the scene.

src/docs/sphinx/Actions/Scenes.rst
```diff
@@ -402,6 +402,63 @@ Now we add a second render to the same example using every available parameter:
...
+- ``focus`` : A 3D point [x, y, z] indicating the point of focus in the scene.
+- ``view_up`` : A vector indicating which direction is up in the rendered view.
+- ``view_angle`` : Field of view angle in degrees.
+- ``parallel_scale`` : Scale used for parallel (orthographic) projections.
```
JustinPrivitera: You could add that `parallel_scale` is half of the image height in world space.

src/docs/sphinx/Actions/Scenes.rst
```diff
@@ -402,6 +402,63 @@ Now we add a second render to the same example using every available parameter:
...
+- ``near_plane`` : Distance from the ``focus`` point along the ``view_normal`` to the near clipping plane.
+- ``far_plane`` : Distance from the ``focus`` point along the ``view_normal`` to the far clipping plane.
+- ``image_pan`` : A 2-element array [x, y] specifying image-space panning.
+- ``image_zoom`` : Zoom factor for the image.
+
+
+Unsupported VisIt Camera Parameters
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+In addition to the above supported VisIt camera parameters, there are a number of VisIt camera parameters that are not completely translatable to the VTKm style camera and are therefore currently ignored
```
JustinPrivitera:
```suggestion
In addition to the above supported VisIt camera parameters, there are a number of VisIt camera parameters that are not completely translatable to the VTKm style camera and are therefore currently ignored.
```

src/docs/sphinx/Actions/Scenes.rst
```diff
@@ -402,6 +402,63 @@ Now we add a second render to the same example using every available parameter:
...
+
+.. Note::
+   These parameters are safely ignored if provided. They will not influence the rendered image but also will not produce errors.
+
+- ``perspective`` : Boolean flag (true/false) indicating whether to use perspective (true) or parallel (false) projection. In Ascent, only perspective projection is currently used so this flag is ignored.
+- ``eye_angle`` : Used in stereo rendering to control the eye separation angle. Ascent does not currently support stereo rendering, so this has no effect.
+- ``center_of_rotation_set`` : Boolean flag indicating whether center_of_rotation has been explicitly set. Not used in Ascent.
```
JustinPrivitera:
```suggestion
- ``center_of_rotation_set`` : Boolean flag indicating whether ``center_of_rotation`` has been explicitly set. Not used in Ascent.
```

src/docs/sphinx/Actions/Scenes.rst
```diff
@@ -402,6 +402,63 @@ Now we add a second render to the same example using every available parameter:
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+The native Ascent camera follows the VTKm camera model which focuses on defining the camera's physical parameters and orientation in 3D space.
+This camera-centric approach specifies where the camera is located, what it is looking at, and how it is oriented.
+The following parameters are supported:
+
+- ``2d`` : Enables 2D rendering mode.
+- ``look_at`` : A 3-element array [x, y, z] indicating the point the camera is directed toward.
+- ``position`` : A 3-element array [x, y, z] indicating the camera's position in space.
+- ``up`` : A 3-element vector that defines the up direction of the camera.
+- ``fov`` : Field of view in degrees for perspective projection.
```
JustinPrivitera: Just curious, how do these parameters change in 2d mode?

src/libs/ascent/runtimes/flow_filters/ascent_runtime_conduit_to_vtkm_parsing.cpp
```diff
@@ -191,6 +196,124 @@ parse_camera(const conduit::Node camera_node, vtkm::rendering::Camera &camera)
...
+  // ----------------------- //
+  // Visit camera parameters //
+  // ----------------------- //
+  if(camera_node.has_child("focus"))
+  {
+      conduit::Node n;
+      camera_node["focus"].to_float64_array(n);
+      const float64 *coords = n.as_float64_ptr();
+      vtkmVec3f look_at(coords[0], coords[1], coords[2]);
+      camera.SetLookAt(look_at);
```
JustinPrivitera: Is there a reason for this style? You could do this more simply like this:
```suggestion
      float64_accessor coords = camera_node["focus"].value();
      // data_accessors give you type-coerced data access.
      vtkmVec3f look_at(coords[0], coords[1], coords[2]);
      camera.SetLookAt(look_at);
```
You could use this for all of the cases to avoid making copies of all of the arguments.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_conduit_to_vtkm_parsing.cpp
```diff
@@ -191,6 +196,124 @@ parse_camera(const conduit::Node camera_node, vtkm::rendering::Camera &camera)
...
+      }
+
+      vtkmVec3f position(pos[0], pos[1], pos[2]);
+      camera.SetPosition(position);
+
+      // Clipping planes require positional information to be computed
+      if(camera_node.has_child("near_plane"))
+      {
+        vtkm::Range clipping_range = camera.GetClippingRange();
+        clipping_range.Min =std::max(0.001, distance + camera_node["near_plane"].to_float64());
```
JustinPrivitera: Why do this max? Is this to avoid the near clipping range going into the negative? Can VTKm handle that case? I think a smaller number than 0.001 would be better but maybe I'm not understanding. Should this case yield an error message instead? Or at least a warning that your near plane value is being discarded?

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rendering_filters.cpp
```diff
@@ -200,6 +189,44 @@ check_renders_surprises(const conduit::Node &renders_node)
...
+  c_visit_valid_paths.push_back("image_pan");
+  c_visit_valid_paths.push_back("image_zoom");
+  c_visit_valid_paths.push_back("perspective");
+  c_visit_valid_paths.push_back("eye_angle");
+  c_visit_valid_paths.push_back("center_of_rotation_set");
+  c_visit_valid_paths.push_back("center_of_rotation");
+  c_visit_valid_paths.push_back("axis_3d_scale_flag");
+  c_visit_valid_paths.push_back("axis_3d_scale");
+  c_visit_valid_paths.push_back("shear");
+  c_visit_valid_paths.push_back("window_valid");
```
JustinPrivitera, line 228: I guess this is the style everywhere? It might be worthwhile to consider using initializer lists for this.

JustinPrivitera (dismissed):

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rendering_filters.cpp
emily-howell: I don't disagree. However, I am trying to maintain the overall code style and make logic changes in this MR so I am inclined to leave it this way personally.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_conduit_to_vtkm_parsing.cpp
emily-howell: Yes, this is to prevent the clipping range from being <= to 0. I have seen 0.001 other places before but would be happy to reduce it to a smaller offset. Do you have a preferred value?

src/docs/sphinx/Actions/Scenes.rst
emily-howell: This is just in 3D. Is there a 2D visit camera?

JustinPrivitera: Yes, VisIt has a 2D and a 3D mode. Params are different in that case.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_conduit_to_vtkm_parsing.cpp
JustinPrivitera: Maybe `CONDUIT_EPSILON`?

src/libs/ascent/runtimes/flow_filters/ascent_runtime_rendering_filters.cpp
JustinPrivitera: Sounds good.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_conduit_to_vtkm_parsing.cpp
JustinPrivitera: But I think it would be good for the user to know if their near plane value is not going to be used.

src/docs/sphinx/Actions/Scenes.rst
cyrush: for `2d` mode, only `2d` is used. 
It is assumed to be an array with 4 entries.
```
[x0,x1,y0,y1]
```
We should add an example / notes about this to the docs

src/docs/sphinx/Actions/Scenes.rst
cyrush: Here are VisIt's 2d view atts:
```
>>> View2DAttributes()
windowCoords = (0, 1, 0, 1)
viewportCoords = (0.2, 0.95, 0.15, 0.95)
fullFrameActivationMode = Auto  # On, Off, Auto
fullFrameAutoThreshold = 100
xScale = LINEAR  # LINEAR, LOG
yScale = LINEAR  # LINEAR, LOG
windowValid = 0
```

The ascent case is equivalent to `windowCoords`.
I don't know if we have anything like viewport controls. 

Not sure we need to parse visit 2d if it will be the same as Ascents.

If they truly are the same, then we should have a docs case for 2d, and then the Ascent + VisIt specific cases for 3d.

src/docs/sphinx/Actions/Scenes.rst
cyrush: I think this is used beyond orthographic projections, its a general param. 
The name is strange, but Justin's description describes what it is.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_conduit_to_vtkm_parsing.cpp
cyrush: Lets use `CONDUIT_EPSILON` here b/c we could have some tiny spatial extents that could have scale around `.001`

cyrush (commented):
This looks great! 
I added a few comments and suggested minor changes.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_conduit_to_vtkm_parsing.cpp
emily-howell: I added a case where if the `view_angle` is present it does set the field of view. I wanted to do that separate of this if statement because it can happen independent of these other settings that require the distance be calculated.

src/docs/sphinx/Actions/Scenes.rst
```diff
@@ -402,6 +402,68 @@ Now we add a second render to the same example using every available parameter:
...
+
+- ``perspective`` : Boolean flag (true/false) indicating whether to use perspective (true) or parallel (false) projection. In Ascent, only perspective projection is currently used so this flag is ignored.
+- ``eye_angle`` : Used in stereo rendering to control the eye separation angle. Ascent does not currently support stereo rendering, so this has no effect.
+- ``center_of_rotation_set`` : Boolean flag indicating whether ``center_of_rotation`` has been explicitly set. Not used in Ascent.
+- ``center_of_rotation`` : A 3D point used as the pivot for interactive rotation in VisIt. Ascent does not support interactive camera manipulation, so this is ignored.
+- ``axis_3d_scale_flag`` : Boolean flag that enables non-uniform axis scaling (e.g., X:Y:Z aspect ratios). Not supported in Ascent.
+- ``axis_3d_scale`` : A 3D vector specifying per-axis scaling for non-uniform visualizations. No effect in Ascent.
+- ``shear`` : A 3-element array used to apply a shear transformation to the view. Not currently implemented.
+- ``window_valid`` : Internal VisIt flag to indicate whether the current window setup is valid. Irrelevant in Ascent and ignored.
```
cyrush, line 467: Docs look great!

cyrush (approved):
Looks great, thanks!

PR #1619, opened 2025-09-19
Add support for tracing multiple energy groups in Rover

siramok:
Resolves bullets in https://github.com/Alpine-DAV/ascent/issues/1513.
- Implements support for computing multiple energy group outputs within one rover execute.
- Converts `ROVER_ERROR` calls to `ASCENT_LOG_ERROR` calls, which correctly throw exceptions.
- Several tests were created, including one to validate MPI support.
- Improved/added some parallel error checking.

Lots still TODO:
- ~~Verify that this approach is what we want and mean by multiple energy group support.~~ Seems to be.
- ~~Add more tests, particularly MPI tests (I haven't tried it with MPI yet).~~ Seems to work fine with MPI.
- ~~Add validation somewhere to ensure that if we have more than one energy group, that the number of absorption and emission fields are the same.~~ We now check for this in at least 2 places.
- ~~Each domain is only traced once, but there seems to be some modest overhead compared to doing multiple rover execute calls with one energy group at a time. My guess is that this is due to deinterleaving optical depth and intensity fields at the end (the access pattern is probably not cache friendly). There's probably a smarter way to deal with it.~~ The performance improvement is significant.
- ~~Some of the baseline images using the `image_topo` topology seem to have abnormally large extents, currently not sure why.~~ This was apparently due to slicing on a boundary edge with z = 0. Bumping this up to z = 0.001 produces a render with the expected extents.


src/libs/rover/engine.cpp
```diff
@@ -117,39 +117,11 @@ Engine::partial_trace(Ray64 &rays, PartialVector64 &partials)
...
-    ROVER_ERROR("Error - Engine::get_num_channels: absorption field size is not evenly divided by num_cells"
-                << "\n       modulo " << modulo
-                << "\n       num cells " << num_cells
-                << "\n       field size " << absorption_size);
-    throw RoverException("absorption field size is not evenly divided by num_cells\n");
-  }
-  vtkm::Id num_bins = absorption_size / num_cells;
-  ROVER_INFO("Engine::get_num_channels: Detected " << num_bins << " bins");
+  const auto &field = m_dataset.GetField(absorption);
+  vtkm::Id num_bins = field.GetData().GetNumberOfComponentsFlat();
```
JustinPrivitera: You could check here that the emission field also has the same number of components.

Regardless, it is nice to see all this old code go away and be replaced by our new, simpler approach.

src/libs/rover/image.cpp
```diff
@@ -272,7 +272,8 @@ Image<FloatType>::flatten_intensity_values()
 #endif
     for(int i = 0; i < size; ++i)
     {
-      output.Set( i * num_channels + c, channel.Get(i));
+      // Deinterleave the output: all pixels for group 0, then all pixels for group 1, etc.
+      output.Set(c * size + i, channel.Get(i));
```
JustinPrivitera, line 276: The old way that this was doesn't make sense to me, but the new way you have here looks correct. I'm not sure how this worked before but I think we are on the right track.

src/libs/rover/typed_scheduler.cpp
```diff
@@ -59,37 +59,37 @@ TypedScheduler<FloatType>::set_ray_generator(RayGenerator *ray_generator)
...
   // Initialize background intensities to 0.0f (by default)
   const float64 background_intensity = rover::settings["background_intensity"].to_float64();
-  m_background.resize(num_channels, background_intensity);
+  m_background.resize(num_energy_groups, background_intensity);
 }
 
 template<typename FloatType>
 int
-TypedScheduler<FloatType>::get_global_channels()
+TypedScheduler<FloatType>::get_global_num_energy_groups()
```
JustinPrivitera, line 72: I agree that we need to use MPI to figure out if all the processors agree on how many groups there are, but I think if any of them disagree then there should be an error. I haven't stared enough at the code here to tell if that's what happens but I don't see any error case. Do we error in this case?

src/libs/rover/typed_scheduler.cpp
```diff
@@ -752,6 +759,9 @@ TypedScheduler<FloatType>::write_blueprint_rays_mesh(Node &data_out,
       scaled_unit_left = dx * vtkm::Normal(left);
       scaled_unit_up = dy * vtkm::Normal(up);
 
+#ifdef ROVER_OPENMP_ENABLED
+      #pragma omp parallel for collapse(2)
+#endif
       for (int j = 0; j < image_width; j++)
```
JustinPrivitera, line 797: Thanks for all this added parallelization.

src/libs/vtkh/rendering/ConnectivityTracer.cpp
```diff
@@ -724,38 +699,44 @@ class Integrate : public vtkm::worklet::WorkletMapField
...
-      BOUNDS_CHECK(absorbtionData, cellOffsetI);
-      FloatType absorb = static_cast<FloatType>(absorbtionData.Get(cellOffsetI));
+      FloatType absorb = static_cast<FloatType>(VecTraits::GetComponent(cellValue, i));
       absorb *= UnitScalar;
 
       const int rayOffsetI = rayOffset + i;
       BOUNDS_CHECK(opticalDepthBins, rayOffsetI);
-      FloatType opticalDepth = static_cast<FloatType>(opticalDepthBins.Get(rayOffsetI));
+      FloatType opticalDepth = static_cast<FloatType>(opticalDepthBins.Get(rayOffsetI));      
       opticalDepthBins.Set(rayOffsetI, opticalDepth + absorb * segmentLength);
```
JustinPrivitera, line 720: I don't really understand what these changes are about. It looks like we changed the approach for absorption but not for optical depth?

src/libs/vtkh/rendering/ConnectivityTracer.cpp
```diff
@@ -796,9 +777,23 @@ class IntegrateEmission : public vtkm::worklet::WorkletMapField
...
+    
+    BOUNDS_CHECK(absorptionData, currentCell);
+    BOUNDS_CHECK(emissionData, currentCell);
+    AbsValueType absCellValue = absorptionData.Get(currentCell);
+    EmisValueType emisCellValue = emissionData.Get(currentCell);
+    
+    // Check if these are vector fields or scalar fields
+    vtkm::IdComponent absNumComponents = AbsVecTraits::GetNumberOfComponents(absCellValue);
+    vtkm::IdComponent emisNumComponents = EmisVecTraits::GetNumberOfComponents(emisCellValue);
+
```
JustinPrivitera: I'm just trying to understand... is this a VTKm pattern for getting this kind of information? I don't see where all of these things are used.

src/tests/_baseline_images/tout_rover_xray_blueprint_multiple_groups_intensities_spatial_z0_000048.png
JustinPrivitera, line 1: I think the image should look thick, especially in the spatial topo case. Maybe we need to add logic to the blueprint mesh write to give it a 3rd dimension? This doesn't look how I'd expect.

JustinPrivitera: Oh I understand now. I should have looked at the tests first. I see that you are taking slices.

src/tests/_baseline_images/tout_rover_xray_blueprint_multiple_groups_intensities_z0_000048.png
JustinPrivitera, line 1: Why is this zoomed in like this?

src/tests/_baseline_images/tout_rover_xray_blueprint_multiple_groups_intensities_spatial_z0_000048.png
JustinPrivitera: The field extents numbers look good which is great to see. Nice job.

JustinPrivitera: I think we should add a test that shows the entire topology. Not useful for someone who wants to see results for a specific energy group, but useful to us for understanding that we set up the output mesh correctly. Although the correct slices are a nice confirmation, it is also good to check and see that everything looks good.

JustinPrivitera: I think I would also say that for the slices, we don't need to have rotated variants of them since we trust that Ascent slice will do its job. I don't know how much things have been abstracted away and if that is a simple change or not.

src/tests/ascent/rover_test_utils.hpp
```diff
@@ -129,6 +129,114 @@ render_fields(const Node &data,
...
+            scenes[scene_name]["plots"][plot_name]["type"] = "pseudocolor";
+            scenes[scene_name]["plots"][plot_name]["field"] = field;
+            scenes[scene_name]["plots"][plot_name]["pipeline"] = pipeline_name;
+            scenes[scene_name]["renders"][render_name]["image_prefix"] = full_output_path;
+
+            // Rotate spatial meshes for test image variety
+            if (field.find("spatial") != std::string::npos)
+            {
+                scenes[scene_name]["renders"][render_name]["camera"]["azimuth"] = 45.0;
+            }
```
JustinPrivitera: Yeah so for the slices I don't think we need the rotation. But we should add a test that does the whole (unsliced) mesh and those should be rotated so we can see the depth.

JustinPrivitera (changes requested):
This is really fantastic work. Thanks so much for this. I just had a few clarifying comments and questions.

src/tests/ascent/rover_test_utils.hpp
siramok: I went ahead and removed rotation for slices and added some whole-mesh renders.

src/tests/_baseline_images/tout_rover_xray_blueprint_multiple_groups_intensities_z0_000048.png
siramok: As best I can tell, it seems to be due to using z = 0 in the slice. Using z values close to 0 (0.1, 0.01, or 0.001) have the same legend values as z = 0 but are not zoomed out, so that's my current "fix" for this. Perhaps this is an ascent bug when slicing precisely on a boundary edge?

src/libs/vtkh/rendering/ConnectivityTracer.cpp
siramok: Totally fair to ask about. The [viskores source explains what `VecTraits` do](https://github.com/Viskores/viskores/blob/9d59acb38e970015e1f063d0bb9e9e68430e5f31/viskores/VecTraits.h#L55), but it essentially lets us treat the input cell data as either being 1D or 2D without having to know its actual shape ahead of time.

In a single-field world, absorptionData and emissionData are 1D vectors where each element is a single scalar value. In a multi-field world, each additional field is interleaved such that absorptionData and emissionData are essentially 2D vectors of arbitrary length. All of the 0th absorption values, then all of the 1st absorption values, etc.

What we had in place before assumed a single-field world, so we were accessing the data cells directly, and IIRC viskores wasn't happy about that in the multi-field case.

src/libs/vtkh/rendering/ConnectivityTracer.cpp
siramok: IIRC this access pattern was only necessary for the input cell data, not the output buffers. I could double check that it's actually necessary, maybe my issue was being caused by something else.

src/libs/rover/typed_scheduler.cpp
siramok: Faster is better!

src/libs/rover/typed_scheduler.cpp
siramok: We did not previously error in this scenario, but I added a check to ensure that all MPI ranks agree that they have the same number of energy groups. It now errors if they don't agree.

src/libs/rover/image.cpp
siramok: I don't think it mattered for the single-field case, but for multi-field, the old line writes the output as-is (interleaved) which was causing the renders to look not as expected.

src/libs/rover/engine.cpp
siramok: I agree! I got rid of the auto type, didn't mean to leave it in. I also added a check to validate that the emission field has the same number of components, which throws an error if they don't agree.

siramok:
Failing Windows CI seems to be unrelated to my latest commit.

Next thing I need to do is implement those additional tests, particularly to make sure that this works with MPI. As far as my comment about performance, I realized I was going by the total test time which includes time spent creating the additional fields. I need to measure just the ascent execute calls for a proper comparison.

src/libs/vtkh/rendering/ConnectivityTracer.cpp
siramok: It does indeed seem to be necessary. The input cell data elements are `vtkm::Vec`s when passing multi-field data, while the output buffers are 1D arrays with enough memory allocated to hold all of the output.

siramok:
Seems to work out of the box with MPI. Separately, I realized that `ROVER_ERROR` wasn't actually throwing an exception, so I've converted those into `ASCENT_LOG_ERROR` calls which does throw an exception. The benefit there is that we can actually test for errors now, like making sure that we error if absorption and emission don't have the same number of components. The other benefit is that rover won't just keep running until it or viskores segfaults.

siramok:
And finally, after looking at just the `ascent.execute` time to run rover, the performance uplift with multi-group support seems to be great. I measured an average of 0.3302 seconds when tracing curv3d with one energy group. Extrapolating, that would be about 0.9906 seconds to trace 3 separate curv3d energy groups one at a time (the only way that was possible before this PR). I then measured an average of 0.5617 seconds to trace 3 energy groups at once with this PR, roughly a 56% improvement. Using the same methodology but with 1024x1024 traces instead of 200x200, it's closer to a 66% improvement.

src/libs/rover/engine.cpp
JustinPrivitera: cool

src/libs/rover/image.cpp
JustinPrivitera: makes sense

src/libs/rover/typed_scheduler.cpp
JustinPrivitera: great, thanks!

src/libs/vtkh/rendering/ConnectivityTracer.cpp
JustinPrivitera: My confusion was 
1. why things were different between absorption and optical depth
2. what these changes were for

I think the answer to 1. is that optical depth is the same for all groups, so we don't need to use the same method of fetching it for the output as we do for absorption?

And then I think I understand what this code is doing now.

src/libs/vtkh/rendering/ConnectivityTracer.cpp
JustinPrivitera: Gotcha. Thank you for the explanation!

src/tests/_baseline_images/tout_rover_xray_blueprint_multiple_groups_intensities_z0_000048.png
JustinPrivitera: Part of me wonders if we should open a ticket, but at the same time, it seems questionable in general to slice on a boundary edge. For the VisIt tests, I calculated the midpoints of the energy groups to take the slices. Glad you got this figured out.

src/tests/ascent/rover_test_utils.hpp
JustinPrivitera: thanks!

src/libs/rover/engine.cpp
```diff
@@ -120,9 +121,26 @@ int
...
+  // as the absorption field
+  if (rover::settings.has_child("emission"))
+  {
+    const std::string emission = rover::settings["emission"].as_string();
+    const vtkm::cont::Field &emission_field = m_dataset.GetField(emission);
+    vtkm::Id num_emission_bins = emission_field.GetData().GetNumberOfComponentsFlat();
+
+    if (num_absorption_bins != num_emission_bins)
+    {
+      ASCENT_LOG_ERROR("Error - Engine::get_num_energy_groups: number of energy groups in absorption field ("
```
JustinPrivitera: I wonder if we need parallel error checking for a case like this. If this gets called by multiple ranks and one fails then we are in trouble. Although its very unlikely we would have a mismatch here.

src/libs/rover/typed_scheduler.cpp
```diff
@@ -81,9 +82,16 @@ TypedScheduler<FloatType>::get_global_num_energy_groups()
...
+  int mpi_max_energy_groups;
+  MPI_Allreduce(&num_energy_groups, &mpi_max_energy_groups, 1, MPI_INT, MPI_MAX, m_comm_handle);
+
+  // Check that all ranks have the same num_energy_groups
+  if (num_energy_groups != mpi_max_energy_groups)
+  {
+    ASCENT_LOG_ERROR("Error - TypedScheduler::get_global_num_energy_groups: MPI ranks have inconsistent number of energy groups. "
+                "Local: " << num_energy_groups << ", Global max: " << mpi_max_energy_groups);
+  }
+
```
JustinPrivitera: This could cause a parallel hang. If we want all ranks to exit we should grab the max and min and compare them. Then all ranks will error or not error as needed.

src/libs/vtkh/rendering/ConnectivityTracer.cpp
JustinPrivitera: Your changes to the names helped me to understand :)

src/tests/_baseline_images/tout_rover_xray_blueprint_multiple_groups_absorption_only_optical_depth_spatial_full_000048.png
JustinPrivitera, line 1: I guess the z axis labels are hidden behind the mesh? It would be nice to know that the spatial extents match our assumptions.

src/tests/_baseline_images/tout_rover_xray_blueprint_multiple_groups_intensities_full_000048.png
JustinPrivitera, line 1: Do we want a rotated version of this mesh as well, so we can see that it is 3d?

src/tests/ascent/t_ascent_mpi_rover.cpp
```diff
@@ -385,6 +384,90 @@ TEST(ascent_rover, test_xray_mpi_blueprint_braid_uniform_single_domain_multiple_
...
+    const std::string query_path = utils::join_file_path(output_path, query_name);
+    const std::string output_data_path = query_path + query_suffix;
+
+    // Remove old test data
+    const int cycle = 48;
+    remove_rover_test_data(query_path, query_suffix, cycle);
+
+    // Generate test data for MPI
+    Node test_data;
+    get_multi_group_curv3d_data(test_data);
```
JustinPrivitera, line 420: Doesn't this mean that every rank owns the same data? We might want to be more creative for the MPI tests.

src/tests/ascent/t_ascent_rover.cpp
```diff
@@ -825,14 +824,130 @@ TEST(ascent_rover, test_xray_blueprint_multiple_groups)
...
+
+    // Define Ascent actions
+    Node extracts;
+    get_common_extract_params(extracts, query_path, "d_multi", "p_multi");
+
+    Node actions;
+    get_default_action_params(actions, extracts);
+
+    // Execute Ascent actions
+    EXPECT_THROW(execute_ascent(test_data, actions), conduit::Error);
```
JustinPrivitera, line 944: ❤️

JustinPrivitera (changes requested):
Just some minor things about parallel error checking and MPI tests.

src/tests/ascent/t_ascent_mpi_rover.cpp
JustinPrivitera: Actually I guess you use mpi load mesh to load it appropriately in the parallel case. So this is correct, but it's not a very interesting MPI case since only one rank has data. We should make another test with `multi_curv3d` that is multi-group and use that for the MPI tests.

src/tests/_baseline_images/tout_rover_xray_blueprint_multiple_groups_intensities_spatial_z0_000048.png
siramok: Just cleaning up the thread. I believe we now have a test that shows the entire topology and the rotated variants of the slice images should also be removed, so I'll mark this feedback as resolved.

src/tests/_baseline_images/tout_rover_xray_blueprint_multiple_groups_intensities_z0_000048.png
siramok: I guess we can defer to @cyrush on whether or not this counts as a bug (context: getting no output when slicing precisely on a boundary). If so, I'd be happy to open an issue.

To your second point, I may as well change the current implementation to use the midpoints also. I don't think it will make a difference in the baselines, but consistency with VisIt is good.

src/tests/_baseline_images/tout_rover_xray_blueprint_multiple_groups_absorption_only_optical_depth_spatial_full_000048.png
siramok: I think adding elevation in addition to azimuth should make all of the axes visible, I'll give that a try and see how it looks.

src/tests/_baseline_images/tout_rover_xray_blueprint_multiple_groups_intensities_full_000048.png
siramok: Interesting point, I'll give it a try.

src/tests/ascent/t_ascent_mpi_rover.cpp
siramok: It's been a few weeks so I'll definitely double check myself and follow up with you, but IIRC I thought I had verified that domains were being loaded evenly across ranks. But to your point, this isn't `multi_curv3d`, and wasn't the difference between the two supposed to be that one was split into multiple domains while the other wasn't? If that's not the case in reality, then I guess all of our `curv3d/multi_curv3d` tests are effectively duplicates? I'll verify what's going on there.

siramok: False alarm. Upon double checking, the helper function is called `get_multi_group_curv3d_data` but in reality it loads `multi_curv3d` and not `curv3d`. Furthermore, I verified that each rank does get a balanced number of multi-group domains in the existing MPI test.

src/tests/_baseline_images/tout_rover_xray_blueprint_multiple_groups_intensities_full_000048.png
siramok: I added rotation and elevation to these meshes as well, and they are indeed 3D, so I think it's okay to keep the change.

src/tests/_baseline_images/tout_rover_xray_blueprint_multiple_groups_absorption_only_optical_depth_spatial_full_000048.png
siramok: Adding elevation did indeed make all of the axes visible.

src/libs/rover/typed_scheduler.cpp
siramok: Fixed, it now grabs the min and max and compares them per your suggestion.

src/libs/rover/engine.cpp
siramok: I solved this, but it was a slightly more invasive change than expected, so let me know what you think. Adding parallel error checking in-place (inside of `engine.cpp`) didn't work because that was causing the MPI test which puts all of the data on rank 0 to fail. It turns out that the ranks that didn't have any data would never reach the `MPI_Allreduce`, causing rank 0 to hang. I was admittedly pleased to see that ranks with nothing to do weren't bothering to do the setup work required before tracing rays.

My workaround was to move the parallel error check to `TypedScheduler` in a place that we know all of the ranks will reach and execute. We still check for a mismatched number of fields inside of `engine.cpp`, but instead of throwing an error right away, detecting a mismatch sets a member variable. Then inside of `TypedScheduler`, each rank loops over each of its domains and checks to see if any of them had a mismatch. Ranks that had no domains won't have anything to iterate over, and will therefore not have any mismatches.

I don't think we currently have a test that exercises this, but I could come up with one if you'd like.

src/tests/ascent/t_ascent_mpi_rover.cpp
JustinPrivitera: Oh I'm sorry, I didn't ready closely. Yes, it does have `multi` in the name. My bad.

src/tests/_baseline_images/tout_rover_xray_blueprint_multiple_groups_intensities_full_000048.png
JustinPrivitera: great!

src/tests/_baseline_images/tout_rover_xray_blueprint_multiple_groups_absorption_only_optical_depth_spatial_full_000048.png
JustinPrivitera: cool, thanks!

src/libs/rover/typed_scheduler.cpp
JustinPrivitera: great, thanks!

src/tests/_baseline_images/tout_rover_xray_blueprint_multiple_groups_intensities_z0_000048.png
siramok: Forgot to follow up, but I did switch to slicing at the midpoints instead.

JustinPrivitera: I don't think slicing on a boundary is an important case to worry about. I like using midpoints.

src/libs/rover/engine.cpp
JustinPrivitera: It is always best to make tests to capture all the cases we can, but we aren't always good about it. If you want to add a test you can but I'm not even sure how you would cause this. Num absorption bins would need to be different than num emission bins on only some ranks. I imagine we would hit a Conduit Relay reading error before we ever made it here.

JustinPrivitera (approved):
Thanks so much for getting this in!

PR #1621, opened 2025-09-23
Task/9 18 25 1607 move from vtk m to viskores

### LINKED ISSUE #1607: move from vtk-m to viskores

emily-howell:


scripts/uberenv_configs/packages/ascent/package.py
emily-howell, line 1: These changes are very much wrong

src/examples/paraview-vis/paraview-package-momentinvariants.patch
```diff
@@ -24,21 +24,21 @@ index 3d60ee9a5..97d518391 100644
...
+diff --git a/var/spack/repos/builtin/packages/viskores/package.py b/var/spack/repos/builtin/packages/viskores/package.py
 index b899ef547..3317766a6 100644
---- a/var/spack/repos/builtin/packages/vtkm/package.py
-+++ b/var/spack/repos/builtin/packages/vtkm/package.py
+--- a/var/spack/repos/builtin/packages/viskores/package.py
++++ b/var/spack/repos/builtin/packages/viskores/package.py
 @@ -20,7 +20,7 @@ class Vtkm(CMakePackage, CudaPackage):
-     url      = "https://gitlab.kitware.com/api/v4/projects/vtk%2Fvtk-m/repository/archive.tar.gz?sha=v1.3.0"
-     git      = "https://gitlab.kitware.com/vtk/vtk-m.git"
+     url      = "https://gitlab.kitware.com/api/v4/projects/vtk%2Fviskores/repository/archive.tar.gz?sha=v1.3.0"
```
emily-howell: Could not find a gitlab project for this. Need to find the github equivalent

scripts/uberenv_configs/packages/ascent/ascent-shared-cuda-pr903.patch
```diff
@@ -238,19 +238,19 @@ index fe9ce937..7f94a82b 100644
...
+--- a/src/cmake/thirdparty/SetupViskores.cmake
++++ b/src/cmake/thirdparty/SetupViskores.cmake
+@@ -26,6 +26,11 @@ if(ENABLE_CUDA AND NOT Viskores_ENABLE_CUDA)
+    message(FATAL_ERROR "Ascent CUDA support requires Viskores with CUDA support (ENABLE_CUDA == TRUE, however Viskores_ENABLE_CUDA == FALSE")
  endif()
  
 +if(ENABLE_CUDA AND BUILD_SHARED_LIBS)
-+  if(VTKm_VERSION VERSION_LESS "1.7.0")
-+    message(FATAL_ERROR "Cannot build shared libs with CUDA when VTKm is < v1.7.0")
++  if(Viskores_VERSION VERSION_LESS "1.7.0")
```
nicolemarsaglia: Viskores version 1.0.0

scripts/uberenv_configs/packages/ascent/package.py
```diff
@@ -214,26 +214,26 @@ class Ascent(CMakePackage, CudaPackage):
...
-    # VTK-m
+    # Viskores
     #######################
     with when("+vtkh"):
-        depends_on("vtk-m +doubleprecision ~64bitids")
-        depends_on("vtk-m@2.1:", when="@0.9.3:")
-        depends_on("vtk-m@2.0:", when="@0.9.2:")
+        depends_on("viskores +doubleprecision ~64bitids")
+        depends_on("viskores@2.1:", when="@0.9.3:")
+        depends_on("viskores@2.0:", when="@0.9.2:")
```
nicolemarsaglia: viskores is version is 1.0 which is equivalent to vtkm 2.3

scripts/uberenv_configs/packages/ascent/package.py
```diff
@@ -249,9 +249,9 @@ class Ascent(CMakePackage, CudaPackage):
         depends_on("vtk-h~cuda", when="@:0.8.0 ~cuda")
         depends_on("vtk-h+shared", when="@:0.8.0 +shared")
         depends_on("vtk-h~shared", when="@:0.8.0 ~shared")
-        # When using VTK-h ascent also needs VTK-m
-        depends_on("vtk-m@:1.7", when="@:0.8.0")
-        depends_on("vtk-m+testlib", when="@:0.8.0 +test")
+        # When using VTK-h ascent also needs Viskores
+        depends_on("viskores@:1.7", when="@:0.8.0")
```
nicolemarsaglia: 1.0 here as well but I'm not sure how that changes the cuda dependency. This will be good for @cyrush to look at

emily-howell: I think this one needed to be put back to vtk-m?

nicolemarsaglia: but I don't think we'll be building vtkm, so there's nothing for it to depend on. I think we stop supporting these cases since we don't have a lower version of viskores?

emily-howell: So, yes, but isn't the @:0.8.0 the ascent version? My best guess at what the file was doing was configuring dependencies for ascent as a bigger project, not just the current version?

cyrush: that is correct 0.8.0 is the ascent version -- but not sure the value of keeping the the logic for very old versions in this spack package. 

We can have a clean slate and we can remove some of the old logic?

If we merge to spack proper they may have opinions -- but they seem to be deprecating and removing things based on a recent interaction where they removed the dray package from spack.

src/examples/paraview-vis/paraview-package-momentinvariants.patch
cyrush: I believe -- Unlike VTK-m Viskores is developed on github:

https://github.com/Viskores/viskores/

I think the paraview examples have atrophied quite a bit, not sure we need to try to update them

scripts/build_ascent/build_ascent.sh
```diff
@@ -653,7 +653,7 @@ cmake -S ${kokkos_src_dir} -B ${kokkos_build_dir} ${cmake_compiler_settings} \
 
 echo "**** Building Kokkos ${kokkos_version}"
 cmake --build ${kokkos_build_dir} --config ${build_config} -j${build_jobs}
-echo "**** Installing VTK-m ${kokkos_version}"
+echo "**** Installing Viskores ${kokkos_version}"
```
cyrush:
```suggestion
echo "**** Installing Kokkos ${kokkos_version}"
```
This looks like a result of a typo by me earlier. This is Kokkos not VTK-m / Viskores

src/cmake/thirdparty/SetupViskores.cmake
```diff
@@ -0,0 +1,83 @@
...
+#viskores default type check
+if(NOT Viskores_USE_DEFAULT_TYPES_FOR_ASCENT)
+  message(FATAL_ERROR "Ascent requires Viskores to be built with default types (Viskores_USE_DEFAULT_TYPES_FOR_ASCENT=ON)")
+endif()
+
+if(ENABLE_CUDA AND NOT Viskores_ENABLE_CUDA)
+   message(FATAL_ERROR "VTK-h CUDA support requires Viskores with CUDA support (ENABLE_CUDA == TRUE, however Viskores_ENABLE_CUDA == FALSE")
+endif()
+
+if(ENABLE_CUDA AND BUILD_SHARED_LIBS)
```
cyrush: we can remove this test, it was needed for very old versions of vtk-m, all versions of viskores should be ok

emily-howell:
Pushed up another round of minor fixes and cleanup.

I got the containers all built and running and ran a few with azemu to check that the tests pass. Locally, I did make some modifications to azure-pipelines.yml (changed c++14 to c++17 and updated the container names) that I have not pushed up because I figure those changes should get pushed with the containers?

@cyrush Sounds like you want to build and push the updated containers? Let me know if you need anything from me with that.

cyrush:
@emily-howell thanks for getting this all sorted!

I will build the containers -- but I don't want to miss the cxx changes in azure pipelines. Can you add those now (even though they will fail). We can also comment out all az pipelines to avoid running right now. When I get the containers ready I will use this branch and we can turn them back on when the PR is ready.

scripts/azemu/azemu.py
emily-howell, line 1: Sorry, These changes snuck into the PR. These were from suggestions nicole had to make running the containers locally work but should not be pushed up.

cyrush: no worries, thanks for noticing and resolving

azure-pipelines.yml
```diff
@@ -59,8 +59,8 @@ stages:
           python scripts/ci/check_blt_time_travel.py
         displayName: 'BLT Time Travel Check'
 
-# main checks build using docker containers that
-# include tpls built with uber
+main checks build using docker containers that
+include tpls built with uber
```
emily-howell: Should these lines still be comments?

```suggestion
# main checks build using docker containers that
# include tpls built with uber
```

emily-howell: Yep, now the azure pipeline seems to be running :)

cyrush: thanks!

cyrush:
Looks like dray has some cuda issues, those would be unrelated to viskores, but may side effect of container update, I'll take a look

emily-howell:
Looks like the only thing that failed was the dray-cuda issue! So that's a win at least.

cyrush:
@emily-howell  Working on the mystery. 
I think the cuda arch is not getting passed, so it's trying to compile for all cuda archs. This may be b/c we are using a newer cmake.

cyrush:
@emily-howell -- that was the issue: the cuda arch was not being passed and it was trying to generate code for very old GPU architectures. 

I suspect (but didn't find a smoking gun) that these flags were implicitly forwarded some how in the vtk-m case, and that changed slightly for viskores.

emily-howell:
Interesting. I was under the impression that the last version of vtk-m and viskores 1.0 were supposed to be identical. I wonder if I missed something in the uberenv config instructions?

cyrush (approved):
Thank you for this!

PR #1632, opened 2025-11-13
uniform grid sampling mpi hang, fix default extents, add box sample

nicolemarsaglia:


nicolemarsaglia:
the yaml from the box unit test in t_ascent_sample.cpp

10x10x10_yaml

No matter what I change the dims to, I end up with something like this, which is very odd. 

10x10x10_weird_diagonal

nicolemarsaglia:
tout_sample_box_3d_000100_new
I have ascent making an image with the correct points.

nicolemarsaglia:
extract_diagonal_still
weirdly the extract is still diagonal

src/docs/sphinx/Actions/Pipelines.rst
```diff
@@ -941,9 +941,88 @@ the data set has more than one domain. Without ghost, the averaging will not be
...
+  params["field"] = "braid";   //required
+  params["box/dims/i"] = 10.0; //default: 1.0 
+  params["box/dims/j"] = 10.0; //default: 1.0 
+  params["box/dims/k"] = 10.0; //default: 1.0 
+  params["box/min/x"]  = 0.0;  //default: (x_max - x_min)/2.0 
+  params["box/min/y"]  = 0.0;  //default: (y_max - y_min)/2.0 
+  params["box/min/z"]  = 0.0;  //default: (z_max - z_min)/2.0 
+  params["box/max/x"]  = 0.0;  //default: (x_max - x_min)/2.0 
+  params["box/max/y"]  = 0.0;  //default: (y_max - y_min)/2.0 
+  params["box/max/z"]  = 0.0;  //default: (z_max - z_min)/2.0
```
cyrush: we have min == max here and I think we need to change the comment.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_vtkh_filters.cpp
```diff
@@ -4142,6 +4175,151 @@ VTKHSample::execute()
...
+            x_min = g_bounds.X.Min;
+          }
+          else
+          {
+            x_min = min_b["x"].as_float64();
+          }
+        }
+        else //not set; default (max-min)/2
+        {
+          x_min = (g_bounds.X.Max - g_bounds.X.Min)/2.0;
```
cyrush: We do we set min to the center? Shouldn't it just be the min of the dataset?

cyrush: We want the default case to sample the entire spatial extents.

cyrush (commented):
A question about box sample input params.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_vtkh_filters.cpp
nicolemarsaglia: lol can do. in an Ascent meeting you told me to do this. or it was a fever dream. something something this is what visit does?

cyrush: I think I fumbled the context when explaining it. 

We do want to be like visit, and visit samples the entire extents by default:

Screenshot 2025-12-12 at 11 31 40 AM

origin will be min and the spacing + number of samples should cover up to max

nicolemarsaglia: Sounds good! I probably misunderstood.

cyrush (approved):

PR #1634, opened 2025-12-04
Sycl build fixes + 64bit ids support

mvictoras:
Various fixes to make ascent compile on Aurora:

- Updated `build_ascent_aurora.sh` to use the correct modules
- Fixed `src/libs/vtkh/vtkh.cpp` to return the correct `device_count`
- Added support for 64bit ids to `build_ascent.sh` (it was needed for a science user with data over 1TB)


mvictoras:
Dependent on: https://github.com/Viskores/viskores/pull/233

cyrush:
@mvictoras  I think we resolved the 64-bit id issue in #1641. Wondering if you have had success since then, or if we need more sycl or 64-bit id fixes.

PR #1636, opened 2025-12-12
Pin GHA dependencies by hash

williamjallen:
Hash-pinned dependencies are an important requirement enforced by the OpenSSF scorecard tool: https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies

This commit pins all GitHub Actions workflows to a commit hash.

### DIFF
--- a/.github/workflows/build_ascent_cuda.yml
+++ b/.github/workflows/build_ascent_cuda.yml
@@ -13,7 +13,7 @@ jobs:
       CMAKE_VERSION: 3.24.4
     steps:
     - name: Checkout
-      uses: actions/checkout@v4
+      uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
       with:
         submodules: 'recursive'
     - name: Install CMake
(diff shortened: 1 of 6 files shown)


cyrush:
@williamjallen  thanks for this report. 

I don't understand the bad practice of using a version instead of a specific hash in the case of github actions provided by github.

If pinned to version 4, or the hash, and something wrong is discovered aren't we are exposed either way?

cyrush:
Ok, I understand. If someone takes over they can change the version to a malicious commit.
Still feels like if github's provided actions are compromised, there might be scarier things afoot.

williamjallen:
@cyrush In the case of GitHub Actions provided by GitHub, you're right that the odds of a tag getting switched to a malicious commit are extremely low.  That said, it certainly can't hurt to hash-pin everything across the board.  That way, you can add CI checks to verify that all new actions have their dependencies pinned.

For 3rd party actions, there have been several cases of tags getting reassigned to malicious commits in major dependencies, most famously [tj-actions/changed-files](https://www.cisa.gov/news-events/alerts/2025/03/18/supply-chain-compromise-third-party-tj-actionschanged-files-cve-2025-30066-and-reviewdogaction), and this is a threat to the community at large even if an individual project is unlikely to be targeted directly.

cyrush (approved):
Verified 34e114876b0b11c390a56381ad16ebd13914f8d5 is 
[v4.3.1](https://github.com/actions/checkout/releases/tag/v4.3.1)

PR #1638, opened 2025-12-17
Fix the parsing of ypan in the ascent camera.

brugger1:
I fixed a bug in the parsing of `ypan` in the ascent camera specification. The code previously set the value of `xpan` when it encountered `ypan` in the camera node, probably the result of copying and pasting code and not updating he pasted code completely. The code now sets `ypan` when it encounters `ypan` in the camera node.

I tested the code by setting a camera and varying the `xpan` and `ypan` values and viewing the saved images.


cyrush (approved):
thanks for finding and fixing this!

PR #1642, opened 2025-12-19
Adding cycles directory check to replay input parameters handling

### LINKED ISSUE #1591: replay: passing dir instead of root file crashes
need to make this an error message, not a crash

emily-howell:


src/utilities/replay/replay.cpp
```diff
@@ -109,6 +110,28 @@ struct Options
       usage();
       exit(1);
     }
+
+    if(!m_cycles_file.empty())
+    {
+      struct stat st;
+      if(stat(m_cycles_file.c_str(), &st) != 0)
```
cyrush: To simplify, you can use:


```
conduit::utils::is_file 
# and
conduit::utils::is_directory
````

cyrush (approved):
Looks good!

PR #1646, opened 2026-01-05
Add support for tiled rendering.

brugger1:
I added support for tiled rendering. It is controlled by 4 new fields in the renderer.

They are:

```
renders/r1/tiled_rendering=[true|false]
renders/r1/tiled_rendering_type=[square_tiles|
                                 rectangular_tiles|
                                 horizontal_strips|
                                 vertical_strips|
                                 optimized_tiles]
renders/r1/tile_width=int
renders/r1/tile_height=int
```

The default values are `true`, `optimized_tiles`, `1024` and `1024`.

The image tiling works by tiling the image with tiles using the specified tiling type.
When using square tiles, the tiles are all square with the last tiles in each row and in the bottom row being padded to fill the square tile. When using rectangular tiles, the tiles are all the same size with the last tiles in each row and the bottom row being padded fill the tile. When using horizontal strips, the strips fill the row and the last strip is padded to fill the strip. When using vertical strips, strips fill the column and the last strip is padded to fill the strip. When using optimized tiles, the width and height are used as a maximum and the size is chosen to minimize the amount of padding.

Viskores did not support pan or zoom in the ray tracer so I modified `build_ascent.sh` to add a patch to viskores that implements pan and zoom. Pan and zoom were implemented with the 3d axes annotations so I made the pan and zoom in the ray tracer match what was done with the 3d axes.

I have tested tiling with the following pan and zoom values, which should cover the general case of pan and zoom values.
```
    scenes["s1/renders/r1/camera/xpan"] = "-0.125";
    scenes["s1/renders/r1/camera/ypan"] = "-0.125";
    scenes["s1/renders/r1/camera/zoom"] = "0.75";
```
I tested this in all 5 modes with combinations of image widths and tile sizes that should cover all the different cases.

I currently have the tiled rendering explicitly turned off in my tests that I checked in since they would currently fail with the existing containers. Viskores needs to be rebuilt with the `ascent_build.sh` so that it is built with my patch applied. I need help with doing that with the docker container and then I can re-enable tiled rendering in the tests.

I'm not sure about the defaults I've chosen. If you have thoughts on different ones, let me know. I think a larger tile width and height may be appropriate.

Here is a sample tiled image at 600x400 with a 256 tile size.
ascent_output_render_var1_zoom_000000


brugger1:
I added tests for tiling.

cyrush (approved):
Looks great -- this is an awesome addition. Thank you for this work!

I updated the spack based containers with the viskores patch and we are now fully green.

PR #1647, opened 2026-01-06
RZ mesh support

### LINKED ISSUE #1386: RZ mesh support
Need to add RZ mesh support to Ascent. A user reported the following:

```
[Error] Ascent::execute 
file: …/conduit-0.9.2/src/libs/conduit/conduit_node.cpp
line: 14396
message: 
Cannot fetch non-existent child "x" from Node([0]/coordsets/coords/values)
```

The mesh is:

```
coordsets: 
  coords: 
    type: "explicit"
    values: 
      z: [-2.0, -1.0, 0.0, ..., 0.0, 1.0]
      r: [0.0, 0.0, 0.0, ..., 4.0, 4.0]
topologies: 
  mesh: 
    type: "structured"
    coordset: "coords"
    elements: 
      dims: 
        i: 3
        j: 4
```

There are conduit examples of RZ meshes so test data will be easy to provide.

emily-howell:
Adding first pass at RZ mesh supoort for structured, unstructured, rectalinear, and uniform rz mesh types.

Focusing on adding support for these: [conduit rz mesh examples](https://github.com/llnl/conduit/blob/8ef9e1aeb65b601e19b262731347b719a3a056dd/src/libs/blueprint/conduit_blueprint_mesh_examples_rz_cylinder.cpp)

Some bugs may exist still


src/libs/ascent/runtimes/ascent_vtkh_data_adapter.cpp
```diff
@@ -1234,55 +1306,104 @@ VTKHDataAdapter::UniformBlueprintToViskoresDataSet
...
     float64 spacing_z = 1.0;
 
+    const bool is_rz = (n_coords.has_child("origin") && (n_coords["origin"].has_child("r") || (n_coords["origin"].has_child("z") && is_2d))) ||
+                 (n_coords.has_child("spacing") && (n_coords["spacing"].has_child("dr") || (n_coords["spacing"].has_child("dz") && is_2d)));
+    const bool is_rectilinear = (n_coords.has_child("origin") && (n_coords["origin"].has_child("x") || n_coords["origin"].has_child("y"))) ||
+                          (n_coords.has_child("spacing") && (n_coords["spacing"].has_child("dx") || n_coords["spacing"].has_child("dy"))) ||
+                          !is_rz;
+
+    if (is_rz && is_rectilinear) {
+        ASCENT_ERROR("Unsupported coordset: expected rectilinear {x,y,(z)} or cylindrical {r,z} but got parameters for both.")
```
cyrush: For the case of the underlying coordinate system (used to define the coordinate set) the right term is `cartesian`.

```suggestion
        ASCENT_ERROR("Unsupported coordset: expected cartesian {x,y,(z)} or cylindrical {r,z} but got parameters for both.")
```

src/libs/ascent/runtimes/ascent_vtkh_data_adapter.cpp
```diff
@@ -1234,55 +1306,104 @@ VTKHDataAdapter::UniformBlueprintToViskoresDataSet
...
     float64 origin_z = 0.0;
 
-
     float64 spacing_x = 1.0;
     float64 spacing_y = 1.0;
     float64 spacing_z = 1.0;
 
+    const bool is_rz = (n_coords.has_child("origin") && (n_coords["origin"].has_child("r") || (n_coords["origin"].has_child("z") && is_2d))) ||
+                 (n_coords.has_child("spacing") && (n_coords["spacing"].has_child("dr") || (n_coords["spacing"].has_child("dz") && is_2d)));
+    const bool is_rectilinear = (n_coords.has_child("origin") && (n_coords["origin"].has_child("x") || n_coords["origin"].has_child("y"))) ||
```
cyrush:
```suggestion
    const bool is_cartesian = (n_coords.has_child("origin") && (n_coords["origin"].has_child("x") || n_coords["origin"].has_child("y"))) ||
```

cyrush (dismissed):
Looks good!
To answer the question you eluded too last week - for the coord system naming, we do want cartesian (instead of rectilinear)

cyrush:
Also, please add a note to the changelog :-)

src/libs/ascent/runtimes/ascent_vtkh_data_adapter.cpp
cyrush: https://llnl-conduit.readthedocs.io/en/latest/blueprint_mesh.html#coordinate-sets

emily-howell: I changed all the applicable rectilinear and is_rectilinear that were refrencing the coordinates specifically to be cartesian and is_cartesian

src/libs/ascent/runtimes/ascent_vtkh_data_adapter.cpp
```diff
@@ -1296,15 +1417,42 @@ VTKHDataAdapter::UniformBlueprintToViskoresDataSet
...
+    }
 
     // todo, use actually coordset and topo names?
     result->AddCoordinateSystem( viskores::cont::CoordinateSystem(coords_name.c_str(),
                                                               dims,
                                                               origin,
                                                               spacing));
+
+    if(!is_2d && is_rz)
+    {
```
cyrush: Q: Which case does this apply to?

cyrush: Seems like old pass at 3d logic, should we remove and then we also dont need to include CylindricalCoordinateTransform

emily-howell: Yes, that is left over 3d rz changes that I missed before. I have removed those and took a second pass over the code to check for any more lingering 3d related changes

cyrush (approved):
Looks great, thanks for this work!

PR #1657, opened 2026-01-25
Restrict workflow permissions to minimum required

williamjallen:
Limiting the scope of the GITHUB_TOKEN via the [permissions](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions) attribute is best practice to prevent a leaked GITHUB_TOKEN from allowing repository write access.


cyrush (approved):
Thanks for sharing this info and this pr!

PR #1662, opened 2026-02-27
add "min"/"max" options for reflect transform

nicolemarsaglia:
Add "min"/"max" string options for the reflect filter's x/y/z params


nicolemarsaglia:
**Add 2D example!

src/docs/sphinx/Actions/Examples.rst
```diff
@@ -935,6 +935,148 @@ Resulting image:
...
+
+YAML actions:
+
+.. literalinclude:: examples/tout_transform_reflect_x_000100.yaml
+
+Resulting image:
+
+.. image:: examples/tout_transform_reflect_x_000100.png
+
+An example of using the transform filter to reflect over the x axis' max bounds
```
cyrush:
```suggestion
An example of using the transform filter to reflect over the x axis max bounds 
```

src/docs/sphinx/Actions/Examples.rst
```diff
@@ -935,6 +935,148 @@ Resulting image:
...
+
+YAML actions:
+
+.. literalinclude:: examples/tout_transform_reflect_x_max_000100.yaml
+
+Resulting image:
+
+.. image:: examples/tout_transform_reflect_x_max_000100.png
+
+An example of using the transform filter to reflect over the y axis' min bounds
```
cyrush:
```suggestion
An example of using the transform filter to reflect over the y axis min bounds 
```

src/libs/ascent/runtimes/flow_filters/ascent_runtime_vtkh_filters.cpp
```diff
@@ -5436,15 +5448,18 @@ VTKHTransform::execute()
...
-    double t_reflect[3]     = {0.0, 0.0, 0.0};
-    double t_matrix[16]     = {0.0, 0.0, 0.0, 0.0,
-                               0.0, 0.0, 0.0, 0.0,
-                               0.0, 0.0, 0.0, 0.0,
-                               0.0, 0.0, 0.0, 0.0};
+    double t_scale[3]          = {1.0, 1.0, 1.0};
+    double t_translate[3]      = {0.0, 0.0, 0.0};
+    double t_rotate_angle      =  0.0;
+    double t_rotate_axis[3]    = {0.0, 0.0, 0.0};
+    //TODO: Good normal default?
```
cyrush: 0,0,0 should be fine.

that said, should add an error if someone actually tries to use a 0,0,0 normal.

src/libs/vtkh/filters/PointTransform.cpp
```diff
@@ -96,57 +96,68 @@ PointTransform::SetScale(const double& sx,
...
-    // std::cout << matrix;
-    m_transform = matrix;
+  // reflect across plane defined by point(p) + normal(n):
+  // M = (I - 2 n n^T) with translation 2 (n·p) n
+
+  viskores::Vec<viskores::Float64,3> normal;
+  normal[0] = normalX;
+  normal[1] = normalY;
+  normal[2] = normalZ;
+  viskores::Normalize(normal);
```
cyrush, line 115: we should add a check for bad normal here, we may have to add it before we call normalize depending on how viskores behaves

src/tests/ascent/t_ascent_transform.cpp
```diff
@@ -708,10 +742,126 @@ TEST(ascent_translate, test_reflect_y)
...
+
+    Ascent ascent;
+    ascent.open();
+    ascent.publish(data);
+    ascent.execute(actions);
+    ascent.close();
+
+    // check that we created an image
+    EXPECT_TRUE(check_test_image(output_file));
+    std::string msg = "An example transform filter using reflect a 2D slice across y's minimum.";
```
cyrush:
```suggestion
    std::string msg = "An example transform filter using reflect a 2D slice across y axis minimum.";
```

cyrush (changes requested):
Looks great, a few minor inline suggestions + I think we should check for a bad normal (Normal with 0 mag)

src/libs/vtkh/filters/PointTransform.cpp
nicolemarsaglia: viskores will nan out, so we should we give the user an error, or force along (0,0,0) as the  normal and nothing will happen to the data?

cyrush: I think we should give the user an error. Better to let them know

nicolemarsaglia: I ended up putting the check in ascent_runtime_vtkh_filters.cpp after we parse the input.

cyrush (approved):
Looks good, thanks!

PR #1663, opened 2026-02-28
Task/2 3 2025 json schema for filters

emily-howell:
This is just a first pass and very much a draft version

Most focus was placed on building the actual structure. Before merging into an official version significant revisions to error handling in the validator should be made to add clarity.

This schema has currently only been added in `ascent_runtime_vtkh_filters`. The goal is to implement this for all filters in Ascent. Once this has been done and the full scope of Json Schema features needed has been determined, the `flow_schema_validator` should be migrated to conduit.


cyrush:
this approach looks nice and for the current cases it looks like less code to define the params.

Q: How are surprises handled, and how does that compare to our older error messages?

src/libs/flow/flow_schema_validator.hpp
```diff
@@ -0,0 +1,108 @@
...
+//-----------------------------------------------------------------------------
+namespace schema
+{
+
+bool FLOW_API validate(const conduit::Node &schema,
+                       const conduit::Node &input,
+                       conduit::Node &info);
+
+
+static bool validate_node(const conduit::Node &schema,
```
cyrush: No need for `static` here, if we want them hidden we could use a detail namespace in the c++ file,but I think it is fine to have them here as well w/o static

src/libs/ascent/runtimes/expressions/ascent_expression_jit_filters.cpp
```diff
@@ -140,31 +140,25 @@ ExprJitFilter::declare_interface(Node &i)
...
+  conduit::Node param_schema;
+  param_schema["type"] = "object";
+  param_schema["additionalProperties"] = true;
+
+  param_schema["properties/func"].set(filters::string_schema());
+  param_schema["properties/filter_name"].set(filters::string_schema());
+
+  conduit::Node inputs_schema = filters::array_schema(filters::ignore_schema());
+  inputs_schema["minItems"] = num_inputs;
+  inputs_schema["miaxItems"] = num_inputs;
```
cyrush:
```suggestion
  inputs_schema["maxItems"] = num_inputs;
```

cyrush: small typo

src/libs/ascent/runtimes/flow_filters/ascent_runtime_vtkh_filters.cpp
```diff
@@ -806,138 +708,102 @@ VTKHSlice::declare_interface(Node &i)
...
+    sphere_schema["type"] = "object";
+    sphere_schema["additionalProperties"] = false;
+    sphere_schema["properties/center"].set(vec3_schema(true));
+    sphere_schema["properties/radius"].set(number_schema(true));
+    sphere_schema["required"].append() = "center";
+    sphere_schema["required"].append() = "radius";
+    param_schema["properties/sphere"].set(sphere_schema);
+
+    // --- cylinder ---
+    conduit::Node cylinder_schema;
```
cyrush: consider using a ref to node in the tree instead of a separate node.

src/libs/flow/flow_schema_validator.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+//-----------------------------------------------------------------------------
+// -- begin flow::schema --
+//-----------------------------------------------------------------------------
+namespace schema
+{
+
+using ExpressionCheckFn = bool (*)(const std::string &expr, std::string &err_msg);
+
+void FLOW_API set_expression_checker(ExpressionCheckFn fn);
```
cyrush: json schema has a concept of string format checkers.

Generalize such that we can register format callbacks for categories of format checkers. 

```
register_format_checker(const std::string &case, callback)
```

and then when we use in ascent we can 

```
register_format_checker("expression",our_ascent_callback)
```

src/libs/flow/flow_filter.cpp
```diff
@@ -209,10 +221,22 @@ Filter::params()
...
 bool
-Filter::verify_params(const Node &, // unused: params,
+Filter::verify_params(const Node &params,
                       Node &info)
 {
     info.reset();
+
+    if (!param_schema().dtype().is_empty() && !(param_schema().dtype().is_object() && param_schema().number_of_children() == 0))
+    {
+        // std::cout << "\nSlice Properties!!" << std::endl;
```
cyrush: we can remove these commented out debug stmts

src/libs/flow/flow_schema_validator.cpp
```diff
@@ -0,0 +1,629 @@
...
+//-----------------------------------------------------------------------------
+// -- begin flow::schema::detail --
+//-----------------------------------------------------------------------------
+namespace detail
+{
+
+// ---------- General Helpers ----------
+ExpressionCheckFn &expr_checker()
+{
+    static ExpressionCheckFn fn = nullptr;
```
cyrush: this can become a map of format checkers to support general cases

src/libs/flow/flow_schema_validator.cpp
```diff
@@ -0,0 +1,629 @@
...
+                conduit::Node &info,
+                const std::string &path)
+{
+    const std::string schema_defined_type = get_type_string(schema);
+    if(schema_defined_type.empty()) return true; // schema didn't specify; treat as "accept anything"
+
+    const auto data_type = input.dtype();
+    bool ok = true;
+
+    if(schema_defined_type == "object") ok = data_type.is_object();
```
cyrush: formatting suggestion, final case uses {}, it would be good for even the oneliners to use 
```
{
oneliner
}
```

src/libs/flow/flow_schema_validator.cpp
```diff
@@ -0,0 +1,629 @@
...
+    const conduit::Node props_dummy;
+    const conduit::Node &props = has_props ? schema["properties"] : props_dummy;
+
+    bool ok = true;
+    for(conduit::index_t i = 0; i < input.number_of_children(); ++i)
+    {
+        const std::string k = input[i].name();
+        if(!has_props || !props.has_child(k))
+        {
+            add_error(info, "Unexpected field '" + conduit::utils::join_file_path(path, k) +
```
cyrush, line 275: You can use join_path, for conduit path, vs join_file_path for filesystem paths

```suggestion
            add_error(info, "Unexpected field '" + conduit::utils::join_path(path, k) +
```

src/libs/flow/flow_schema_validator.cpp
```diff
@@ -0,0 +1,629 @@
...
+        }
+        else if(tmp.has_child("errors") && tmp["errors"].number_of_children() > 0)
+        {
+            option_msgs.push_back(tmp["errors"].child(0).as_string());
+        }
+    }
+
+    if(matches >= 1) return true;
+
+    add_error(info, "anyOf violation at '" + (path.empty() ? std::string("<root>") : path) +
```
cyrush, line 499: maybe we can improve what the users sees so b/c anyOf may confuse them

src/libs/flow/flow_schema_validator.cpp
```diff
@@ -0,0 +1,629 @@
...
+bool validate_format(const conduit::Node &schema,
+                            const conduit::Node &input,
+                            conduit::Node &info,
+                            const std::string &path)
+{
+    if(!schema.has_child("format")) return true;
+    if(!schema["format"].dtype().is_string()) return true;
+
+    const std::string fmt = schema["format"].as_string();
+    if(fmt != "expression") return true;
```
cyrush: this is were we would implement general format callbacks

cyrush (commented):
Looks great a few inline comments. 

Also, we should consider accepting snake_case schema specifiers along with camelCase ones -- not sure it is worth the effort but would look nicer

src/libs/flow/flow_schema_validator.cpp
emily-howell: The language in the error messages does get improved upon in the documentation MR

src/libs/ascent/runtimes/flow_filters/ascent_runtime_vtkh_filters.cpp
emily-howell: Updated everything to use refrences to nodes rather than new seperate nodes. I also updated the docs in #1682 to reflect these changes

src/libs/ascent/runtimes/expressions/ascent_expression_jit_filters.cpp
emily-howell: Fixed in a commit locally.

src/libs/flow/flow_filter.cpp
emily-howell: Removed!

src/libs/flow/flow_schema_validator.cpp
emily-howell: I went through and fixed as many of these as I could find

src/libs/flow/flow_schema_validator.cpp
emily-howell: Change applied locally in a commit (on the branch #1682 whoops)

emily-howell:
@cyrush I went through and addressed all of your feedback. on all three branches. I tried to keep things still relatively with the original changes to keep things less noisy. I did update the documentation in #1682 to reflect the changes that I made so that might be a good place to start reviewing from.

src/libs/ascent/ascent.cpp
```diff
@@ -649,6 +651,9 @@ Ascent::open(const conduit::Node &options)
 
         m_runtime->Initialize(m_options);
 
+        // Set the flow filter expression checker:
+        runtime::filters::ascent_register_flow_schema_hooks();
```
cyrush, line 655: We only need init this once, but the cost isn't high so I am ok with keeping this

cyrush (approved):
Looks good, thanks for refactoring to create named string validators

PR #1666, opened 2026-03-11
add flow viz html file save option to `save_info`

### LINKED ISSUE #1586: make the workflow graph available via ascent_options
Users might want the flow chart ascent can generate: https://github.com/Alpine-DAV/ascent/blob/41b58a5a3ce459258e344d7b68c199dc50b7f1c8/src/libs/ascent/runtimes/ascent_main_runtime.cpp#L2154

also make it easier for me to not have to remember where this line is when I want to comment it back in

I thought it being in ascent_options would be a good fit, but can be convinced otherwise

cyrush:
resolves #1586 

* add option to save flow viz html file to save info
* fix with save info param name
* add test for save info with options

### DIFF
--- a/src/docs/sphinx/Actions/Actions.rst
+++ b/src/docs/sphinx/Actions/Actions.rst
@@ -57,7 +57,8 @@ Here is an example of an ascent actions yaml file using ``save_info``:
 
   -
     action: "save_info"
-    filename: "info_output.yaml" #optional
+    filename: "info_output.yaml" # optional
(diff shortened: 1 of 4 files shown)


cyrush:
@nicolemarsaglia  can you review and let me know if I missed anything?

nicolemarsaglia (approved):
Looks good! Great addition!

cyrush:
Thanks!

PR #1670, opened 2026-03-13
check for renderer type and return

nicolemarsaglia:
memory leak from erroneous rendererContainer being created when we do mesh plots by themselves


cyrush (approved):
Thanks for the fix!

PR #1671, opened 2026-03-16
add support for args passing to command callback

cyrush:
Adds support to plumb conduit `arguments` from an action to the command / callback filters.


src/tests/ascent/t_ascent_commands.cpp
```diff
@@ -66,6 +66,35 @@ bool bool_callback_2()
     return !example_bool_2;
 }
 
+//-----------------------------------------------------------------------------
```
cyrush, line 69: Example callback that processes arguments via `params`

src/tests/ascent/t_ascent_commands.cpp
```diff
@@ -371,6 +400,99 @@ TEST(ascent_commands, bool_callback_trigger_condition)
     ascent.close();
 }
 
+
+//-----------------------------------------------------------------------------
```
cyrush, line 404: Example of passing arbitrary arguments to a callback.

In this case we pass an actions file name to a callback that we expect to use another Ascent instance to render.

cyrush:
@benjaminjeliot Keep an eye on this PR, not ready yet but we are using it to help with your request.

benjaminjeliot:
Hi @cyrush, this is awesome -- thank you for the help.

PR #1679, opened 2026-03-24
unstructured topology with point shape: fix performance problem

albestro:
This PR fixes a problem that shouldn't produce any functional error, but it made this naive check super expensive. Indeed, due to the wrong type used for the `unordered_set`, with this fix my use-case went from ~3m to 12s.


cyrush (approved):
Thanks for this fix as well!

albestro:
Thanks for the confirmation @cyrush! Let me know if you'd expect something else to be done here or it can be merged as is.

In case, it would be nice to get at least a patch release 0.9.6 with the latest fixes if possible. Is there any schedule about it?

cyrush:
@albestro thanks again, the uniqueness check is expensive as is -- I wonder if it could be optionally relaxed even more.

Release plans:

We plan to release in late April.

We have been focusing on a large VisIt release and Conduit development.

In April, we will release Conduit, then Ascent.

PR #1685, opened 2026-04-03
Fix: build errors with VTKH enable with Kokkos HIP backend and ROCm 7.2.1

abdessamedkhediri-cmyk:
I'm currently working on in-situ GPU, and building ascent with vtkh enabled and the Kokkos HIP backend, I had to make some fixes. Details of each fix are available in each commit message.

My configuration: ROCm 7.2.1 with the hip gfx1201 architecture.


cyrush (approved):
Thank you for these fixes!

PR #1687, opened 2026-04-03
Disable CleanGrid point merging after threshold on point meshes

### LINKED ISSUE #1686: Threshold filter on point meshes crashes due to CleanGrid merging coincident points
**Description:**
When a threshold filter is applied to a point (particle) mesh, the post-threshold `CleanGrid` step runs with `MergePoints` enabled by default. If two or more particles on the same domain happen to be within the merge tolerance, `PointMerger` reduces the coordinate count but leaves all cells intact. For point topology, this produces a cell set where `GetNumberOfCells() > GetNumberOfPoints()`, which is invalid.
At render time, `SphereExtractor::SetPointIdsFromCoords` sizes the point ID array from `coords.GetNumberOfPoints()`, but the worklet dispatches over `GetNumberOfCells()` iterations. The final iterations index past the end of the coordinate array, triggering an assertion failure in `ArrayPortalBasicRead::Get`. See also https://github.com/Viskores/viskores/issues/275

The crash is intermittent — it requires a specific particle configuration where two particles on the same tile are within the merge tolerance. In our case this occurs under turbulent flow conditions in a PIC  simulation (MFiX-Exa) after approximately 9 seconds of simulation time.

**Steps to reproduce:**

  *  Publish a Conduit Blueprint mesh with an unstructured point topology (e.g., from `amrex::ParticleContainerToBlueprint`)
   * Apply a threshold filter to the particle data (e.g., threshold on particle phase)
   * Render the filtered particles using a pseudocolor point plot
   * Run until two particles on the same domain are within the `CleanGrid` default merge tolerance
   See also the new test `vtkh_threshold_point_mesh_coincident` which fails if `MergePoints` is not disabled in `Threshold.cpp`
 
**Crash location:**
With assertions enabled (debug build), there is an assertion failure at `viskores/internal/ArrayPortalBasic.h:77` —` assert(index < this->NumberOfValues) `inside `FindSphereAABBs` worklet.  Without assertions (`Release`/`RelWithDebInfo`), the out-of-bounds read returns garbage data, producing NaN coordinates that propagate through the AABB computation and eventually trigger a floating-point exception (SIGFPE) in the ray tracer.

**Root cause:**
`vtkh::Threshold::DoExecute()`  runs `CleanGrid` with default settings, including `MergePoints = true`. Point merging is inappropriate after threshold — threshold only removes cells, it does not create duplicate points.

**Fix**
Add required functions so that `vtkh::Threshold` can set `MergePoints = false` in Viskores:
`Threshold::DoExecute → CleanGrid::MergePoints(false) → viskoresCleanGrid::merge_points(false) → cleaner.SetMergePoints(false)` (the function names match existing coding style at each layer).  This is implemented in PR #1687 

A new test is added to `t_vtk-h_threshold.cpp` which exhibits the failure if `cleaner.MergePoints(false);` is not present in `vtk::Threshold` (src/libs/vtkh/filters/Threshold.cppL55)

charlesgwaldman:
fixes #1686

### DIFF
--- a/src/libs/vtkh/filters/CleanGrid.cpp
+++ b/src/libs/vtkh/filters/CleanGrid.cpp
@@ -43,6 +43,7 @@ CleanGrid::DoExecute()
     {
       cleaner.tolerance(m_tolerance);
     }
+    cleaner.merge_points(m_merge_points);
     auto dataset = cleaner.Run(dom, this->GetFieldSelection());
     this->m_output->AddDomain(dataset, domain_id);
   }
@@ -67,4 +68,9 @@ CleanGrid::Tolerance(const viskores::Float64 tolerance)
   m_tolerance = tolerance;
 }
 
+void
+CleanGrid::MergePoints(bool merge)
+{
+  m_merge_points = merge;
+}
 } // namespace vtkh
--- a/src/libs/vtkh/filters/CleanGrid.hpp
+++ b/src/libs/vtkh/filters/CleanGrid.hpp
@@ -17,11 +17,13 @@ public:
   virtual ~CleanGrid();
   std::string GetName() const override;
   void Tolerance(const viskores::Float64 tolerance);
+  void MergePoints(bool merge);
 protected:
   void PreExecute() override;
   void PostExecute() override;
   void DoExecute() override;
   viskores::Float64 m_tolerance;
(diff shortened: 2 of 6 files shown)


cyrush (approved):
Thank you for this description, fix, and test!

PR #1691, opened 2026-04-07
Task/4 2 2026 1027 plotting with just one color

emily-howell:
Implemented for both VTK-H and Devil Ray rendering. I took two different approaches. The implementation in the Devil Ray parsing builds a color_table with the one color and passes it to parse color table, the other just sets the color table directly. Curious which is prefered.

Also, I built this on top of the Json Schema changes. I figured it was easier to implement the changes to the parameter checking there then to have to update them later.


cyrush:
Since this builds off of the params refactor, I can't really isolate the changes for the mesh plotting easily. I will review when we have the other changes merged in.

src/tests/ascent/t_ascent_render_3d.cpp
```diff
@@ -2140,6 +2140,81 @@ TEST(ascent_render_3d, test_render_3d_compressed_color_table)
...
+    remove_test_image(output_file);
+
+    //
+    // Create the actions.
+    //
+
+    conduit::Node scenes;
+    scenes["s1/plots/p1/type"]  = "pseudocolor";
+    scenes["s1/plots/p1/field"] = "braid";
+    scenes["s1/plots/p1/color"] = {0.0, .184, .655}; // {0.748, 0.004, 0.008}
```
cyrush: how about
```suggestion
    scenes["s1/plots/p1/color_table/solid"] = {0.0, .184, .655};
```

emily-howell: I applied this change locally

emily-howell:
@cyrus All of the dependencies for this branch are merged in. This should be ready for review now

src/docs/sphinx/Actions/Scenes.rst
```diff
@@ -321,6 +321,24 @@ Here is an example of adding a custom color table to the volume plot:
...
+Monochrome / Solid Color Plotting
+^^^^^^^^^^^^^^^^^^^
+
+Generally, plotting is done using a range of colors from a color table; however, in some cases
+monochrome coloring is desirable. In these cases, the single color can be passed as a list of
+RGB or RGBA percentages values between 0 and 1 to the ``color_table/solid`` node.
+
+.. code-block:: c++
+  conduit::Node scenes;
+  scenes["s1/plots/p1/type"]  = "volume";
```
cyrush: should this be `pseudocolor` ?

cyrush (commented):
This looks good!

Users would also like to be able to create these style plots w/o a field.

```
s1:
  plots:
     p1:
       type: "mesh"
       topology: "mytopo" # optional, needed if mutiple topos exist
       color_table:
           solid: [0.0, .184, .655]
```


We should add an option to the interface that allows a solid color mesh plot.

This will require plumbing behind the scenes, maybe it is realized as a pc plot with a dummy field.


Side note:
We should also consider accepting hex style colors since those are ubiquitous vs floating point style colors.  Let's add a separate ticket for this,.

src/tests/ascent/t_ascent_dray.cpp
```diff
@@ -963,6 +963,70 @@ TEST(ascent_devil_ray, test_dray_compressed_color_table)
     // ASCENT_ACTIONS_DUMP(actions,output_file,msg);
 }
 
+//-----------------------------------------------------------------------------
+TEST(ascent_devil_ray, test_dray_Kazimir_Malevich)
```
cyrush, line 967: ❓

src/tests/_baseline_images/tout_render_3d_monochrome_mesh_000100.png
cyrush, line 1: We actually want this case to look exactly like the solid case. 

The idea is a colored plot of the surface without having identify or know about any of the fields.

src/tests/ascent/t_ascent_dray.cpp
cyrush: Kazimir_Malevich

emily-howell: The name? I was laughing because plotting a braid function in a solid color just makes it a cube and Nicole started listing off artists that just paint solid color squares.

cyrush: Thats a good explanation!

emily-howell: That's also why the render_3d test is named Yves_Klein because he is known for blue squares. Arguably Kazimir_Malevich is better known for black squares but he does red as well.

src/tests/_baseline_images/tout_render_3d_monochrome_mesh_000100.png
emily-howell: Sorry, I just assumed that you wanted this to just be a colored version of the mesh plot that is produced. Do you want a different backend to reroute this?

cyrush: Yes, I think we need to do something behind the scenes to make this work transparently for the user.

emily-howell: Okay, I added another test this time that does pseudo color plots without input fields.

cyrush:
we discussed and agreed that pseudocolor accepts a color table with a single color and no field is the best way to expose this to users.

cyrush:
We can also discuss aliases for plot types, here are some ideas:

```
wireframe ==> mesh
surface ==> pseudocolor with a single color
```

nicolemarsaglia (approved):
looks good!

cyrush (approved):
Looks good, thanks for getting the details ironed out!

PR #1696, opened 2026-04-13
Filename metadata broadcast

### LINKED ISSUE #1695: Filename expansion fails when rank 0 has no domains
**Description:**
When the MPI domain decomposition assigns no domains to rank 0 (e.g., 8 boxes across 128 ranks), filename expansion fails,  producing filenames like `test_%06d.png` instead of `test_000000.png`.

**Root cause:**: 
`PopulateMetadata()` reads state/cycle and state/time only from local domains. Ranks with no local domains leave cycle and time at the sentinel value -1, and the metadata is never populated.   The `expand_path_special_variables` function then passes `%06d` and `{cycle:06d}` through unexpanded.

**To reproduce:**
Run an Ascent pipeline where the number of boxes is less than the number of MPI ranks (e.g., `max_grid_size` large enough that only a few boxes are created) and `image_prefix: "test_%06d"`.  Output filenames will contain literal `%06d` instead of timestep numbers.

**Fix:**
Broadcast metadata (cycle, time) to all MPI ranks.

charlesgwaldman:
fixes #1695

### DIFF
--- a/scripts/build_ascent/build_ascent.sh
+++ b/scripts/build_ascent/build_ascent.sh
@@ -667,7 +667,7 @@ fi # if enable_hip || enable_sycl
 ################
 # Viskores
 ################
-viskores_version=1.1.0
+viskores_version=1.0.0
 viskores_src_dir=$(ospath ${source_dir}/viskores-${viskores_version})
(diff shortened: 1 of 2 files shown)


cyrush (approved):
Thank for the fix!

PR #1699, opened 2026-04-14
Tiled rendering depth test

### LINKED ISSUE #1698: Tiled rendering overwrites valid pixels
**Description:**
When rendering multiple domains using the tiled rendering path (`Renderer::RenderTiled`), tile pixels are copied unconditionally to the output canvas buffer. Domains rendered later overwrite pixels from earlier domains, even when the later domain contributes only background (no geometry in that tile region). This produces black gaps in the output image, as shown below

Image

The bug is visible when domain decomposition produces many small domains (e.g., a slice filter on an AMReX mesh with small max_grid_size). With fewer, larger domains the issue is less apparent because fewer overwrites occur.

**To reproduce:**
Run any Ascent pipeline with a slice filter on a domain-decomposed structured mesh where `max_grid_size` is small enough to produce multiple domains per rank. The output image shows a blocky/checkerboard pattern with missing patches instead of a continuous cross-section.

**Root cause:**
In `Renderer::RenderTiled`, the inner copy loop copies tile color and depth buffers to the canvas without checking whether the new pixel is closer than the existing one.

**Fix:** 
Add a depth test before overwriting canvas pixels, consistent with the z-buffer surface compositing used elsewhere in the rendering pipeline.

charlesgwaldman:
fixes #1698

### DIFF
--- a/scripts/build_ascent/build_ascent.sh
+++ b/scripts/build_ascent/build_ascent.sh
@@ -667,7 +667,7 @@ fi # if enable_hip || enable_sycl
 ################
 # Viskores
 ################
-viskores_version=1.1.0
+viskores_version=1.0.0
 viskores_src_dir=$(ospath ${source_dir}/viskores-${viskores_version})
 viskores_build_dir=$(ospath ${build_dir}/viskores-${viskores_version})
 viskores_install_dir=$(ospath ${install_dir}/viskores-${viskores_version}/)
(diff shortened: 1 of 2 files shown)


cyrush (approved):
Thanks for finding and resolving this issue!

brugger1:
Thanks a lot for the fix!

PR #1700, opened 2026-04-14
Adding a warning if an "ascent.py" exists in the current working directory

emily-howell:


cyrush (approved):
Thanks for adding this and the test!

Issue #1587, opened 2025-07-22
2025/07/22 tutorial feedback 


cyrush:
- [x] tutorial landing page:

```
docker run -p 8888:8888 -t -i alpinedav/ascent-jupyter:develop 
```

needs to be:

```
docker run -p 8888:8888 -t -i alpinedav/ascent-jupyter:latest
```

- [ ] triggers: update the ls examples to respect the new file format cycle output

JustinPrivitera:
- [x] "calcualte"

https://github.com/Alpine-DAV/ascent/blob/41b58a5a3ce459258e344d7b68c199dc50b7f1c8/src/examples/tutorial/ascent_intro/python/blueprint_example3.py#L24

JustinPrivitera:
- [x] missing apostrophe:

https://github.com/Alpine-DAV/ascent/blob/41b58a5a3ce459258e344d7b68c199dc50b7f1c8/src/examples/tutorial/ascent_intro/notebooks/03_conduit_blueprint_mesh_examples.ipynb#L15

JustinPrivitera:
- [x] Link doesn't render properly

<img width="1110" height="103" alt="Image" src="https://github.com/user-attachments/assets/2eab8ea8-277d-4e68-8c97-43192590de5b" />

PR #1701, opened 2026-04-14
Small docs updates addressing tutorial feedback and build warnings

emily-howell:


src/docs/sphinx/Actions/Examples.rst
```diff
@@ -793,51 +793,35 @@ Resulting image:
...
+-----------------------------------------------------------
 
 YAML actions:
 
 .. literalinclude:: examples/tout_sample_line_2d_000100.yaml
 
-Resulting image:
-
-.. image:: examples/tout_sample_line_2d_000100.png
-
```
emily-howell, line 805: This example and the following 3 showed an extract rather than images being plotted so I removed the `Resulting image` sections. Is that correct @nicolemarsaglia ?

src/docs/sphinx/developer_docs/BuildEnvironment.rst
```diff
@@ -190,14 +190,12 @@ file was created that contains the CMake variables need for an Ascent build:
...
 I Want To Develop Viskores and VTK-h Pipelines
----------------------------------------------
+-----------------------------------------------
 If you want to add new features to VTK-h, its source is developed inside
 the Ascent repo in the `src/libs/vtkh` directory.
 
 If your changes also require new features in Viskores, you will need to build
 and install your own version of Viskores. 
-Follow the examples in :ref:`building_manually` to create a Viskores build.
-
```
emily-howell, line 200: Could not find the section this was referencing. @cyrush do you know?

cyrush (approved):
Thanks for these changes!

src/docs/sphinx/developer_docs/BuildEnvironment.rst
cyrush: Might be a hold over from long ago.
I just noticed that read the docs only has `latest` and `stable` for Ascent. So I can't check old version. We should investigate that as well.

cyrush:
This resolves a subset of the issues in #1587 , i added checkboxes to that ticket we can mark off when this is merged. 

@emily-howell I think this is ready to go, modulo any feedback from @nicolemarsaglia about changes to the result images.

emily-howell:
Thanks for adding the checkboxes @cyrush. That helped me realize that I had missed a line item in the issue.

src/docs/sphinx/Actions/Examples.rst
nicolemarsaglia: yes, thank you!

Issue #1703, opened 2026-04-15
add plane mode to the sample filter


cyrush:
which will support non axis aligned uniform sampling  

We can use VisIt's IC Seeding Params for inspiration for the input params:

<img width="766" height="326" alt="Image" src="https://github.com/user-attachments/assets/aab07e1f-d8a2-4dd1-aa02-3199d7813e0b" />

```
- 
  action: "add_pipelines"
  pipelines: 
    pl1: 
      f1: 
        type: "sample"
        params: 
          fields: 
            - "braid"
          plane: 
            dims: 
              i: 5.0
              j: 5.0
            spacing:
              dx: 1.0
              dy: 1.0
            point: 
              x: 0.0
              y: 0.0
              z: 0.0
            normal: 
              x: 0.0
              y: 1.0
              z: 1.0
          invalid_value: -10.0
```

We need to think about `dx` vs `extents`

nicolemarsaglia:
I'm 90% sure this is already possible. Is someone trying to sample a plane and it is not working? 

Edit: Need to add normal and point

current use case (I think) is dims, spacing, origin

nicolemarsaglia:
Added in PR #1717

PR #1717, opened 2026-05-05
add plane to  sample filter using normal, point, dims, and spacing

nicolemarsaglia:
add plane sampling which needs normal, point, dims, and spacing


nicolemarsaglia:
This is for #1703

cyrush (approved):
Thanks for adding this!

PR #1720, opened 2026-05-08
use windows 2022 image due to gh actions running issue with vs 2026

cyrush:


cyrush:
There is an issue with detecting windows due to an env change. Will resolve with a new pr.

PR #1724, opened 2026-05-28
Removing all tilling rendering outputs before running tests

emily-howell:
This test was failing for me on subsequent runs locally due to test artifacts causing family value to increment past expected values. Added a more broad cleanup to start of test to remove all test artifacts before running the test.


cyrush (approved):
Thanks for this change!

PR #1728, opened 2026-05-29
Add topology as sample filter input

nicolemarsaglia:
Users can specify a topology that is present in their dataset to use as the input for the sample filter


src/libs/ascent/runtimes/flow_filters/ascent_runtime_vtkh_filters.cpp
```diff
@@ -3447,8 +3448,87 @@ VTKHSample::execute()
...
+#ifdef ASCENT_MPI_ENABLED
+      MPI_Comm mpi_comm = MPI_Comm_f2c(Workspace::default_mpi_comm());
+      int rank = 0;
+      MPI_Comm_rank(mpi_comm, &rank);
+
+      std::shared_ptr<conduit::Node> blueprint_data = data_object->as_low_order_bp();
+      conduit::Node sample_mesh;
+      if(rank == 0)
+      {
+        std::vector<const conduit::Node*> domains;
```
cyrush: conduit provides a helper to grab domain pts regardless of single or multiple domain meshes, here is an example:

https://github.com/llnl/conduit/blob/42e80cc194fd88de2f52b36e58f5ce03906d34ab/src/libs/blueprint/conduit_blueprint_mesh_utils.cpp#L479

lets use this to reduce code

cyrush (changes requested):
Looks good!

Two suggestions:

See inline suggestion about using a conduit helper to simplify grabbing domain pointers. 


For the topology sample outputs (the plane and sphere results) -- lets also render those and have baselines for them. Helps us visually confirm the sampling is working as expected.

src/libs/ascent/runtimes/flow_filters/ascent_runtime_vtkh_filters.cpp
```diff
@@ -3471,19 +3471,7 @@ VTKHSample::execute()
...
-        }
-        else //multi domain
-        {
-          const int num_domains = blueprint_data->number_of_children();
-          for(int i = 0; i < num_domains; ++i)
-          {
-            domains.push_back(&blueprint_data->child(i));
-          }
-        }
+        const std::vector<const Node *> domains = blueprint::mesh::domains(*blueprint_data);
```
cyrush: 👍

cyrush (approved):
Looks good -- thanks for the updates!

Issue #1725, opened 2026-05-28
image_height/width reversed


nicolemarsaglia:
User reports that the results of using `image_height` and `image_width` appear to be flipped. 

Also need to make sure we are catching any non-integer inputs and exiting cleanly.

nicolemarsaglia:
Could not recreate

PR #1732, opened 2026-06-05
frustum aspect ratio cast width and height as doubles

nicolemarsaglia:
Found trying to recreate #1725


cyrush (approved):
Thanks for this fix!

cyrush:
@nicolemarsaglia does this resolve#1725?

nicolemarsaglia:
> @nicolemarsaglia does this resolve#1725?

Yes and no. I couldn't recreate the error and this was a fix found along the way.

PR #1735, opened 2026-06-11
add materials to field_filtering

nicolemarsaglia:
account for marbl's special material naming scheme - do it like visit


cyrush:
To test, we can use:

https://github.com/visit-dav/visit/blob/develop/data/axom_klee_test_data.tar.xz

nicolemarsaglia:
@cyrush cuda build is running out of space again

nicolemarsaglia:
@cyrush this should get green checks. I tried to use codex to fix some spacing issues and it made things worse. I think I fixed all the spots, but apologies if you encounter an oddly spaced line of code, point it out and I'll fix it.

src/tests/ascent/t_ascent_conduit_extract.cpp
```diff
@@ -32,6 +34,49 @@ using namespace ascent;
 
 index_t EXAMPLE_MESH_SIDE_DIM = 10;
 
+namespace
```
cyrush, line 37: We should move these tests to where the MIR tests are, not the conduit extract tests, we can do that in a follow up PR.

cyrush (approved):
Looks good! There is some test logic that is in the conduit extract test source file that we should move to MIR.

PR #1736, opened 2026-06-12
Addressing infinite triggering loop for ascent_actions.yaml files

### LINKED ISSUE #1494: support for actions_file[s] ?
### Discussed in https://github.com/Alpine-DAV/ascent/discussions/1491

<div type='discussions-op-text'>

<sup>Originally posted by **jfavre** March 24, 2025</sup>
I was happily using a commit from a few weeks ago (git checkout 00621ed) with an ascent_actions.yaml as below:

```
-
  action: "add_triggers"
  triggers:
    t1:
      params:
        condition: "cycle() % 100 == 0"
        actions_file: "./add_scene.yaml"
```

        
Today, I checked out the latest commit and my execution now seg-faults. I only recompiled Ascent, did not touch my application itself and the yaml file. Moving a few steps back to commit ac3d1bc does the same. Moving one more step back to 00621ed, and my application works again. Was the syntax above working just by chance, or is it valid yaml, and commit ac3d1bc makes it fail? </div>

emily-howell:


cyrush:
azure pipelines cuda static builds are failing b/c we are running out of disk space

cyrush (approved):
Thanks for this change!

PR #1747, opened 2026-07-17
align cuda arch for cuda 12 testing

cyrush:


cyrush:
Found the source of sm 70 sneaking in to the cuda 12 test. 
Hopefully this also yields more disk space to work with, since we are not compiling for multiple arches.

PR #1749, opened 2026-07-21
Task/7 20 2026 hex color code support

emily-howell:


emily-howell:
@cyrush @nicolemarsaglia I think this is in a much more complete state now. The original MR just had support for hex with solid colors but now it works with all of the different color table formats. I also added a bunch of documentation and tests. Let me know what you think :)

cyrush (approved):
Looks great, thanks for this addition!

PR #1756, opened 2026-07-30
Remove unused variables

vil02:
These variables were allocated but never used.


src/libs/ascent/runtimes/flow_filters/ascent_runtime_genten_filters.cpp
```diff
@@ -382,8 +382,6 @@ Learn::execute()
     MPI_Allreduce(&num_domains, &global_blocks, 1, MPI_INT, MPI_SUM, mpi_comm);
 #endif
```
cyrush: Thanks for the suggestion, this code was a work in progress. 

Just in case the genten folks want to pick it up in the future, we should comment these out instead of fully removing the source.

```suggestion
//    double *kVecs = new double[num_fields*num_fields]; // TODO: need one per domain!!!!
//    double *eigvals = new double[num_fields];
```

cyrush (approved):

PR #1764, opened 2026-08-26
Fix a couple of bugs with tiled rendering.

### LINKED ISSUE #1739: Elements being dropped from render when resolution increases past 1024
Hi,

I have come across an issue where some elements don't get rendered when the resolution increases past 1024, this might be related to #1681.

I first came across this via [PyFR](https://github.com/PyFR/PyFR)'s ascent insitu visualisation plugin with mixed element meshes but I have created a standalone c++ reproducer separate from any solver. Two domains are published, 1 of quadrilateral elements and 1 of triangle elements which are interleaved in a checkerboard pattern to mimic a mixed element mesh. At or below 1024px everything appears to render correctly:

Image

But above this resolution half of the one domain isn't rendered and almost all of the other domain has been dropped:

Image

This is the reproducer:

[ascent-repro.tar.gz](https://github.com/user-attachments/files/29552201/ascent-repro.tar.gz)

I am using the build_ascent.sh script and using the head of develop. Let me know if there are any further details I can provide.

It does render correctly at all resolutions if I set tiled_rendering to false.

Thanks,
Toby

brugger1:
Resolves #1739 

This fixes the two bugs with tiled rendering from ticket #1739.

- Some of the tiles are blank in the image generated. This is due to a fix that wasn't quite correct where an if test was added around the code that copied a pixel from a tile to the final image so that the pixel was only copied if the depth in the tile was less than the depth in the final image. The if test included some code that incremented some indexes that should always get executed.
- The tiles that weren't blank exhibited a checker board pattern of missing image. The data had two domains where one domain had the red squares of a checkerboard and the second domain had the black squares of a checkerboard. There was a bug with the order of two loops such that the  last domain only made it to the final image. The loop order was changed to loop over tiles and within tiles, loop over domains.

Here is the image generated by the reproducer in the ticket after the change.

repro_600_000000

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,8 @@ and this project aspires to adhere to [Semantic Versioning](https://semver.org/s
 
 ### Changed
 - Redefined the parameter verification and surprise checking behavior for Filters to instead use a conduit based JSON Schema style verification schema. Implemented for all existing filters.
-- Change tiled rendering default to `false`.
+- Changed tiled rendering default to `false`.
+- A couple of bugs were fixed with tiled rendering and the tiled rendering default was changed to `true`.
 
 ### Fixed
 - Fixed the aspect ratio for frustum renderings
--- a/src/libs/vtkh/rendering/Render.cpp
+++ b/src/libs/vtkh/rendering/Render.cpp
@@ -18,7 +18,7 @@ Render::Render()
     m_render_background(true),
     m_shading(true),
     m_canvas(m_width, m_height),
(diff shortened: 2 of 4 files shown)


cyrush:
Looks great, thanks for these fixes.

One suggestion:
We have a note in the changelog that tiled rendering default was changed:
https://github.com/Alpine-DAV/ascent/blob/develop/CHANGELOG.md#changed

We should update this to say that a bug was fixed and tiled rendering defaults to on.

cyrush (approved):
Thanks this work and the changelog update

PR #1766, opened 2026-08-31
add viskores rotational and linear extrusion

nicolemarsaglia:


src/docs/sphinx/Actions/Pipelines.rst
```diff
@@ -226,6 +226,70 @@ The plane is defined by a point (on the plane) and a normal vector (not required
...
+The rotation axis is defined by a point on the axis and an axis direction vector.
+Angles are specified in degrees.
+Required parameters include: 
+specifying the rotational extrusion vector (``axis``)  via ``(r,z)`` or ``(x,y,z)``;
+and the degrees of rotation (``angle``).
+Optional parameters include: 
+ the ``point`` of rotation (default: (0,0)); 
+the ``start_angle``, in degrees, of where to begin the rotation (default: 0);
+the number of ``steps`` used to discretize the rotational extrusion (default: 32);
+``periodic`` is ``true`` when the final plane of revolution wraps to the first (default: ``false``);
```
cyrush: for periodic:

Does this mean that the first set of vertices are reused when the angle is 360 degrees?

cyrush: Reading code further that appears to be the case, implied by  `revolver.SetCloseSweep`.

Does this do anything for non-360 degree cases for revolve?

Wondering if we want it to be an option, or to just default to closed for cases where it makes sense (360 or start angle + sweep angle is equiv to 360)

cyrush (approved):
Looks great, one question about `periodic` and closed sweep semantics.

I will need to update the containers soon for conduit release and the arb ray pr. I will make sure this patch is in those containers.

src/docs/sphinx/Actions/Pipelines.rst
nicolemarsaglia: Interesting question! I can definitely make it automatic if degrees == 360 then periodic == true. 

Is there any use case of setting periodic to true if degrees is not 360? I'm not even sure if it would work, but say we have revolve with 270 degrees and 9 segments, but set periodic to true so the ninth segment goes an extra 90 degrees to get back to the start (?). Would anyone want that?  I'll give it a try to see what happens.

nicolemarsaglia: it fails in viskores if the sweep is not 360. Great. I'll make it automatic

cyrush: Sounds good! Thanks for running this down.

PR #1770, opened 2026-09-15
Fix Material Handling

### LINKED ISSUE #1667: Matset layout type recognition makes faulty assumptions
Here in `AddMatSets()`
https://github.com/Alpine-DAV/ascent/blob/f67dc0b80f2fa7bbb344d32af286be386235f0ab/src/libs/ascent/runtimes/ascent_vtkh_data_adapter.cpp#L3043

We handle three cases for matset layout types:
https://github.com/Alpine-DAV/ascent/blob/f67dc0b80f2fa7bbb344d32af286be386235f0ab/src/libs/ascent/runtimes/ascent_vtkh_data_adapter.cpp#L3128-L3131
https://github.com/Alpine-DAV/ascent/blob/f67dc0b80f2fa7bbb344d32af286be386235f0ab/src/libs/ascent/runtimes/ascent_vtkh_data_adapter.cpp#L3224-L3227
https://github.com/Alpine-DAV/ascent/blob/f67dc0b80f2fa7bbb344d32af286be386235f0ab/src/libs/ascent/runtimes/ascent_vtkh_data_adapter.cpp#L3318-L3321

We are not recognizing the cases properly. There are 4 cases:
- multi-buffer element-dominant ("full" case)
- multi-buffer material-dominant ("sparse by material" case)
- uni-buffer element-dominant ("sparse by element" case)
- uni-buffer material-dominant (unsupported but valid case)

**All cases may include a material map; it is required for the uni-buffer cases.** So checking for the existence of a material map to determine if we are in the uni-buffer element-dominant "sparse by element" case will lead to false positives.

**Both material-dominant cases include `element_ids`**, so checking for the existence of the `element_ids` is not going to always yield the multi-buffer material-dominant "sparse by material" case as the code assumes.

The way to determine if you are multi-buffer or uni-buffer is
```c++
conduit::blueprint::mesh::matset::is_multi_buffer();
conduit::blueprint::mesh::matset::is_uni_buffer();
// these are mutually exclusive states
```

The way to determine if you are element-dominant or material-dominant is
```c++
conduit::blueprint::mesh::matset::is_element_dominant();
conduit::blueprint::mesh::matset::is_material_dominant();
// these are mutually exclusive states
```

We also provide conversion routines in Conduit for moving between all the layout types.

We should audit the ascent code base to see if there is anywhere else where we are making assumptions about material set layouts.

I need to rewrite the conduit material set documentation to clear up the ambiguity, plus there is incorrect information in the current docs. The `venn` blueprint mesh example is not sufficient for demonstrating the various cases.

I am happy to work on this if need be; feel free to assign me to the ticket. Otherwise I can assist as needed.

JustinPrivitera:
Resolves #1667

WIP


JustinPrivitera:
depends on Conduit 0.9.9

TODO fix failing MIR test

